PR #3, opened 2019-04-24
Update Java wrapper

kevinAlbs:
basic.json now passes with command_started expectations


kevinAlbs:
Let's wait until we're ready to have the Java wrapper reviewed by someone else on the Java team.

PR #8, opened 2019-05-14
CDRIVER-2989: Java bindings

jyemin:
- Update Java binding tests to match libmongocrypt
- Add Evergreen testing of Java binding

Windows is not linking properly so I haven't enabled the tests yet for that variant:

Patch build from before I disabled Windows: https://evergreen.mongodb.com/version/5cdb1664e3c3312d98b517e0

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -40,6 +40,14 @@ functions:
           ${compile_env|} ./libmongocrypt/.evergreen/compile.sh
           ${test_env|} ./libmongocrypt/.evergreen/test.sh
 
+  "build and test java":
+    - command: "shell.exec"
+      params:
+        script: |-
+          eval "${inject_env_from_evergreen}"
+          ${compile_env|} ./libmongocrypt/.evergreen/compile.sh
+          cd ./libmongocrypt/bindings/java/mongocrypt && ${test_env|} ./.evergreen/test.sh
+
 tasks:
 - name: build-and-test
   commands:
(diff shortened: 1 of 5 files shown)


.evergreen/config.yml
```diff
@@ -92,24 +105,28 @@ buildvariants:
...
   - build-and-test-asan
+  - build-and-test-java
 - name: rhel72-zseries-test
   display_name: "RHEL 7.2 on zSeries"
   run_on: rhel72-zseries-test
   tasks:
   - build-and-test
   - build-and-test-asan
+  - build-and-test-java
 - name: windows-test
```
kevinAlbs: Consider adding a TODO comment for windows (or commenting on ticket) so we remember to do that. Unless we're tracking that elsewhere.

bindings/java/mongocrypt/.evergreen/test.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ "Windows_NT" = "$OS" ]; then
+   export JAVA_HOME=/cygdrive/c/java/jdk8
+else
+   export JAVA_HOME=/opt/java/jdk8
+fi
+
+./gradlew -version
+
+./gradlew -Djna.library.path="../../../cmake-build" clean check
```
kevinAlbs: Perhaps the fix for windows is just changing this path based on system, like Mark did in the test script:
https://github.com/10gen/libmongocrypt/blob/813131d4d15eec9a832ec14d573364c86fdb90ca/.evergreen/test.sh#L19-L22

So probably append `/Debug` to this path if this is Windows.

jyemin: I tried this earlier and it didn't work.

    Adding paths from jna.library.path: C:/data/mci/libmongocrypt/cmake-build/Debug
    Trying C:\data\mci\libmongocrypt\cmake-build\Debug\mongocrypt.dll

So I'm going to defer.

.evergreen/config.yml
jyemin: Opened https://jira.mongodb.org/browse/CDRIVER-3131

PR #9, opened 2019-05-20
Build: Ensure Jna libs are packaged alongside the java code

rozza:
* Migrated to kotlin
* Added pre publish check to ensure the Jna libs are as expected
* Added publishSnapshots task to push to the sonatype snapshots repo
* Added publishArchives task which checks the git version is a tagged
* Added jnaLibsPath flag to provide the Jna libs for inclusion in the build
* Added test helper to download the latest versions of the libs when testing

CDRIVER-3122


bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+tasks.withType<Test> {
+    @Suppress("UNCHECKED_CAST")
+    systemProperties((System.getProperties().toMap() as Map<String, Any>).filter { it.key.startsWith("jna.") })
+}
+
+/*
+ * Jna Libs - temporary testing helpers
```
jyemin: Seems like the right thing to do for testing is just to build locally from the cmake-build directory, rather than downloading.  This is what we do now in Evergreen and seems like the right thing, as you're guaranteed to be in sync and you only need the platform that you're actually running the tests on. 

 I think we only need to download files when we are publishing a snapshot or a release.

rozza:
Closing - will create a new pr once updated to capture the latest shared objects

PR #12, opened 2019-06-04
Update Java bindings to match new libmongocrypt C API

jyemin:


bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
```diff
@@ -96,37 +98,51 @@
...
 
     @Override
-    public MongoCryptContext createEncryptionContext(final String namespace, final BsonDocument localSchemaDocument) {
+    public MongoCryptContext createEncryptionContext(final String database, final BsonDocument commandDocument) {
         isTrue("open", !closed);
+        notNull("database", database);
+        notNull("commandDocument", commandDocument);
         mongocrypt_ctx_t context = mongocrypt_ctx_new(wrapped);
         if (context == null) {
             MongoCryptContextImpl.throwExceptionFromStatus(context);
```
kevinAlbs: Unrelated, but if the new context is NULL, that's an error on the mongocrypt_t handle. I think this should just be:

```
throwExceptionFromStatus();
```

jyemin: I'll fix it.

jyemin:
Rebased and pushed

PR #17, opened 2019-06-11
Build: Java add extra artifacts for individual builds

rozza:
Custom builds for:
* RHEL70
* Ubuntu1604
* osx
* Windows

CDRIVER-3175

### DIFF
--- a/bindings/java/mongocrypt/build.gradle.kts
+++ b/bindings/java/mongocrypt/build.gradle.kts
@@ -17,9 +17,7 @@
 
 import de.undercouch.gradle.tasks.download.Download
 import java.io.ByteArrayOutputStream
-import java.net.HttpURLConnection
 import java.net.URI
-import java.net.URL
 
 
 buildscript {
@@ -52,7 +50,7 @@ java {
 
 
 dependencies {
-    api("org.mongodb:bson:[3.10,4.1)")
+    api("org.mongodb:bson:[3.10,5.0)")
     api("net.java.dev.jna:jna:4.5.2")
     implementation("org.slf4j:slf4j-api:1.7.6")
 
--- a/bindings/java/mongocrypt/build.gradle.kts
+++ b/bindings/java/mongocrypt/build.gradle.kts
@@ -52,7 +52,7 @@ java {
 
 
 dependencies {
-    api("org.mongodb:bson:3.10.1")   // TODO: what version to depend on?
+    api("org.mongodb:bson:[3.10,4.1)")
     api("net.java.dev.jna:jna:4.5.2")
     implementation("org.slf4j:slf4j-api:1.7.6")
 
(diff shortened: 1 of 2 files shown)


bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -52,7 +52,7 @@ java {
 
 
 dependencies {
-    api("org.mongodb:bson:3.10.1")   // TODO: what version to depend on?
+    api("org.mongodb:bson:[3.10,4.1)")
```
rozza: @jyemin this range should be sufficient enough for now.

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -62,88 +62,8 @@ dependencies {
 }
 
 /*
- * Publishing
+ * Git version information
  */
-tasks.register<Jar>("sourcesJar") {
```
rozza: These had to be moved until after the classifier tasks were defined.

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -241,21 +117,30 @@ val copyResources by tasks.register<Copy>("copyResources") {
...
-    "https://s3.amazonaws.com/mciuploads/libmongocrypt/windows-test/master/$revision/libmongocrypt.tar.gz" to "win32-x86-64",
-    "https://s3.amazonaws.com/mciuploads/libmongocrypt/macos/master/$revision/libmongocrypt.tar.gz" to "darwin")
+data class LibMongoCryptS3Data(val evergreenName: String, val classifier: String, val osArch: String) {
+    fun downloadUrl(): String {
+        return "https://s3.amazonaws.com/mciuploads/libmongocrypt/$evergreenName/master/$revision/libmongocrypt.tar.gz"
+    }
+}
+
+// TODO - Any more libraries?
+val jnaMappingList: List<LibMongoCryptS3Data> = listOf(
```
rozza: This is the new mapping

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -342,3 +243,136 @@ tasks.withType<Test> {
...
+publishing {
+    publications {
+        create<MavenPublication>("mavenJava") {
+            artifactId = "mongocrypt"
+            from(components["java"])
+
+            artifact(tasks["sourcesJar"])
+            artifact(tasks["javadocJar"])
+            // Add special classifier jars
+            tasks.filter { it.name.endsWith("ClassifierJar") }.forEach {
```
rozza: Include the classifier jars

bindings/java/mongocrypt/build.gradle.kts
jyemin: We know that 4.0 is compatible, and therefore 4.x will be compatible, so let's make the outer limit 5.0.  That way we don't need to re-publish this when we release the 4.1 driver.

rozza: Done

PR #18, opened 2019-06-12
Publish Java snapshots on evergreen

rozza:
CDRIVER-3175

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -189,7 +189,6 @@ buildvariants:
   tasks:
   - build-and-test-and-upload
   - build-and-test-shared-bson
-  - build-and-test-java
 - name: linux-64-amazon-ami
   display_name: "Amazon Linux"
   run_on: amazon1-2018-test
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -48,6 +48,14 @@ functions:
           ${compile_env|} ./libmongocrypt/.evergreen/compile.sh
           cd ./libmongocrypt/bindings/java/mongocrypt && ${test_env|} ./.evergreen/test.sh
 
+  "publish snapshot":
+    - command: shell.exec
+      params:
+        silent: true
+        script: |-
+          eval "${inject_env_from_evergreen}"
+          cd ./libmongocrypt/bindings/java/mongocrypt && PROJECT_DIRECTORY=${project_directory} NEXUS_USERNAME=${nexus_username} NEXUS_PASSWORD=${nexus_password} SIGNING_PASSWORD=${signing_password} SIGNING_KEY_ID=${signing_keyId} RING_FILE_GPG_BASE64=${ring_file_gpg_base64} ./.evergreen/publish.sh
+
 tasks:
 - name: build-and-test-and-upload
   commands:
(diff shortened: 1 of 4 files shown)


rozza:
Patch Build: https://evergreen.mongodb.com/version/5d012696a4cf4734c3360399

After some debugging I determined it's missing the classifiers in sonatype becayse the patch builds to a different location in s3 and because it can't download the libs it won't add them as classifiers. 

I did check publishingSnapshots locally which is why you can see some uploaded to [sonatype](https://oss.sonatype.org/content/repositories/snapshots/org/mongodb/mongodb-crypt/1.0.0-SNAPSHOT/):

.evergreen/config.yml
```diff
@@ -228,6 +257,7 @@ buildvariants:
   - build-and-test-shared-bson
   - build-and-test-asan
   - build-and-test-java
+- name: suse12-s390x
```
jyemin: Was this a bug?

.evergreen/config.yml
```diff
@@ -117,8 +139,14 @@ pre:
            REMOTE_PATH="${project}/${build_variant}/${branch_name}/${tag_name}/libmongocrypt.tar.gz"
         fi
         unset tag_name
+        PROJECT_DIRECTORY="$(pwd)"
         echo "libmongocrypt_s3_path: $REMOTE_PATH"
-        echo "libmongocrypt_s3_path: $REMOTE_PATH" > expansion.yml
+        echo "project_directory: $PROJECT_DIRECTORY"
```
jyemin: Can you explain what's going on here?

.evergreen/config.yml
rozza: Yeah - appears to be two lots of settings but only one name

.evergreen/config.yml
```diff
@@ -161,6 +189,7 @@ buildvariants:
   tasks:
   - build-and-test-and-upload
   - build-and-test-shared-bson
+  - build-and-test-java
```
rozza: I'll remove this - it doesn't pass

.evergreen/config.yml
rozza: Yes, adding the current dir (project dir) to the expansion yaml file.  I believe its processed later via the call to `eval "${inject_env_from_evergreen}"`.  Its done differently in the java driver but not that differently.

.evergreen/config.yml
```diff
@@ -48,6 +48,14 @@ functions:
           ${compile_env|} ./libmongocrypt/.evergreen/compile.sh
           cd ./libmongocrypt/bindings/java/mongocrypt && ${test_env|} ./.evergreen/test.sh
 
+  "publish snapshot":
+    - command: shell.exec
+      params:
+        silent: true
+        script: |-
+          eval "${inject_env_from_evergreen}"
```
rozza: After this call `project_directory` is now available and in scope.

.evergreen/config.yml
jyemin: Which part doesn't pass.  Just Windows?  If so, can we just disable Windows?  IIUC this will disable all Java binding testing.

rozza: Thats just the windows build task list. There's already a ticket to fix / test windows and the java driver (CDRIVER-3131) that will re add it.

.evergreen/config.yml
jyemin: Ack

jyemin:
@kevinAlbs let us know if you want to review this.  Otherwise, please remove yourself from the reviewers list.  Thanks!

.evergreen/config.yml
kevinAlbs: inject_env_from_evergreen is only for the (now unused AWS credentials). Since I see `nexus_username` as a project variable in the Evergreen project page, it's always accessible with ${nexus_password}`. I can remove all occurrences of `${inject_env_from_evergreen}`, or if you'd like to here, go for it.

bindings/java/mongocrypt/.evergreen/publish.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# DO NOT ECHO COMMANDS AS THEY CONTAIN SECRETS!
```
kevinAlbs: Good call

.evergreen/config.yml
rozza: removed

PR #19, opened 2019-06-13
NODE-1854: add node bindings for client side encryption

mbroadst:
Many apologies for the massive single commit, but here we are. Here's a quick breakdown of what's going on in the bindings:

- Documentation is minimal at present, dramatic improvements forthcoming.
- The C++ code here is using Nan to bind to the `libmongocrypt` API. Generally, options management has been deferred to JavaScript (because it's easier to check values there). There is gratuitous use of smart pointers to protect ourselves from memory leaks. A lot of this should look familiar from our other addons
- The public API is all JavaScript, and comes in the form of two classes:
  - `AutoEncrypter`: this type is meant to be used internally in the driver if a user has provided the `autoEncryption` option to the `MongoClient` constructor (see more in forthcoming `node-mongodb-native` companion PR)
  - `ClientEncryption`: this type is used for "explicit encryption" (please see the spec for more details), and is meant to be exposed to users as public API for the client-side encryption feature.

Both of the exported JavaScript classes are backed by a common `StateMachine` class. I created it as a class mostly so we could mock out each of the "actions" it performs (see the tests for more details here). 

There are a number of places where we are doing explicit BSON serialization. Ideally, we would use `raw` queries for most of this, but our BSON implementation is just not up to the task today. There is follow-on work to improve how we handle raw values which would greatly improve performance.


bindings/node/src/mongocrypt.cc
```diff
@@ -0,0 +1,689 @@
...
+        return "Operation failed";
+    }
+
+    return errorMessage;
+}
+
+std::pair<bool, std::string> setKmsProviderOptions(mongocrypt_t* crypt,
+                                                   v8::Local<v8::Object> options) {
+    Nan::HandleScope scope;
+    v8::Local<v8::String> AWS_KEY = Nan::New("aws").ToLocalChecked();
```
daprahamian: q: these look relatively static. Can they be defined as global consts, or does v8/nan require to be defined inside the function invocation?

bindings/node/src/mongocrypt.cc
```diff
@@ -0,0 +1,689 @@
...
+        target, Nan::New("MongoCrypt").ToLocalChecked(), Nan::GetFunction(tpl).ToLocalChecked());
+}
+
+void MongoCrypt::logHandler(mongocrypt_log_level_t level,
+                            const char* message,
+                            uint32_t message_len,
+                            void* ctx) {
+    MongoCrypt* mongoCrypt = static_cast<MongoCrypt*>(ctx);
+    if (!mongoCrypt) {
+        fprintf(stderr, "Log handler called without `MongoCrypt` instance\n");
```
daprahamian: q: why are these going directly to stderr and not being propagated back up to the js side?

bindings/node/test/data/cmd.json
```diff
@@ -0,0 +1,6 @@
+{
+    "find": "test",
+    "filter": {
+        "ssn": "457-55-5462"
+    }
+}
```
daprahamian: nit: these files should end with newlines per unix convention.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -0,0 +1,153 @@
+'use strict';
+
+const fs = require('fs');
+const BSON = new (require('bson'))();
```
daprahamian: request: Add `//TODO: upgrade to 4.x bson`
q: should we also be testing against the 4.x bson? or is that overkill?

bindings/node/test/autoEncrypter.test.js
```diff
@@ -0,0 +1,153 @@
...
+const EJSON = require('mongodb-extjson');
+const expect = require('chai').expect;
+const sinon = require('sinon');
+const StateMachine = require('../lib/stateMachine').StateMachine;
+
+const SegfaultHandler = require('segfault-handler');
+SegfaultHandler.registerHandler();
+
+function readExtendedJsonToBuffer(path) {
```
daprahamian: nit/note: we should add `EJSON.toBuffer` and `EJSON.fromBuffer` methods to `bson`

bindings/node/test/autoEncrypter.test.js
```diff
@@ -0,0 +1,153 @@
...
+      });
+
+    sandbox.stub(StateMachine.prototype, 'fetchKeys').callsFake((client, ns, filter, callback) => {
+      // mock data is already seriaized, our action deals with the result of a cursor
+      const deserializedKey = BSON.deserialize(MOCK_KEYDOCUMENT_RESPONSE);
+      callback(null, [deserializedKey]);
+    });
+  });
+
+  after(() => {
```
daprahamian: q: stub/restore `beforeEach`/`afterEach` to prevent pollution across tests? Or are you only using sinon for stubbing and not for call tracking?

bindings/node/lib/autoEncrypter.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+const mc = require('bindings')('mongocrypt');
+const common = require('./common');
+const databaseNamespace = common.databaseNamespace;
+const StateMachine = require('./stateMachine').StateMachine;
+
+/**
```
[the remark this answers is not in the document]
mbroadst: My goal in keeping this all in the module was so that we could keep documentation isolated, similar to kerberos. I think we should speak more with docs about this, but I think for now it's early days and we can decide to move it by the 4.2 release if need be.

bindings/node/lib/autoEncrypter.js
```diff
@@ -0,0 +1,112 @@
...
+    // TODO: we should be able to do this without a try/catch
+    let context;
+    try {
+      context = this._mongocrypt.makeEncryptionContext(databaseNamespace(ns), commandBuffer);
+    } catch (err) {
+      callback(err, null);
+      return;
+    }
+
+    // TODO: should these be accessors from the addon?
```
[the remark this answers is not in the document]
mbroadst: there used to be a bunch more properties I was tacking on here, but the general question was "should this be something we track in C++ and expose from there?". I think generally the answer is: no, but I'm leaving the comment there to remind us to keep asking.

bindings/node/lib/autoEncrypter.js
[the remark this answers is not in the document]
daprahamian: We could still have it be part of the documentation, similar to how we handle `bson` [here](https://github.com/mongodb/node-mongodb-native/blob/fb0718689ca74deeb2a3471af199257758babd11/conf.json#L27)

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+      }
+
+      // terminal states
+      case MONGOCRYPT_CTX_READY:
+        callback(null, bson.deserialize(context.finalize()));
+        return;
+
+      case MONGOCRYPT_CTX_ERROR: {
+        const message = context.status.message;
+        callback(new Error(message));
```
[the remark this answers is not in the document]
mbroadst: :+1: added

bindings/node/test/autoEncrypter.test.js
```diff
@@ -0,0 +1,153 @@
...
+        logger: () => {},
+        kmsProviders: {
+          aws: { accessKeyId: 'example', secretAccessKey: 'example' },
+          local: { key: Buffer.alloc(96) }
+        }
+      });
+
+      mc.encrypt('test.test', TEST_COMMAND, (err, encrypted) => {
+        if (err) return done(err);
+        expect(encrypted).to.have.keys('filter', 'find');
```
kvwalker: I'm also confused by this test. Do we assert that the encryption was successful?

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+        return;
+    }
+  }
+
+  /**
+   *
+   * @param {*} kmsContext
+   */
+  kmsRequest(request) {
+    const options = { host: request.endpoint, port: 443 };
```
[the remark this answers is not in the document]
mbroadst: this is the required TLS port (defined in RFC2818)

bindings/node/test/autoEncrypter.test.js
```diff
@@ -0,0 +1,153 @@
...
+        logger,
+        kmsProviders: {
+          aws: { accessKeyId: 'example', secretAccessKey: 'example' },
+          local: { key: Buffer.alloc(96) }
+        }
+      });
+
+      mc.encrypt('test.test', TEST_COMMAND, (err, encrypted) => {
+        if (err) return done(err);
+        expect(encrypted).to.have.keys('filter', 'find');
```
kvwalker: Same with this test.

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+    const message = request.message;
+
+    return new Promise((resolve, reject) => {
+      const socket = tls.connect(options, () => {
```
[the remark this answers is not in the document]
mbroadst: nope! thanks, fixed.

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+        socket.write(message);
+      });
+
+      socket.on('timeout', () => {
```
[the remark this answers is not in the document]
mbroadst: fixed

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+   * @param {function} callback
+   */
+  fetchCollectionInfo(client, ns, filter, callback) {
+    const bson = client.topology.bson;
+    const dbName = databaseNamespace(ns);
+
+    client
+      .db(dbName)
+      .listCollections(filter)
+      .toArray((err, collections) => {
```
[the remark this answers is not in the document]
mbroadst: yeah, it's an implementation detail of the spec

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+          return;
+        }
+
+        const info = collections.length > 0 ? bson.serialize(collections[0]) : null;
+        callback(null, info);
+      });
+  }
+
+  /**
+   *
```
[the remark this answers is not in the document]
mbroadst: I've added them to remind myself to write the docs, they wont be empty in the future

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+   */
+  markCommand(client, ns, command, callback) {
+    const bson = client.topology.bson;
+    const dbName = databaseNamespace(ns);
+    const rawCommand = bson.deserialize(command);
```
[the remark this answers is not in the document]
mbroadst: part of it, but really almost everywhere we are serializing/deserializing, we should be able to work with a raw bson value

bindings/node/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "mongodb-client-encryption",
+  "version": "0.0.1",
```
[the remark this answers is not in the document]
mbroadst: :+1:

bindings/node/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "",
```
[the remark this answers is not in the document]
mbroadst: :+1:

bindings/node/test/autoEncrypter.test.js
mbroadst: I think until we move to bson 4.x in the driver, it's okay not to test here.

bindings/node/test/autoEncrypter.test.js
mbroadst: :+1:, ticket?

bindings/node/test/autoEncrypter.test.js
mbroadst: only using it for stubbing, no call tracking needed (yet)

bindings/node/test/data/cmd.json
mbroadst: :+1:

bindings/node/lib/stateMachine.js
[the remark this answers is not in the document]
daprahamian: nit: make it a global `const`

bindings/node/test/autoEncrypter.test.js
daprahamian: https://jira.mongodb.org/browse/NODE-2008

bindings/node/test/clientEncryption.test.js
```diff
@@ -0,0 +1,139 @@
...
+      keyVaultNamespace: 'client.encryption',
+      kmsProviders: { local: { key: Buffer.alloc(96) } }
+    });
+
+    encryption.createDataKey('local', (err, dataKey) => {
+      expect(err).to.not.exist;
+
+      const encryptOptions = {
+        keyId: dataKey,
+        algorithm: 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'
```
kvwalker: Is this a constant somewhere?

bindings/node/test/clientEncryption.test.js
```diff
@@ -0,0 +1,139 @@
...
+      };
+
+      encryption.encrypt('hello', encryptOptions, (err, encrypted) => {
+        expect(err).to.not.exist;
+        expect(encrypted._bsontype).to.equal('Binary');
+        expect(encrypted.sub_type).to.equal(6);
```
kvwalker: Can we add an `expect` for what the ciphertext should look like?

bindings/node/test/clientEncryption.test.js
```diff
@@ -0,0 +1,139 @@
...
+        encryption.decrypt(encrypted, (err, decrypted) => {
+          expect(err).to.not.exist;
+          expect(decrypted).to.equal('hello');
+          done();
+        });
+      });
+    });
+  });
+
+  it.skip('should explicitly encrypt and decrypt with the "aws" KMS provider', function(done) {
```
kvwalker: Why is this skipped?

bindings/node/test/clientEncryption.test.js
```diff
@@ -0,0 +1,139 @@
...
+      expect(err).to.not.exist;
+
+      const encryptOptions = {
+        keyId: dataKey,
+        algorithm: 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'
+      };
+
+      encryption.encrypt('hello', encryptOptions, (err, encrypted) => {
+        expect(err).to.not.exist;
+        expect(encrypted).to.have.property('v');
```
kvwalker: What is the v property?

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,231 @@
...
+
+    return new Promise((resolve, reject) => {
+      const socket = tls.connect(options, () => {
+        socket.write(message);
+      });
+
+      socket.once('timeout', () => {
+        socket.removeAllListeners();
+        socket.destroy();
+        reject(new Error('KMS request timed out'));
```
daprahamian: Should we make this our `MongoCryptError`?

bindings/node/lib/autoEncrypter.js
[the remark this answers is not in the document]
mbroadst: agreed

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+          if (err) {
+            return callback(err, null);
+          }
+
+          if (collInfo) {
+            context.addMongoOperationResponse(collInfo);
+          }
+
+          context.finishMongoOperation();
+          this.execute(autoEncrypter, context, callback);
```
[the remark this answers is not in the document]
mbroadst: yeah. It's a bit of a funky state machine, but you're correct that the `finish` actions move the machine to the next state.

bindings/node/lib/stateMachine.js
```diff
@@ -0,0 +1,227 @@
...
+          .catch(err => {
+            callback(err, null);
+          });
+
+        return;
+      }
+
+      // terminal states
+      case MONGOCRYPT_CTX_READY:
+        callback(null, bson.deserialize(context.finalize()));
```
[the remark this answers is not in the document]
mbroadst: No, it will always return a value

bindings/node/lib/stateMachine.js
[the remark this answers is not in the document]
mbroadst: done

bindings/node/lib/stateMachine.js
mbroadst: done

bindings/node/test/autoEncrypter.test.js
```diff
@@ -0,0 +1,153 @@
...
+        logger: () => {},
+        kmsProviders: {
+          aws: { accessKeyId: 'example', secretAccessKey: 'example' },
+          local: { key: Buffer.alloc(96) }
+        }
+      });
+
+      mc.decrypt(input, (err, decrypted) => {
+        if (err) return done(err);
+        expect(decrypted).to.eql({ filter: { find: 'test', ssn: '457-55-5462' } });
```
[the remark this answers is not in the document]
mbroadst: This is just because the `encrypted-document.json` file we are using has an encrypted BSON document that is a filter for a find operation. Just a coincidence, could have been `{ llamas: 'are cool' }`

bindings/node/test/autoEncrypter.test.js
mbroadst: so, we can't really assert the encrypted values, so what we are doing is asserting that the keys still exist

bindings/node/test/autoEncrypter.test.js
mbroadst: same thing as above. This is just copy-paste so that we could trigger a code path that makes a logging message occur

bindings/node/test/clientEncryption.test.js
mbroadst: it's not, but I'm not sure what the constant would be, I guess just the full string? This is a string meant to be passed to libmongocrypt, it will validate in the library

bindings/node/test/clientEncryption.test.js
mbroadst: we don't have a unit response of what the second AWS response is, I figured we could add to this in follow-on work. This is not the primary use case of this feature, so lower priority

bindings/node/test/clientEncryption.test.js
mbroadst: The `v` is a requirement from embedding libmongocrypt. In order to encrypt a single value we need to ensure that it's wrapped in a document because BSON only supports documents. By wrapping in `v`, we ensure we can handle documents and scalar values through encryption

bindings/node/test/autoEncrypter.test.js
kvwalker: Why can't we assert the encrypted values?

mbroadst: updated to assert the deterministically encrypted value

bindings/node/test/clientEncryption.test.js
mbroadst: we could convert it to base64 and assert the length, but otherwise the output is non-deterministic, so the prose tests don't ask for assertion here other than the binary subType

PR #20, opened 2019-06-27
Java: Fix exceptions when there is no context

rozza:
JAVA-3334


rozza:
rebased and merged in  79c6686

PR #21, opened 2019-06-27
PYTHON-1881 Initial in-line ABI mode CFFI python binding

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1881

This change adds a minimal python binding using "in-line" "ABI mode" CFFI. Tests pass locally:
```
$ python3.7 setup.py test
running test
running egg_info
writing pymongocrypt.egg-info/PKG-INFO
writing dependency_links to pymongocrypt.egg-info/dependency_links.txt
writing requirements to pymongocrypt.egg-info/requires.txt
writing top-level names to pymongocrypt.egg-info/top_level.txt
reading manifest file 'pymongocrypt.egg-info/SOURCES.txt'
writing manifest file 'pymongocrypt.egg-info/SOURCES.txt'
running build_ext
test_libmongocrypt_version (test.test_binding.TestBinding) ... ok
test_mongocrypt_binary_new (test.test_binding.TestBinding) ... ok
test_mongocrypt_new (test.test_binding.TestBinding) ... ok
test_mongocrypt_status_new (test.test_binding.TestBinding) ... ok
test_pymongocrypt_version (test.test_binding.TestBinding) ... ok

Ran 5 tests in 0.001s

OK
```

A higher level API to abstract the libbmongocrypt types and memory management will come later. As will Evergreen testing.


bindings/python/strip_header.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Generate a CFFI.cdef() string from a C header file
+
+Usage: python strip_header.py ../../src/mongocrypt.h | pbcopy
```
prashantmital: Same note about this being OS-X only. 

Suggest: 
```
Usage (on OS-X): python strip_header.py ../../src/mongocrypt.h | pbcopy
```

bindings/python/pymongocrypt/binding.py
```diff
@@ -0,0 +1,749 @@
...
+ */
+void
+mongocrypt_ctx_destroy (mongocrypt_ctx_t *ctx);
+""")
+
+lib = ffi.dlopen("mongocrypt")
+
+
+def libmongocrypt_version():
+    """Returns the version of libmongocrypt."""
```
prashantmital: Can you update the docstring to indicate that this is a convenience method added to avoid an additional call to `decode`? Also, can you say why a similar utility isn't added for `mongocrpyt_status_message`? At first glance, it appears it might be worthwhile to define a private utility to `decode` return values from functions bearing return type `* char *`.

```
def _sanitize_output(retval):
    assert isinstance(retval, bytes_type)     # bytes_type is str on PY2 defined in compat
    if PY3:
        return retval.decode()
    return retval
```

bindings/python/strip_header.py
ShaneHarvey: Done.

bindings/python/pymongocrypt/binding.py
ShaneHarvey: libmongocrypt_version is a convenient method to get the version of libmongocrypt that the pymongocrypt library is using. It will be useful for people reporting bugs as described in the "How To Ask For Help" section of the readme. 

We may add a `_sanitize_output` like method in a future PR but it's not needed yet.

ShaneHarvey: Actually I just added a `_to_string()` method. It's useful because [`ffi.string`](https://cffi.readthedocs.io/en/release-1.12/ref.html#ffi-string-ffi-unpack) returns `str` on Python 2 and `bytes` on Python 3. `_to_string()` returns `str` on both Python 2 and 3.

ShaneHarvey:
Rebased to master and updated the cdef header.

bindings/python/pymongocrypt/binding.py
prashantmital: Nice.

PR #22, opened 2019-07-02
Add support for keyAltName

rozza:
JAVA-3335

### DIFF
--- a/bindings/java/mongocrypt/build.gradle.kts
+++ b/bindings/java/mongocrypt/build.gradle.kts
@@ -42,7 +42,7 @@ repositories {
 }
 
 group = "org.mongodb"
-version = "1.0.0-beta2"
+version = "1.0.0-SNAPSHOT"
 description = "MongoDB client-side crypto support"
 
 java {
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptException.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptException.java
@@ -44,9 +44,13 @@ public class MongoCryptException extends RuntimeException {
      * @param status the status
      */
     MongoCryptException(final mongocrypt_status_t status) {
-        super(CAPI.mongocrypt_status_message(status, null).toString());
-        isTrue("status not ok", !CAPI.mongocrypt_status_ok(status));
-        code = mongocrypt_status_code(status);
+        super(getStatusMessage(status));
+        code = !CAPI.mongocrypt_status_ok(status) ? mongocrypt_status_code(status) : -1;
+    }
+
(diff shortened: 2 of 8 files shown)


bindings/java/mongocrypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
```diff
@@ -154,9 +154,50 @@ public void testExplicitEncryptionDecryption() throws IOException, URISyntaxExce
         mongoCrypt.close();
     }
 
+    @Test
+    public void testExplicitEncryptionDecryptionKeyAltName() throws IOException, URISyntaxException {
```
rozza: A copy of the previous but had to progress the decryptor context as well.

rozza:
Patch: https://evergreen.mongodb.com/version/5d1b623c3627e0746f41d279

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptException.java
```diff
@@ -44,9 +44,13 @@ public MongoCryptException(final String msg) {
      * @param status the status
      */
     MongoCryptException(final mongocrypt_status_t status) {
-        super(CAPI.mongocrypt_status_message(status, null).toString());
```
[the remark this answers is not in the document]
kevinAlbs: Hmm, do you recall how this came up? If libmongocrypt is returning false (indicating error) on a function that takes a status and not setting the status's message, that's a bug in libmongocrypt.

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -42,7 +42,7 @@ repositories {
 }
 
 group = "org.mongodb"
-version = "1.0.0-beta2"
+version = "1.0.0-SNAPSHOT"
```
kevinAlbs: I'm not sure what the convention here is. If this is meant to track the libmongocrypt version, than perhaps this should remain beta2 until we tag libmongocrypt with the beta3 tag?

rozza: I just wanted a SNAPSHOT to be published on the snapshots repo until the Java driver work is done. Once thats done a beta3 sounds good.

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptException.java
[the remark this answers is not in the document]
rozza: I can't trigger the error now on master - but I was previously able to fail the `isTrue` check and then that raises and IllegalArgumentException instead.  I'll revert the change.

rozza:
Rebased and merged

PR #25, opened 2019-07-10
NODE: rely on request.bytesNeeded for ending kms requests

daprahamian:
We were incorrectly calculating how many bytes we needed
to read, resulting in cryptic TLS errors. Now we
will rely solely on request.bytesNeeded.

TODO: Write tests


daprahamian:
Closing in favor of #26

PR #27, opened 2019-07-15
Ensure that JNA-managed memory is not GC'd before we're done with it

jyemin:
Since mongocrypt_binary_t is non-owning, the binding must ensure that
the data it creates one with is not freed before the mongocrypt_binary_t
is destroyed.  We accomplish this by creating a BinaryHolder class to
wrap the JNA memory and the mongocrypt_binary_t and make it
AutoCloseable.

CDRIVER-3233


jyemin:
Tested against linkbench and no longer seeing any crashes or malformed bson errors

PR #30, opened 2019-07-16
PYTHON-1904 Add wrapper classes around libmongocrypt

ShaneHarvey:
This changes makes managing the libmongocrypt CFFI objects safe and idiomatic. It also adds some tests that churn through the mongocrypt state machine using sample input/output data.

~Also included is the `AutoEncrypter` class and `MongoCryptCallback` interface which PyMongo will use to implement the required I/O to mognocryptd, mongod, and kms. I don't plan to add an implementation of `MongoCryptCallback` to this library because I want to avoid any cyclical dependencies: pymongo->pymongocrypt->pymongo~

Edit: I've removed the `AutoEncrypter`/`MongoCryptCallback` classes to simplify this PR. I'll add them back in the next PR.


bindings/python/test/test_mongocrypt.py
```diff
@@ -0,0 +1,224 @@
...
+    def test_mongocrypt_binary_in(self):
+        with MongoCryptBinaryIn(b'1\x0023') as binary:
+            self.assertEqual(binary.to_bytes(), b'1\x0023')
+        self.assertIsNone(binary.bin)
+
+        with MongoCryptBinaryIn(b'') as binary:
+            self.assertEqual(binary.to_bytes(), b'')
+        self.assertIsNone(binary.bin)
+
+        binary = MongoCryptBinaryIn(b'123')
```
prashantmital: Can you clarify what is being tested here with the creation of `binary`? Is it just the use of this type outside of a context manager?

prashantmital: Can you clarify what is being tested here with the creation of `binary`? Is it just the use of this type outside of a context manager?

bindings/python/test/test_mongocrypt.py
```diff
@@ -0,0 +1,224 @@
...
+        binary.close()
+        binary.close()
```
prashantmital: Why do we close this twice? Is it to ensure that closing an already closed instance doesn't cause an error?

bindings/python/test/test_mongocrypt.py
```diff
@@ -0,0 +1,224 @@
...
+            ({'aws': {'accessKeyId': '', 'secretAccessKey': ''}}, None),
+            ({'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'foo'}}, None),
+            ({'local': {'key': b'1'*96}}, None),
+            ({'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'foo'},
+              'local': {'key': b'1'*96}}, None),
+        ]
+        for kms_providers, schema_map in valid:
+            opts = MongoCryptOptions(kms_providers, schema_map)
+            self.assertEqual(opts.kms_providers, kms_providers)
+            self.assertEqual(opts.schema_map, schema_map)
```
prashantmital: Seems odd that we always use a `NULL`/`None` schema_map. Is it possible to test this with some 'dummy' schema?

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -0,0 +1,395 @@
...
+    def to_bytes(self):
+        """Returns this mongocrypt_binary_t as bytes."""
+        data = lib.mongocrypt_binary_data(self.bin)
+        if data == ffi.NULL:
+            return b''
+        data_len = lib.mongocrypt_binary_len(self.bin)
+        return ffi.unpack(ffi.cast("char*", data), data_len)
+
+
+class MongoCryptBinaryOut(_MongoCryptBinary):
```
prashantmital: Can you add a docstring to indicate that this is the type libmongocrypt returns?

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+class MongoCryptBinaryIn(_MongoCryptBinary):
+    __slots__ = ("__copy",)
+
+    def __init__(self, data):
+        """Creates a mongocrypt_binary_t from binary data."""
+        # mongocrypt_binary_t does not own the data it is passed so we need to
+        # create a copy to keep the data alive.
+        self.__copy = ffi.new("uint8_t[]", data)
```
prashantmital: Why use `uint8` here and `char*` on L49? Am I wrong to think that these should have the same type?

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -0,0 +1,395 @@
...
+        """Cleanup resources."""
+        if self.__ctx is None:
+            return
+        lib.mongocrypt_ctx_destroy(self.__ctx)
+        self.__ctx = None
+
+    def __del__(self):
+        self.close()
+
+    def __enter__(self):
```
prashantmital: Is there a reason we aren't using `contextlib`?

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -0,0 +1,395 @@
...
+        :Parameters:
+          - `ctx`: A mongocrypt_kms_ctx_t.
+        """
+        self.__ctx = ctx
+
+    def close(self):
+        """Cleanup resources."""
+        self.__ctx = None
+
+    def __enter__(self):
```
prashantmital: Same comment about contextlib.

bindings/python/test/test_mongocrypt.py
ShaneHarvey: I removed this. We always use these types in a with-statement so we don't need `__del__` or `close()`.

bindings/python/test/test_mongocrypt.py
ShaneHarvey: I removed this. We always use these types in a with-statement so we don't need `__del__` or `close()`.

bindings/python/test/test_mongocrypt.py
ShaneHarvey: Done.

bindings/python/pymongocrypt/mongocrypt.py
prashantmital: My bad. Ignore.

bindings/python/test/test_mongocrypt.py
prashantmital: Nice 👍

prashantmital:
LGTM

PR #33, opened 2019-07-26
feat(autoSpawn): handle auto-spawning mongocryptd in node bindings

daprahamian:


bindings/node/lib/common.js
```diff
@@ -1,69 +1,71 @@
 'use strict';
 
-function debug(msg) {
-  if (process.env.MONGODB_CRYPT_DEBUG) {
-    console.log(msg);
+module.exports = function() {
```
mbroadst: I think its probably fine to skip the dependency injection for modules that don't need it, or which are exclusively internal modules (like this one)

bindings/node/lib/mongocryptdManager.js
```diff
@@ -0,0 +1,123 @@
...
+      if ((state === STATE.VALID_PID_FILE) === up) {
+        return callback();
+      }
+
+      tries -= 1;
+      setTimeout(() => waitForPidFile(up, tries, callback), checkIntervalMS);
+    });
+  }
+
+  class MongocryptdManager {
```
mbroadst: similarly, this file doesn't seem to need any injection. Should it still follow the pattern?

bindings/node/lib/mongocryptdManager.js
```diff
@@ -0,0 +1,123 @@
...
+  const platform = require('os').platform;
+
+  const MONGOCRYPTD_PID_FILE = 'mongocryptd.pid';
+  const STATE = {
+    NO_PID_FILE: Symbol('NO_PID_FILE'),
+    EMPTY_PID_FILE: Symbol('EMPTY_PID_FILE'),
+    VALID_PID_FILE: Symbol('VALID_PID_FILE')
+  };
+
+  const checkIntervalMS = 100;
```
mbroadst: this seems a bit fast, how about 500ms?

bindings/node/lib/mongocryptdManager.js
```diff
@@ -0,0 +1,123 @@
...
+ * @prop {string} [mongocryptdSpawnPath] the path to the mongocryptd executable
+ * @prop {string[]} [mongocryptdURI] command line arguments to pass to the mongocryptd executable
+ */
+
+module.exports = function() {
+  const platform = require('os').platform;
+
+  const MONGOCRYPTD_PID_FILE = 'mongocryptd.pid';
+  const STATE = {
+    NO_PID_FILE: Symbol('NO_PID_FILE'),
```
mbroadst: bikeshed: I've been considering using the node core convention for constants and using `kPrefixedValues` (this actually just comes from Google's C++ style guide, and subsequently V8 I think).

bindings/node/lib/common.js
daprahamian: yeah, I'm okay with getting rid of it for `common` and `mongocryptdManager`. `stateMachine` is still going to need it to get the `MongoTimeoutError`

bindings/node/lib/mongocryptdManager.js
daprahamian: I was going off of the example [here](https://github.com/markbenvenuto/mongo/blob/cryptd_json/mongocryptd_controller.py#L65). tbh, I'd like this to be spec-d, but I'm fine with 500ms.

bindings/node/lib/mongocryptdManager.js
daprahamian: that sounds too much like hungarian notation to me :-/. I'd prefer to either keep them as all uppercase, or follow the [airbnb convention](https://github.com/airbnb/javascript#naming--uppercase), which in this case says it should just be camelCase.

bindings/node/lib/mongocryptdManager.js
daprahamian: this has the effect of making every autoEncryption startup take at least 500ms, which among other things significantly slows down the test runner.

Could we perhaps do 50 ms intervals with 20 attempts? that way, if it does indeed take a full second to start up, we will catch it, but we will immediately resolve if we have it?

PR #35, opened 2019-07-27
NODE-2071: implement crypto hooks for the node bindings

mbroadst:
This implements crypto callbacks so that the node bindings can be implemented without having to link to OpenSSL. The required methods are passed in via options passed into `MongoCrypt`. Ideally, this will be refactored to set the crypto hooks one time on the top-level module, but that can be follow-on work.

### DIFF
--- a/bindings/node/lib/cryptoCallbacks.js
+++ b/bindings/node/lib/cryptoCallbacks.js
@@ -1,0 +1,74 @@
+'use strict';
+const crypto = require('crypto');
+
+function aes256CbcEncryptHook(key, iv, input, output) {
+  let result;
+
+  try {
+    let cipher = crypto.createCipheriv('aes-256-cbc', key, iv);
+    cipher.setAutoPadding(false);
+    result = cipher.update(input);
+  } catch (e) {
+    console.dir({ e });
+  }
+
+  result.copy(output);
+  return result.length;
+}
+
+function aes256CbcDecryptHook(key, iv, input, output) {
+  let result;
+  try {
+    let cipher = crypto.createDecipheriv('aes-256-cbc', key, iv);
+    cipher.setAutoPadding(false);
+    result = cipher.update(input);
+  } catch (e) {
+    console.dir({ e });
+  }
+
+  result.copy(output);
+  return result.length;
+}
+
+function randomHook(buffer, count) {
+  crypto.randomFillSync(buffer, count);
+}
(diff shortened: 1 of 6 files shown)


bindings/node/lib/cryptoCallbacks.js
```diff
@@ -0,0 +1,74 @@
...
+
+function aes256CbcEncryptHook(key, iv, input, output) {
+  let result;
+
+  try {
+    let cipher = crypto.createCipheriv('aes-256-cbc', key, iv);
+    cipher.setAutoPadding(false);
+    result = cipher.update(input);
+  } catch (e) {
+    console.dir({ e });
```
daprahamian: do we want there to be `console.dir` here? seems like something you'd want to pass to whatever the `logger` is?

bindings/node/src/mongocrypt.cc
```diff
@@ -230,8 +238,129 @@ void MongoCrypt::logHandler(mongocrypt_log_level_t level,
...
+            v8::Local<v8::Object> outputBuffer = BufferFromBinaryNoCopy(out);
+
+            // TODO: error checking, use status if passed back value is not undefined
+            v8::Local<v8::Value> argv[] = {keyBuffer, inputBuffer, outputBuffer};
+            Nan::Call(*hook, Nan::GetCurrentContext()->Global(), 3, argv);
+            return true;
+        };
+
+    auto sha_256 =
+        [](void *ctx, mongocrypt_binary_t *in, mongocrypt_binary_t *out, mongocrypt_status_t *status) -> bool {
```
daprahamian: to understand the workflow of these: when `libmongocrypt` calls these functions, the first parameter is going to be the original `CryptoHooks` struct that we passed into `mongocrypt_setopt_crypto_hooks`, so we just have to cast it back out?

mbroadst: yep! it's a common pattern with C api's that you allow the caller to pass through `void*` data, so that you can use your own types in the callback. Particularly in this case its needed because C++ doesn't allow lambdas to capture in these situations.

bindings/node/lib/cryptoCallbacks.js
mbroadst: Created [NODE-2082](https://jira.mongodb.org/browse/NODE-2082) to track this work

PR #37, opened 2019-07-29
Support s390x and aarch64 platforms in Java binding build

jyemin:
https://jira.mongodb.org/browse/CDRIVER-3252
https://evergreen.mongodb.com/version/5d3f6acca4cf4754d1b78e62


bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -27,6 +27,9 @@ buildscript {
         jcenter()
         mavenCentral()
     }
+    dependencies {
+        "classpath"(group = "net.java.dev.jna", name = "jna", version = "4.5.2")
```
jyemin: Should I be using this weird `"classpath"` syntax?

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -99,8 +102,8 @@ val copyResources by tasks.register<Copy>("copyResources") {
         include("**/libmongocrypt.so", "**/libmongocrypt.dylib", "**/mongocrypt.dll")
     } else if (file(cmakeBuildPath).exists()){
         val jnaMapping = mapOf(
-                "libmongocrypt.so" to "linux-x86-64",
-                "mongocrypt.dll" to "win32-x86-64",
+                "libmongocrypt.so" to "linux-" + com.sun.jna.Platform.ARCH,
```
jyemin: Seemed the simplest way of using the same rules as JNA does to name the directory

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -116,39 +119,41 @@ val copyResources by tasks.register<Copy>("copyResources") {
 val downloadJnaLibs by tasks.register<DefaultTask>("downloadJnaLibs")
 val revision: String = System.getProperty("gitRevision", if (gitVersion == version) gitVersion else gitHash)
 
-data class LibMongoCryptS3Data(val evergreenName: String, val classifier: String, val osArch: String) {
+data class LibMongoCryptS3Data(val evergreenName: String, val osArch: String) {
```
jyemin: `classifier` was never fully removed.  Now it is.

bindings/java/mongocrypt/build.gradle.kts
rozza: SGTM

bindings/java/mongocrypt/build.gradle.kts
rozza: What does this do?

.evergreen/config.yml
```diff
@@ -134,7 +134,11 @@ tasks:
 
 - name: publish-snapshot
   depends_on:
-    - variant: ubuntu1604
+    - variant: rhel-62-64-bit
+      name: build-and-test-java
+    - variant: rhel-67-s390x
```
rozza: This test seems to have failed on evergreen.

bindings/java/mongocrypt/build.gradle.kts
jyemin: It's making jna a dependency on the build itself, so we can use the ARCH constant on line 102

rozza: Gotcha!

bindings/java/mongocrypt/build.gradle.kts
rozza: Ack

rozza:
So looks like its just the test on evergreen - I'm hoping its not a kotlin compatibility / build issue on s390x

jyemin:
Yeah, I ran it again and it still failed.  We have three other s390x variants that are passing so I think it's ok to proceed even thought that's red.

rozza:
weird - should the dependency for snapshots and the downloads be changed to a variant that builds?

rozza:
Actually looks like it could be this: https://github.com/gradle/gradle/issues/8505 - perhaps trying with gradle 5.3 will fix it.

jyemin:
I upgraded to Gradle 5.5.1 and the build proceeds, but then ran into another issue, reported here: https://jira.mongodb.org/browse/CDRIVER-3276.

For now I've disabled that variant and made publish-snapshot dependent on the C build succeeding for RHEL 6.7 s390x.

jyemin:
New patch build: https://evergreen.mongodb.com/version/5d49bb3f32f4176d1a9d3c90

PR #38, opened 2019-08-01
PYTHON-1928 Add state machine/callback/auto encryption API

ShaneHarvey:
Patch: https://evergreen.mongodb.com/version/5d436d763066150e7cbd0e28

This is the branch that https://jira.mongodb.org/browse/PYTHON-1884 is using.

### DIFF
--- a/bindings/python/pymongocrypt/state_machine.py
+++ b/bindings/python/pymongocrypt/state_machine.py
@@ -1,0 +1,139 @@
+# Copyright 2019-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


bindings/python/pymongocrypt/state_machine.py
```diff
@@ -0,0 +1,139 @@
...
+        elif state == lib.MONGOCRYPT_CTX_READY:
+            return ctx.finish()
+        elif state == lib.MONGOCRYPT_CTX_DONE:
+            return None
+
+        if state == lib.MONGOCRYPT_CTX_NEED_MONGO_COLLINFO:
+            list_colls_filter = ctx.mongo_operation()
+            coll_info = callback.collection_info(
+                ctx.database, list_colls_filter)
+            if coll_info:
```
behackett: I assume coll_info can be None? Should there be an else case?

ShaneHarvey: Yeah it can be None when the listCollections finds no matching collection. Skipping the `add_mongo_operation_result` and jumping straight to `complete_mongo_operation` is what we're supposed to do: https://jira.mongodb.org/browse/CDRIVER-3261

PR #39, opened 2019-08-05
NODE: Adding node tests to evergreen

daprahamian:
Have done the following:

1. Updated the unit tests so that all tests that require a `mongod` or `mongocryptd` can be skipped with the environment variable `NODE_SKIP_LIVE_TESTS`
1. Refactored `etc/build-static.sh` to properly keep track of directories and to copy over files installed in the "wrong" folder (`lib64` instead of `lib`).
1. Added a test task to libmongocrypt to build and test node.
    - Most tests are disabled in CI b/c they depend on a running `mongod` and/or `mongocryptd`. However, this at least gets a test that the library can be built and loaded

### DIFF
--- a/bindings/node/.evergreen/setup_environment.sh
+++ b/bindings/node/.evergreen/setup_environment.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+# set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+echo "PROJECT_DIRECTORY=${PROJECT_DIRECTORY}"
+NODE_LTS_NAME=${NODE_LTS_NAME:-dubnium}
+echo "NODE_LTS_NAME=${NODE_LTS_NAME}"
+NODE_BINDINGS_PATH="${PROJECT_DIRECTORY}/bindings/node"
+NODE_ARTIFACTS_PATH="${NODE_BINDINGS_PATH}/node-artifacts"
+NPM_CACHE_DIR="${NODE_ARTIFACTS_PATH}/npm"
+NPM_TMP_DIR="${NODE_ARTIFACTS_PATH}/tmp"
+BUILD_STATIC_SCRIPT="${NODE_BINDINGS_PATH}/etc/build_static.sh"
+
+# create node artifacts path if needed
+mkdir -p ${NODE_ARTIFACTS_PATH}
+mkdir -p ${NPM_CACHE_DIR}
+mkdir -p "${NPM_TMP_DIR}"
+
+# this needs to be explicitly exported for the nvm install below
+export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
+mkdir -p ${NVM_DIR}
+
+# install Node.js
+echo "Installing NVM"
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | bash
+[ -s "${NVM_DIR}/nvm.sh" ] && \. "${NVM_DIR}/nvm.sh"
+echo "Installing Node ${NODE_LTS_NAME}"
+nvm install --lts=${NODE_LTS_NAME}
+
+# setup npm cache in a local directory
+cat <<EOT > .npmrc
+devdir=${NPM_CACHE_DIR}/.node-gyp
+init-module=${NPM_CACHE_DIR}/.npm-init.js
+cache=${NPM_CACHE_DIR}
+tmp=${NPM_TMP_DIR}
+EOT
+
+# Add mongodb toolchain to path
+export PATH="/opt/mongodbtoolchain/v2/bin:$PATH"
+
+# if no mongocryptd installed, install mongocryptd and add it to path
+if ! [ -x "$(command -v mongocryptd)" ]
+then
+  echo "Installing mongocryptd"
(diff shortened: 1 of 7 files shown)


bindings/node/.evergreen/setup_environment.sh
```diff
@@ -0,0 +1,48 @@
...
+init-module=${NPM_CACHE_DIR}/.npm-init.js
+cache=${NPM_CACHE_DIR}
+tmp=${NPM_TMP_DIR}
+EOT
+
+# if no mongocryptd installed, install mongocryptd and add it to path
+if ! [ -x "$(command -v mongocryptd)" ]
+then
+  echo "Installing mongocryptd"
+  curl -o mongocryptd.tgz https://s3.amazonaws.com/mciuploads/mongodb-mongo-v4.2/enterprise-ubuntu1604-64/f92115cad9d2a4c2ddcf3c2c65092dda2fd7147a/binaries/mongo-cryptd-mongodb_mongo_v4.2_enterprise_ubuntu1604_64_f92115cad9d2a4c2ddcf3c2c65092dda2fd7147a_19_06_13_17_31_40.tgz
```
mbroadst: should this be defined as a variable above? I imagine the filename will change regularly

bindings/node/.evergreen/setup_environment.sh
```diff
@@ -0,0 +1,48 @@
...
+  mkdir -p mongocryptd && tar xzf mongocryptd.tgz -C mongocryptd --strip-components 1
+  export PATH="$PATH:$(pwd)/mongocryptd/bin"
```
mbroadst: consider always exporting something like `$(pwd)/bin` and using that for all installed binaries. This way there are fewer paths to manage

bindings/node/.evergreen/setup_environment.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+NODE_LTS_NAME=${NODE_LTS_NAME:-dubnium}
```
mbroadst: just to clarify, we're only testing `dubnium` right now, correct?

bindings/node/.evergreen/setup_environment.sh
```diff
@@ -0,0 +1,48 @@
...
+NODE_BINDINGS_PATH="${PROJECT_DIRECTORY}/bindings/node"
+NODE_ARTIFACTS_PATH="${NODE_BINDINGS_PATH}/node-artifacts"
+NPM_CACHE_DIR="${NODE_ARTIFACTS_PATH}/npm"
+NPM_TMP_DIR="${NODE_ARTIFACTS_PATH}/tmp"
+BUILD_STATIC_SCRIPT="${NODE_BINDINGS_PATH}/etc/build_static.sh"
```
mbroadst: is this used at all in this file?

bindings/node/etc/build-static.sh
```diff
@@ -11,22 +11,68 @@ mkdir -p $BUILD_DIR
 mkdir -p $BUILD_DIR/bson-build
 mkdir -p $BUILD_DIR/libmongocrypt-build
 
-pushd $BUILD_DIR
+# Copied from the mongo-c-driver
```
mbroadst: this seems specific to evergreen, can we leave it in a utilities file there which you `source` in here, or better yet run it before this file is run and ensure that `cmake` is in the `PATH` (perhaps have it installed to the `$(pwd)/bin` mentioned above)

bindings/node/test/autoEncrypter.test.js
```diff
@@ -183,6 +184,13 @@ describe('AutoEncrypter', function() {
   });
 
   describe('autoSpawn', function() {
+    beforeEach(function() {
+      if (process.env.NODE_SKIP_LIVE_TESTS) {
```
mbroadst: nit: conventionally we prefix these with `MONGODB_`

bindings/node/test/autoEncrypter.test.js
```diff
@@ -183,6 +184,13 @@ describe('AutoEncrypter', function() {
...
+        this.test.skip();
+        sandbox.restore();
```
mbroadst: is this call to `restore` needed? Does is a sandbox ever set up in this case?

bindings/node/.evergreen/setup_environment.sh
daprahamian: We are only testing dubnium, but we are still building all node version prebuilds from dubnium in #39 . I was a little concerned about overwhelming the testing matrix with node variants.

bindings/node/.evergreen/setup_environment.sh
daprahamian: oops

bindings/node/.evergreen/setup_environment.sh
daprahamian: good idea!

bindings/node/.evergreen/setup_environment.sh
daprahamian: yeah, I can try changing that. though tbh, we're not actually using the `mongocryptd` right now, so I am inclined to comment it out.

bindings/node/etc/build-static.sh
daprahamian: I'll try it out

bindings/node/test/autoEncrypter.test.js
daprahamian: It might not be needed anymore, I will check locally.

PR #40, opened 2019-08-06
PYTHON-1937 Add explicit encryption API

ShaneHarvey:
Still todo: 
- [x] Create jira ticket
- [x] Add test for create_data_key.
- [x] Fix copy/pasted doc strings.

Jira: https://jira.mongodb.org/browse/PYTHON-1937
Patch: https://evergreen.mongodb.com/version/5d48d9c657e85a23e31c596f

### DIFF
--- a/bindings/python/pymongocrypt/explicit_encrypter.py
+++ b/bindings/python/pymongocrypt/explicit_encrypter.py
@@ -1,0 +1,141 @@
+# Copyright 2019-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pymongocrypt.compat import PY3
+from pymongocrypt.mongocrypt import MongoCrypt
+from pymongocrypt.state_machine import run_state_machine
+
+
+class ExplicitEncryptOpts(object):
+    def __init__(self, algorithm, key_id=None, key_alt_name=None):
+        """Options for explicit encryption.
+
+        :Parameters:
+          - `algorithm`: The algorithm to use.
+          - `key_id`: The data key _id.
+          - `key_alt_name`: Identifies a key vault document by 'keyAltName'.
+        """
+        self.algorithm = algorithm
+        self.key_id = key_id
+        self.key_alt_name = key_alt_name
+        if PY3:
+            self.algorithm = algorithm.encode()
+            if key_alt_name is not None:
+                self.key_alt_name = key_alt_name.encode()
+
+
+class DataKeyOpts(object):
+    def __init__(self, master_key=None, key_alt_names=None):
+        """Options for creating encryption keys.
+
+        :Parameters:
+          - `master_key`: Identifies a KMS-specific key used to encrypt the
+            new data key. If the kmsProvider is "aws" it is required and must
+            have the following fields:
+            {
+               region: String,
+               key: String // The Amazon Resource Name (ARN) to the AWS
+                           // customer master key (CMK).
+            }
+            If the kmsProvider is "local" the masterKey is not applicable.
+          - `key_alt_name`: An optional list of string alternate names used to
+            reference a key. If a key is created with alternate names, then
+            encryption may refer to the key by the unique alternate name
+            instead of by _id
+        """
+        self.master_key = master_key
+        self.key_alt_names = key_alt_names
+
+
+class ExplicitEncrypter(object):
+    def __init__(self, callback, mongo_crypt_opts):
+        """Encrypts and decrypts BSON values.
+
+        This class is used by a driver to support for explicit encryption and
+        decryption of individual fields in a BSON document.
+
+        :Parameters:
+          - `callback`: A :class:`MongoCryptCallback`.
+          - `mongo_crypt_opts`: A :class:`MongoCryptOptions`.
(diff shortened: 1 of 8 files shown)


ShaneHarvey:
Updated patch: https://evergreen.mongodb.com/version/5d49cf332a60ed73d4cef09d

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -366,7 +388,130 @@ def __init__(self, ctx, command):
...
+          - `kms_provider`: The KMS provider.
+          - `opts`: An optional class:`DataKeyOpts`.
+        """
+        super(DataKeyContext, self).__init__(ctx)
+        try:
+            if kms_provider == 'aws':
+                if opts is None or opts.master_key is None:
+                    raise ValueError('opts.master_key is required for '
+                                     'kms_provider: "aws"')
+                if 'region' not in opts.master_key or 'key' not in \
```
behackett: Use parens, not \ for line continuation.

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -366,7 +388,130 @@ def __init__(self, ctx, command):
...
+                        opts.master_key:
+                    raise ValueError(
+                        'opts.master_key must include "region" and '
+                        '"key" for kms_provider: "aws"')
+                region = opts.master_key['region']
+                key = opts.master_key['key']
+                if PY3:
```
behackett: Same question about unicode in python 2.

bindings/python/test/test_mongocrypt.py
```diff
@@ -223,6 +225,9 @@ def fetch_keys(self, filter):
     def insert_data_key(self, data_key):
         raise NotImplementedError
 
+    def bson_encode(self, doc):
+        return BSON.encode(doc)
```
behackett: Yuck. You have bson.encode now and none of this will work with older PyMongo versions.

bindings/python/test/test_mongocrypt.py
```diff
@@ -261,6 +266,69 @@ def test_decrypt(self):
...
+    def __init__(self, kms_reply=None):
+        super(KeyVaultCallback, self).__init__(kms_reply=kms_reply)
+        self.data_key = None
+
+    def fetch_keys(self, filter):
+        return self.data_key
+
+    def insert_data_key(self, data_key):
+        self.data_key = data_key
+        return BSON(data_key).decode()['_id']
```
behackett: bson.decode()

bindings/python/test/test_mongocrypt.py
ShaneHarvey: I added the `bson_encode(self, doc) ` abstract method to deal with the fact that libmongocrypt requires the keyAltName to be passed as a BSON encoded document: https://jira.mongodb.org/browse/CDRIVER-3275. Instead of having pymongocrypt require pymongo I've implemented bson encoding via dependency injection.

This test implements a mock version of the callback API that pymongo will implement. In pymongo we can implement `bson_encode` using `bson.encode`.

> You have bson.encode now and none of this will work with older PyMongo versions.

I don't there's any concern for compatibility. This code uses the standard BSON.encode() API that has existed in pymongo forever. Only the tests for pymongocrypt require pymongo.

bindings/python/test/test_mongocrypt.py
ShaneHarvey: We can't use `bson.decode()` because it hasn't been released yet.

bindings/python/pymongocrypt/explicit_encrypter.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        self.master_key = master_key
+        self.key_alt_names = key_alt_names
+
+
+class ExplicitEncrypter(object):
+    def __init__(self, callback, mongo_crypt_opts):
+        """Encrypts and decrypts BSON values.
+
+        This class is used by a driver to support for explicit encryption and
```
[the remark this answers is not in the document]
ShaneHarvey: Done.

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -366,7 +388,130 @@ def __init__(self, ctx, command):
...
+            if not lib.mongocrypt_ctx_setopt_algorithm(
+                    ctx, opts.algorithm, -1):
+                self._raise_from_status()
+
+            if opts.key_id is not None:
+                with MongoCryptBinaryIn(opts.key_id) as binary:
+                    if not lib.mongocrypt_ctx_setopt_key_id(ctx, binary.bin):
+                        self._raise_from_status()
+
+            if opts.key_alt_name is not None:
```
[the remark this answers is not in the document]
ShaneHarvey: Right now we're depending on libmongocrypt to report the error if both options are set or any of the options are invalid.

bindings/python/pymongocrypt/mongocrypt.py
ShaneHarvey: Fixed.

bindings/python/test/test_mongocrypt.py
behackett: Oh, right, since this lives outside of PyMongo this would make testing more difficult at the moment,

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -351,7 +363,17 @@ def __init__(self, ctx, database, command):
         """
         super(EncryptionContext, self).__init__(ctx)
         self.database = database
-        self._encrypt_init(database, command)
+        try:
+            with MongoCryptBinaryIn(command) as binary:
+                if PY3:
```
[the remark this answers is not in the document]
ShaneHarvey: I fixed this by adding a `str_to_bytes` method. I also added a test with a unicode aws `master_key`.

bindings/python/pymongocrypt/mongocrypt.py
ShaneHarvey: I fixed this by adding a `str_to_bytes` method.

PR #42, opened 2019-08-06
Node use prebuild

daprahamian:
this is an extension of #39, and will be rebased once #39 is merged.

This adds support for using `prebuild` to publish prebuilt node bindings that can be auto-downloaded. This allows users to avoid having to install `libbson` or `libmongocrypt` locally.

### DIFF
--- a/bindings/node/.evergreen/prebuild.sh
+++ b/bindings/node/.evergreen/prebuild.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -o error
+
+get_version_at_git_rev () {
+  local REV=$1
+  local GIT_OUTPUT=$(git show $REV:./package.json)
+  local VERSION=$(echo $GIT_OUTPUT | node -r fs -e 'console.log(JSON.parse(fs.readFileSync("/dev/stdin", "utf-8")).version);')
+  echo $VERSION
+}
+
+run_prebuild() {
+  if [[ -z $NODE_GITHUB_TOKEN ]];then
+    echo "No github token set. Cannot run prebuild."
+    exit 1
+  else
+    echo "Github token detected. Running prebuild."
+    npm run prebuild -- -u $NODE_GITHUB_TOKEN
+  fi
+}
+
+VERSION_AT_HEAD=$(get_version_at_git_rev "HEAD")
+VERSION_AT_HEAD_1=$(get_version_at_git_rev "HEAD~1")
+
+if [[ ! -z $NODE_FORCE_PUBLISH ]]; then
+  echo '$NODE_FORCE_PUBLISH detected'
+  echo "Beginning prebuild"
+  run_prebuild
+elif [[ $VERSION_AT_HEAD != $VERSION_AT_HEAD_1 ]]; then
+  echo "Difference is package version ($VERSION_AT_HEAD_1 -> $VERSION_AT_HEAD)"
+  echo "Beginning prebuild"
+  run_prebuild
+else
+  echo "No difference is package version ($VERSION_AT_HEAD_1 -> $VERSION_AT_HEAD)"
+  echo "No prebuild required"
+fi
--- a/bindings/node/package.json
+++ b/bindings/node/package.json
@@ -7,12 +7,14 @@
     "lib": "lib"
   },
   "scripts": {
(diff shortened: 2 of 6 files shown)


bindings/node/.evergreen/setup_environment.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# set -o xtrace   # Write all commands first to stderr
```
kevinAlbs: Up to you if you wanna change this. Though verbose, sometimes this helps debugging failures in evergreen since all executed commands are logged. If it there are commands that are super verbose, you could disable that around those commands with `set +o xtrace`.

bindings/node/.evergreen/prebuild.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
```
kevinAlbs: `set -o error` and/or `set -o xtrace`?

bindings/node/.evergreen/prebuild.sh
```diff
@@ -0,0 +1,33 @@
...
+get_version_at_git_rev () {
+  local REV=$1
+  local GIT_OUTPUT=$(git show $REV:./package.json)
+  local VERSION=$(echo $GIT_OUTPUT | node -r fs -e 'console.log(JSON.parse(fs.readFileSync("/dev/stdin", "utf-8")).version);')
```
kevinAlbs: Oh nice.

bindings/node/.evergreen/prebuild.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo $VERSION
+}
+
+run_prebuild() {
+  if [[ -z $NODE_GITHUB_TOKEN ]];then
+    echo "No github token set. Cannot run prebuild."
```
kevinAlbs: `exit 1` ?

src/mongocrypt-ctx-datakey.c
```diff
@@ -107,7 +107,7 @@ _finalize (mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out)
 {
    _mongocrypt_ctx_datakey_t *dkctx;
    bson_t key_doc, child;
-   int64_t current_time_ms;
+   struct timeval tp;
```
kevinAlbs: This fix is in master per https://github.com/mongodb/libmongocrypt/commit/f9bde7222499314e561fabe8f21587ddbdbb169f

I'm not sure why it's showing up in this PR.

daprahamian: this needs to be rebased off of #39. H/o, I will update them.

bindings/node/.evergreen/setup_environment.sh
daprahamian: My concern with `set -o xtrace` was that we were using a secret token for the github push, and I did not want to risk that getting exposed. I might be a little overly paranoid on that front though, and I guess I could just border the `prebuild` push with `set +o xtrace`

bindings/node/package.json
```diff
@@ -7,20 +7,23 @@
...
-    "install": "node-gyp rebuild",
+    "install": "prebuild-install || node-gyp rebuild",
     "format-cxx": "git-clang-format",
     "format-js": "prettier --print-width 100 --tab-width 2 --single-quote --write index.js 'test/**/*.js' 'lib/**/*.js'",
     "lint": "eslint lib test",
     "docs": "jsdoc2md --template etc/README.hbs --plugin dmd-clear --files lib/**/*.js > README.md",
-    "test": "mocha test"
+    "test": "mocha test",
+    "rebuild": "prebuild --compile",
+    "prebuild": "prebuild  --strip --verbose --tag-prefix node-v -t 10.16.0 -t 8.16.0 -t 6.17.1 -t 4.9.1"
```
mbroadst: is there a reason we're not just building `--all`? Do we want to deprive users in e.g. electron, and do we want to get in to the habit of having to hand maintain this list of supported node versions?

daprahamian: We are not build `--all` b/c Node 12 is broken right now. See https://jira.mongodb.org/browse/NODE-2097. It's the usual Nan stuff.

PR #43, opened 2019-08-07
NODE: feat(cryptoCallbacks): pass crypto errors to libmongocrypt

daprahamian:
Passes crypto errors that occur on the js side back to
libmongocrypt. Requires every cryptoCallback to return a number
on success, or an error on failure.

Also required an update to the state machine logic to account for
an error occurring during mongocrypt_ctx_finalize.

~Note: errors in cryptoCallbacks sha256 and hmacSha256 will
cause a segfault until CDRIVER-3282 is fixed~ This is fixed

Fixes NODE-2082

### DIFF
--- a/bindings/node/lib/cryptoCallbacks.js
+++ b/bindings/node/lib/cryptoCallbacks.js
@@ -10,6 +10,7 @@ function aes256CbcEncryptHook(key, iv, input, output) {
     result = cipher.update(input);
   } catch (e) {
     console.dir({ e });
+    return e;
   }
 
   result.copy(output);
@@ -24,6 +25,7 @@ function aes256CbcDecryptHook(key, iv, input, output) {
     result = cipher.update(input);
   } catch (e) {
     console.dir({ e });
+    return e;
   }
 
   result.copy(output);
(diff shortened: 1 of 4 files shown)


bindings/node/test/cryptoCallbacks.test.js
```diff
@@ -0,0 +1,211 @@
...
+        });
+      });
+    });
+  });
+
+  describe('error testing', function() {
+    ['aes256CbcEncryptHook', 'aes256CbcDecryptHook', 'hmacSha512Hook'].forEach(hookName => {
+      it(`should properly propagate an error when ${hookName} fails`, function(done) {
+        const error = new Error('some random error text');
+        this.sinon.stub(cryptoCallbacks, hookName).returns(error);
```
kevinAlbs: oh this is neat

bindings/node/lib/cryptoCallbacks.js
```diff
@@ -24,14 +25,15 @@ function aes256CbcDecryptHook(key, iv, input, output) {
...
+    return e;
   }
 
   result.copy(output);
   return result.length;
 }
 
 function randomHook(buffer, count) {
-  crypto.randomFillSync(buffer, count);
+  crypto.randomFillSync(buffer, 0, count);
```
mbroadst: d'oh, sorry

PR #45, opened 2019-08-08
Remove clunky init method from python binding

ShaneHarvey:
Patch: https://evergreen.mongodb.com/version/5d4c7d9c7742ae7cdbbcc850

This change makes pymongocrypt load the library at import time and removes the clunky `init` method. I've also renamed MONGOCRYPT_LIB to PYMONGOCRYPT_LIB.

### DIFF
--- a/bindings/python/README.rst
+++ b/bindings/python/README.rst
@@ -95,10 +95,10 @@ like this:
   >>> import pymongocrypt
   Traceback (most recent call last):
     File "<stdin>", line 1, in <module>
-    File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/__init__.py", line 15, in <module>
-      from pymongocrypt.binding import mongocrypt_version, lib
-    File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/binding.py", line 741, in <module>
-      lib = ffi.dlopen("mongocrypt")
+    File "pymongocrypt/__init__.py", line 15, in <module>
+      from pymongocrypt.binding import libmongocrypt_version, lib
+    File "pymongocrypt/binding.py", line 803, in <module>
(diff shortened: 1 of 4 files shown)


bindings/python/README.rst
```diff
@@ -95,10 +95,10 @@ like this:
...
     File "<stdin>", line 1, in <module>
-    File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/__init__.py", line 15, in <module>
-      from pymongocrypt.binding import mongocrypt_version, lib
-    File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/binding.py", line 741, in <module>
-      lib = ffi.dlopen("mongocrypt")
+    File "pymongocrypt/__init__.py", line 15, in <module>
+      from pymongocrypt.binding import libmongocrypt_version, lib
+    File "pymongocrypt/binding.py", line 803, in <module>
+      lib = ffi.dlopen(os.environ.get('PYMONGOCRYPT_LIB', 'mongocrypt'))
     File "/Users/shane/venv/libmongocrypt/py3.7/lib/python3.7/site-packages/cffi/api.py", line 146, in dlopen
```
prashantmital: Do you think we should trim these paths to remove your name as well? AFAICT, the reader needs to be able to tell from the path that the CFFI package in site-packages is what is raising the error. The rest of the path is just noise. Maybe you can simply write:

```
File ".../pymongocrypt/binding.py", line 803, in <module>
      lib = ffi.dlopen(os.environ.get('PYMONGOCRYPT_LIB', 'mongocrypt'))
File ".../python3.7/site-packages/cffi/api.py", line 146, in dlopen
...
```

ShaneHarvey: Done.

PR #46, opened 2019-08-08
PYTHON-1925 Add libmongocrypt cypto hooks

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/PYTHON-1925
Patch: https://evergreen.mongodb.com/version/5d4cb6062a60ed12e5dd77f6

### DIFF
--- a/bindings/python/pymongocrypt/binary.py
+++ b/bindings/python/pymongocrypt/binary.py
@@ -28,7 +28,7 @@ def _to_bytes(mongocrypt_binary):
 
 
 def _write_bytes(mongocrypt_binary, data):
-    """Returns this mongocrypt_binary_t as bytes."""
+    """Writes the given data to a mongocrypt_binary_t."""
     buf = lib.mongocrypt_binary_data(mongocrypt_binary)
     if buf == ffi.NULL:
         raise MongoCryptError('mongocrypt_binary_data returned NULL')
--- a/bindings/python/pymongocrypt/binary.py
+++ b/bindings/python/pymongocrypt/binary.py
@@ -1,0 +1,96 @@
+# Copyright 2019-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Internal helpers for dealing with mongocrypt_binary_t."""
+
+from pymongocrypt.binding import ffi, lib
+from pymongocrypt.errors import MongoCryptError
+
+
+def _to_bytes(mongocrypt_binary):
+    """Returns this mongocrypt_binary_t as bytes."""
+    data = lib.mongocrypt_binary_data(mongocrypt_binary)
+    if data == ffi.NULL:
+        raise MongoCryptError('mongocrypt_binary_data returned NULL')
+    data_len = lib.mongocrypt_binary_len(mongocrypt_binary)
+    return ffi.unpack(ffi.cast("char*", data), data_len)
+
+
+def _write_bytes(mongocrypt_binary, data):
+    """Returns this mongocrypt_binary_t as bytes."""
+    buf = lib.mongocrypt_binary_data(mongocrypt_binary)
+    if buf == ffi.NULL:
+        raise MongoCryptError('mongocrypt_binary_data returned NULL')
+
+    ffi.memmove(buf, data, len(data))
+
+
+class _MongoCryptBinary(object):
+    __slots__ = ("bin",)
+
+    def __init__(self, binary):
+        """Wraps a mongocrypt_binary_t."""
+        self.bin = binary
+        if binary == ffi.NULL:
+            raise MongoCryptError(
+                "unable to create new mongocrypt_binary object")
+
+    def _close(self):
+        """Cleanup resources."""
+        if self.bin is None:
+            return
+        lib.mongocrypt_binary_destroy(self.bin)
+        self.bin = None
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._close()
+
+    def to_bytes(self):
+        """Returns this mongocrypt_binary_t as bytes."""
+        data = lib.mongocrypt_binary_data(self.bin)
+        if data == ffi.NULL:
+            return b''
+        data_len = lib.mongocrypt_binary_len(self.bin)
+        return ffi.unpack(ffi.cast("char*", data), data_len)
+
+
+class MongoCryptBinaryOut(_MongoCryptBinary):
+    __slots__ = ()
+
+    def __init__(self):
+        """Wraps a mongocrypt_binary_t."""
+        super(MongoCryptBinaryOut, self).__init__(lib.mongocrypt_binary_new())
+
+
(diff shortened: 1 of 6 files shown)


bindings/python/pymongocrypt/binary.py
```diff
@@ -0,0 +1,96 @@
...
+        """Wraps a mongocrypt_binary_t."""
+        self.bin = binary
+        if binary == ffi.NULL:
+            raise MongoCryptError(
+                "unable to create new mongocrypt_binary object")
+
+    def _close(self):
+        """Cleanup resources."""
+        if self.bin is None:
+            return
```
behackett: I kinda hate returns like this. I think putting everything under "if self.bin is not None" reads better.

bindings/python/pymongocrypt/binary.py
```diff
@@ -0,0 +1,96 @@
...
+    ffi.memmove(buf, data, len(data))
+
+
+class _MongoCryptBinary(object):
+    __slots__ = ("bin",)
+
+    def __init__(self, binary):
+        """Wraps a mongocrypt_binary_t."""
+        self.bin = binary
+        if binary == ffi.NULL:
```
behackett: Why not do this before setting the instance variable?

bindings/python/pymongocrypt/binary.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class MongoCryptBinaryIn(_MongoCryptBinary):
+    __slots__ = ("__copy",)
+
+    def __init__(self, data):
+        """Creates a mongocrypt_binary_t from binary data."""
+        # mongocrypt_binary_t does not own the data it is passed so we need to
+        # create a copy to keep the data alive.
+        self.__copy = ffi.new("uint8_t[]", data)
```
behackett: Does uint8_t work everywhere? For example, Python 2.7 on windows? You might have to use char[].

bindings/python/pymongocrypt/binary.py
```diff
@@ -0,0 +1,96 @@
...
+        super(MongoCryptBinaryIn, self).__init__(
+            lib.mongocrypt_binary_new_from_data(self.__copy, len(data)))
+
+    def _close(self):
+        """Cleanup resources."""
+        super(MongoCryptBinaryIn, self)._close()
+        if self.__copy is None:
+            return
```
behackett: Same thing here about bare returns.

bindings/python/pymongocrypt/crypto.py
```diff
@@ -0,0 +1,123 @@
...
+from pymongocrypt.binding import ffi, lib
+from pymongocrypt.compat import str_to_bytes
+
+
+def _callback_error_handler(exception, exc_value, tb):
+    """Set the mongocrypt_status_t on error."""
+    # From cffi docs: "First check if traceback is not None (it is None e.g.
+    # if the whole function ran successfully but there was an error converting
+    # the value returned: this occurs after the call)."
+    if tb is None:
```
behackett: I love that the cffi docs agree with me about "is not None". :-)

It's also weird that you return None here, but False later.

bindings/python/pymongocrypt/crypto.py
```diff
@@ -0,0 +1,123 @@
...
+
+@ffi.callback(
+    "bool(void *, mongocrypt_binary_t *, mongocrypt_binary_t *,"
+    "     mongocrypt_binary_t *, mongocrypt_binary_t *, uint32_t *,"
+    "     mongocrypt_status_t *)",
+    onerror=_callback_error_handler)
+def aes_256_cbc_encrypt(ctx, key, iv, input, output, bytes_written, status):
+    backend = default_backend()
+    # TODO: Java uses noPadding but CBC says padding is required:
+    # https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.modes.CBC
```
ShaneHarvey: @jyemin This comment is referencing the Java bindings:
https://github.com/mongodb/libmongocrypt/blob/5a2ff2d72c255b0f291891880b5163a8f18ab486/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java#L105-L108

Is that a bug in the Java bindings?

ShaneHarvey:
Updated passing patch: https://evergreen.mongodb.com/version/5d4cbcc10ae6064513aaa022

bindings/python/pymongocrypt/crypto.py
ShaneHarvey: Fixed.

bindings/python/pymongocrypt/crypto.py
ShaneHarvey: Jeff says:
> Its not a bug.  The library does the padding.

ShaneHarvey:
Updated passing patch: https://evergreen.mongodb.com/version/5d4cd8be1e2d172b4fe8fb0b

bindings/python/pymongocrypt/crypto.py
```diff
@@ -0,0 +1,117 @@
...
+from cryptography.hazmat.primitives.hashes import Hash, SHA256, SHA512
+from cryptography.hazmat.primitives.hmac import HMAC
+from cryptography.hazmat.backends import default_backend
+
+from pymongocrypt.binary import _to_bytes, _write_bytes
+from pymongocrypt.binding import ffi, lib
+from pymongocrypt.compat import str_to_bytes
+
+
+def _callback_error_handler(exception, exc_value, tb):
```
ShaneHarvey: See the cffi docs `onerror` docs for why we set the status via `tb.tb_frame.f_locals['status']`: 
> Note the following hack: in onerror, you can access the original callback arguments as follows. First check if traceback is not None (it is None e.g. if the whole function ran successfully but there was an error converting the value returned: this occurs after the call). If traceback is not None, then traceback.tb_frame is the frame of the outermost function, i.e. directly the frame of the function decorated with @ffi.def_extern(). So you can get the value of argname in that frame by reading traceback.tb_frame.f_locals['argname'].

https://cffi.readthedocs.io/en/latest/using.html?highlight=onerror#extern-python-reference

bindings/python/setup.py
```diff
@@ -19,7 +19,7 @@
     description="Python bindings for libmongocrypt",
     long_description=LONG_DESCRIPTION,
     packages=find_packages(),
-    install_requires=["cffi>=1.12.0,<2"],
+    install_requires=["cffi>=1.12.0,<2", "cryptography>=2.7,<3"],
```
ShaneHarvey: cryptography 2.7 is the latest release. I haven't attempted to use earlier releases yet but we may be able to do `"cryptography>=2,<3"` instead. That would provide greater compatibility for users who already have cryptography installed.

behackett: Do some spot testing, back to the earliest 2.x version to ensure they provide what we need.

bindings/python/pymongocrypt/binary.py
behackett: Fingers crossed. :-)

bindings/python/pymongocrypt/crypto.py
behackett: Specifically, libmongocrypt is doing the padding. We (or cryptography) can't do further padding on top of that. You should add a comment in our code making the situation clear. I'd hate to have a security researcher opening a security issue over a non-issue.

bindings/python/pymongocrypt/binary.py
ShaneHarvey: It seems to work so far on windows + Python 2.7 :)

bindings/python/pymongocrypt/crypto.py
ShaneHarvey: Done.

bindings/python/setup.py
ShaneHarvey: I'll do this in https://jira.mongodb.org/browse/PYTHON-1953.

PR #48, opened 2019-08-09
Node 2069/support key alt names

daprahamian:
Adds support for keyAltNames when defining data keys and during explicit encryption.

### DIFF
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -9,6 +9,43 @@ module.exports = function(modules) {
   const StateMachine = modules.stateMachine.StateMachine;
   const cryptoCallbacks = require('./cryptoCallbacks');
 
+  function sanitizeDataKeyOptions(bson, options) {
+    options = Object.assign({}, options);
+
+    // To avoid using libbson inside the bindings, we pre-serialize
+    // any keyAltNames here.
+    if (options.keyAltNames) {
+      if (!Array.isArray(options.keyAltNames)) {
+        throw new TypeError(
+          `Option "keyAltNames" must be an array of string, but was of type ${typeof options.keyAltNames}.`
+        );
+      }
+      const serializedKeyAltNames = [];
+      for (let i = 0; i < options.keyAltNames.length; i += 1) {
+        const item = options.keyAltNames[i];
+        const itemType = typeof item;
+        if (itemType !== 'string') {
+          throw new TypeError(
+            `Option "keyAltNames" must be an array of string, but item at index ${i} was of type ${itemType} `
+          );
+        }
+
+        serializedKeyAltNames.push(bson.serialize({ keyAltName: item }));
+      }
+
+      options.keyAltNames = serializedKeyAltNames;
+    } else if (options.keyAltNAmes == null) {
+      // If keyAltNames is null or undefined, we can assume the intent of
+      // the user is to not pass in the value. B/c Nan::Has will still
+      // register a value of null or undefined as present as long
+      // as the key is present, we delete it off of the options
+      // object here.
+      delete options.keyAltNames;
+    }
+
+    return options;
+  }
+
   /**
    * The public interface for explicit client side encryption
    */
@@ -42,34 +79,7 @@ module.exports = function(modules) {
      */
     createDataKey(provider, options, callback) {
       if (typeof options === 'function') (callback = options), (options = {});
-      options = Object.assign({}, options);
-
-      // To avoid using libbson inside the bindings, we pre-serialize
-      // any keyAltNames here.
-      if (options.keyAltNames) {
-        if (!Array.isArray(options.keyAltNames)) {
-          throw new TypeError(
-            `Option "keyAltNames" must be an array of string, but was of type ${typeof options.keyAltNames}.`
-          );
-        }
-        const bson = this._bson;
-        const serializedKeyAltNames = [];
-        for (let i = 0; i < options.keyAltNames.length; i += 1) {
-          const item = options.keyAltNames[i];
-          const itemType = typeof item;
-          if (itemType !== 'string') {
-            throw new TypeError(
-              `Option "keyAltNames" must be an array of string, but item at index ${i} was of type ${itemType} `
-            );
-          }
-
-          serializedKeyAltNames.push(bson.serialize({ keyAltName: item }));
-        }
-
-        options.keyAltNames = serializedKeyAltNames;
-      } else {
-        delete options.keyAltNames;
-      }
+      options = sanitizeDataKeyOptions(this._bson, options);
 
       const context = this._mongoCrypt.makeDataKeyContext(provider, options);
       const stateMachine = new StateMachine();
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -42,7 +42,34 @@ module.exports = function(modules) {
      */
     createDataKey(provider, options, callback) {
       if (typeof options === 'function') (callback = options), (options = {});
-      options = options || {};
+      options = Object.assign({}, options);
+
+      // To avoid using libbson inside the bindings, we pre-serialize
+      // any keyAltNames here.
+      if (options.keyAltNames) {
+        if (!Array.isArray(options.keyAltNames)) {
+          throw new TypeError(
+            `Option "keyAltNames" must be an array of string, but was of type ${typeof options.keyAltNames}.`
+          );
+        }
+        const bson = this._bson;
+        const serializedKeyAltNames = [];
+        for (let i = 0; i < options.keyAltNames.length; i += 1) {
+          const item = options.keyAltNames[i];
+          const itemType = typeof item;
+          if (itemType !== 'string') {
+            throw new TypeError(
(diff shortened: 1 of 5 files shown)


bindings/node/lib/clientEncryption.js
```diff
@@ -42,7 +42,34 @@ module.exports = function(modules) {
...
+              `Option "keyAltNames" must be an array of string, but item at index ${i} was of type ${itemType} `
+            );
+          }
+
+          serializedKeyAltNames.push(bson.serialize({ keyAltName: item }));
+        }
+
+        options.keyAltNames = serializedKeyAltNames;
+      } else {
+        delete options.keyAltNames;
```
mbroadst: I think we shouldn't delete anything from a user passed in object, let's just ignore it if its not a type we care about.

bindings/node/lib/clientEncryption.js
```diff
@@ -86,6 +113,20 @@ module.exports = function(modules) {
...
       }
+      if (options.keyAltName) {
+        const keyAltName = options.keyAltName;
+        if (options.keyId) {
+          throw new TypeError(`"options" cannot contain both "keyId" and "keyAltName"`);
+        }
+        const keyAltNameType = typeof keyAltName;
+        if (keyAltNameType !== 'string') {
+          throw new TypeError(
+            `"options.keyAltName" must be of type string, but was of type ${keyAltNameType}`
```
mbroadst: This error seems inconsistent with the ones above where you're saying:
```
 `Option "keyAltNames" must be an array of string, but was of type ${typeof options.keyAltNames}.`
```

bindings/node/test/clientEncryption.test.js
```diff
@@ -32,7 +35,7 @@ describe('ClientEncryption', function() {
   });
 
   beforeEach(function() {
-    if (process.env.MONGODB_NODE_SKIP_LIVE_TESTS) {
+    if (requirements.SKIP_LIVE_TESTS) {
```
mbroadst: looks like you might need to rebase

bindings/node/test/clientEncryption.test.js
```diff
@@ -186,3 +189,143 @@ describe('ClientEncryption', function() {
 
   it('should be able to auto decrypt explicit encryption');
 });
+
+describe('ClientEncryptionKeyAltNames', function() {
```
mbroadst: should this just be a `describe` block within `ClientEncryption` called `keyAltNames`?

bindings/node/test/clientEncryption.test.js
```diff
@@ -186,3 +189,143 @@ describe('ClientEncryption', function() {
...
+  });
+
+  it('should create a key with multiple keyAltNames', function() {
+    let dataKey;
+    let doc1;
+    let doc2;
+    return this.encryption
+      .createDataKey('aws', makeOptions(['foobar', 'fizzbuzz']))
+      .then(_dataKey => (dataKey = _dataKey))
+      .then(() => this.collection.findOne({ keyAltNames: 'foobar' }))
```
mbroadst: nit: these two can just be a `Promise.all` instead of the redundant `then` blocks and storing to a variable outside the promise chain.

bindings/node/test/clientEncryption.test.js
```diff
@@ -186,3 +189,143 @@ describe('ClientEncryption', function() {
...
+        expect(doc1)
+          .to.have.property('keyAltNames')
+          .that.includes.members(['foobar', 'fizzbuzz']);
+        expect(doc1).to.have.nested.property('_id.buffer');
+        expect(Buffer.compare(doc1._id.buffer, dataKey.buffer)).to.equal(0);
+        expect(doc2)
+          .to.have.property('keyAltNames')
+          .that.includes.members(['foobar', 'fizzbuzz']);
+        expect(doc2).to.have.nested.property('_id.buffer');
+        expect(Buffer.compare(doc2._id.buffer, dataKey.buffer)).to.equal(0);
```
mbroadst: nit: chai supports buffer comparison with `eql`: `expect(doc2._id.buffer).to.eql(dataKey.buffer)`

bindings/node/lib/clientEncryption.js
```diff
@@ -42,7 +42,34 @@ module.exports = function(modules) {
      */
     createDataKey(provider, options, callback) {
       if (typeof options === 'function') (callback = options), (options = {});
-      options = options || {};
+      options = Object.assign({}, options);
+
+      // To avoid using libbson inside the bindings, we pre-serialize
+      // any keyAltNames here.
+      if (options.keyAltNames) {
```
mbroadst: nit: does it make sense to move this all out into its own function to keep readability simple here?

bindings/node/test/clientEncryption.test.js
```diff
@@ -186,3 +189,143 @@ describe('ClientEncryption', function() {
...
+      return client.close();
+    }
+  });
+
+  function makeOptions(keyAltNames) {
+    return Object.assign({}, dataKeyOptions, { keyAltNames });
+  }
+
+  describe('errors', function() {
+    [42, 'hello', { keyAltNames: 'foobar' }, /foobar/].forEach(val => {
```
kevinAlbs: Oh, these are really nice and succinct tests.

bindings/node/lib/clientEncryption.js
daprahamian: I think I can refactor this into a helper.

bindings/node/lib/clientEncryption.js
daprahamian: The specific reason I am deleting this is b/c of cases where the property exists, but the value is falsey. Specifically, the following two cases:

```js
// undefined value
{ keyAltNames: undefined }

// null value
{ keyAltNames: null }
```

In both cases, the bindings `Nan::Has` call will return true, b/c the object *has* the property, even though the value is falsey. This would later trigger a hard-to-trace error inside of the binding, and might be confusing to users. IMO it makes sense to delete the property off of the copy so we do not think it is there when checking inside the C++ code.

To make it a little more clear, maybe I could change the check to:

```js
else if (options.keyAltNames == null)
```

so it is clearly for the case of `null` and `undefined`?

bindings/node/test/clientEncryption.test.js
daprahamian: This is an intentional change I made in 2c903d8ea45d2c1eb6af30b5929a480b14e28fae. `requirements.SKIP_LIVE_TESTS === process.env.MONGODB_NODE_SKIP_LIVE_TESTS`

bindings/node/test/clientEncryption.test.js
daprahamian: I specifically removed it b/c the `before` block inside `ClientEncryption` is stubbing out `stateMachine.prototype.kmsRequest`. The other two options would be:

a. move all of the other tests into another describe block that includes stubbing `stateMachine.prototype.kmsRequest` so that it only applies to the tests that use it
b. clearing the stub once we get to the `keyAltNames` tests.

I'm okay with any of the above approaches.

bindings/node/test/clientEncryption.test.js
daprahamian: really? does that also work with a `deepEquals`? I've been writing myself in knots for this :(

mbroadst: yeah!

bindings/node/test/clientEncryption.test.js
mbroadst: I'm okay with either as well

bindings/node/src/mongocrypt.cc
```diff
@@ -579,10 +536,57 @@ NAN_METHOD(MongoCrypt::MakeExplicitEncryptionContext) {
...
+            }
+
+            v8::Local<v8::Object> keyId =
+                Nan::To<v8::Object>(Nan::Get(options, KEY_ID_KEY).ToLocalChecked()).ToLocalChecked();
+            if (!node::Buffer::HasInstance(keyId)) {
+                Nan::ThrowTypeError("`keyId` must be a Buffer");
+                return;
+            }
+
+            mongocrypt_binary_t* binary = BufferToBinary(keyId);
```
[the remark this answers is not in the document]
mbroadst: agreed, these should all be wrapped in a smart pointer

bindings/node/lib/clientEncryption.js
mbroadst: I was mostly referring to calling it `options.keyAltName`, rather than just `keyAltName`. No need to hold up review for this nit though

bindings/node/src/mongocrypt.cc
```diff
@@ -546,15 +512,6 @@ NAN_METHOD(MongoCrypt::MakeEncryptionContext) {
...
-        v8::Local<v8::Object> options = Nan::To<v8::Object>(info[2]).ToLocalChecked();
-        auto result = setEncryptionOptions(context.get(), options);
-        if (!result.first) {
-            Nan::ThrowTypeError(result.second);
-            return;
-        }
-    }
-
     if (!mongocrypt_ctx_encrypt_init(
             context.get(), ns.c_str(), ns.size(), BufferToBinary(commandBuffer))) {
```
mbroadst: This line contains another leaked `mongocrypt_binary_t`

PR #53, opened 2019-08-13
NODE: chore(osx): fix build on osx in evergreen

daprahamian:
Fixes NODE-2110


bindings/node/binding.gyp
```diff
@@ -22,7 +22,7 @@
         'src/mongocrypt.cc'
       ],
       'xcode_settings': {
-        'MACOSX_DEPLOYMENT_TARGET': '10.7',
+        'MACOSX_DEPLOYMENT_TARGET': '10.14',
```
mbroadst: Did you try switching the DEVELOPER_DIR? We want to support as old a version of OSX as we can get away with. I’m not sure what that number is, but just checking that we looked into this

daprahamian: I tried exporting the `DEVELOPER_DIR` like the `mongo-swift-driver` does in the bash script, but it did not seem to fix the issue. Should I have been doing it from inside the `bindings.gyp`? that seems like it could be a little risky.

daprahamian: Also, I did some testing, and it looks like everything works if I set the target to osx 10.12, which came out nearly 3 years ago. Would that be far enough back?

PR #59, opened 2019-08-15
CryptOptions fix.

DmitryLukyanov:


bindings/cs/MongoDB.Crypt.Test/BasicTests.cs
```diff
@@ -37,13 +36,20 @@ public BasicTests(ITestOutputHelper output)
             _output = output;
         }
         
-        CryptOptions CreateOptions()
+        CryptOptions CreateOptions(bool withLocalCredentials = false)
```
jyemin: It's probably fine to always add local credentials.

bindings/cs/MongoDB.Crypt/CryptOptions.cs
```diff
@@ -14,28 +14,55 @@
...
 namespace MongoDB.Crypt
 {
     /// <summary>
-    /// Options to configure mongocrypt with
+    /// Options to configure mongocrypt with.
     /// </summary>
     public class CryptOptions
     {
-        public IKmsCredentials KmsCredentials { get; }
+        public Dictionary<KmsType, IKmsCredentials> KmsCredentialsOptions { get; } = new Dictionary<KmsType, IKmsCredentials>();
```
DmitryLukyanov: I open for another naming here.

bindings/cs/MongoDB.Crypt/MongoDB.Crypt.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>net452;netstandard1.5</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.5</TargetFrameworks>
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <PlatformTarget>x64</PlatformTarget>
```
DmitryLukyanov: Without this line, I had the error with code 193.

bindings/cs/MongoDB.Crypt/CryptOptions.cs
```diff
@@ -14,28 +14,55 @@
...
         public byte[] Schema { get; }
 
-        public CryptOptions(IKmsCredentials kmsCredentials)
+        public CryptOptions(IKmsCredentials kmsCredentials) : this(kmsCredentials, null)
```
jyemin: Remove the old constructors.  The driver shouldn't be using them as the driver API always takes a map of kms providers anyway.

bindings/cs/MongoDB.Crypt/MongoDB.Crypt.csproj
jyemin: I'm skeptical about this change.  @vincentkam, what do you think?

vincentkam: Hm.. I think this undoes the previous "bitness" PR that we had made. We purposely set the target to AnyCPU so that the driver, which also targets AnyCPU, can consume the wrapper without warnings. Under what circumstances are you still seeing error 193? The wrapper should in theory be aborting prior to this error happening through the bitness detection code [here](https://github.com/mongodb/libmongocrypt/blob/master/bindings/cs/MongoDB.Crypt/LibraryLoader.cs#L43)

bindings/cs/MongoDB.Crypt.Test/BasicTests.cs
DmitryLukyanov: Done

bindings/cs/MongoDB.Crypt/CryptOptions.cs
DmitryLukyanov: Done

bindings/cs/MongoDB.Crypt.Test/BasicTests.cs
```diff
@@ -212,6 +218,36 @@ public void EncryptExplicit()
...
+                encryptedBytes = encryptedBinary.ToArray(); // need to copy bytes out before the context gets destroyed
+            }
+
+            using (var cryptClient = CryptClientFactory.Create(CreateOptions(true)))
+            using (var context = cryptClient.StartExplicitDecryptionContext(encryptedBytes))
+            {
+                var (decryptedResult, _) = ProcessContextToCompletion(context);
+
+                decryptedResult.ToArray().Should().Equal(testData);
+            }
```
vincentkam: This looks good to me: we're not encrypting/decrypting with the local key but I don't think there's an easy way to create that test since we'd need the mock replies, but the driver side tests should cover this case

bindings/cs/MongoDB.Crypt.Example/Program.cs
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- 
+
```
DmitryLukyanov: There are two changes types in the below files
1. Replacing a single income parameter in `CryptOptions` on an array with one item (for example line `104`).
2. All other changes are autoformatting.

bindings/cs/MongoDB.Crypt/CryptOptions.cs
```diff
@@ -14,28 +14,55 @@
...
 
+        private void InitializeKmsCredentialsOptions(List<IKmsCredentials> kmsCredentialsOptions)
+        {
+            if (kmsCredentialsOptions != null && kmsCredentialsOptions.Count != 0)
+            {
+                foreach (var kmsCredential in kmsCredentialsOptions)
+                {
+                    if (KmsCredentialsOptions.ContainsKey(kmsCredential.KmsType))
+                    {
+                        KmsCredentialsOptions[kmsCredential.KmsType] = kmsCredential;
```
vincentkam: Is it possible to have more than one credential of a particular type? @jyemin

bindings/cs/MongoDB.Crypt.Test/BasicTests.cs
```diff
@@ -461,89 +429,89 @@ public void TestLocalKeyCreationStepwise()
...
-                    var binary = context.GetOperation();
-                    var doc = BsonUtil.ToDocument(binary);
-                    _output.WriteLine("ListCollections: " + doc);
-                    var reply = ReadJsonTestFile("collection-info.json");
-                    _output.WriteLine("Reply:" + reply);
-                    context.Feed(BsonUtil.ToBytes(reply));
-                    context.MarkDone();
-                    return (CryptContext.StateCode.MONGOCRYPT_CTX_NEED_MONGO_COLLINFO, binary, doc);
-                }
+                    {
```
vincentkam: Looks like a lot of whitespace changes made it in by accident?

bindings/cs/MongoDB.Crypt/MongoDB.Crypt.csproj
DmitryLukyanov: Not sure what happened, but when I rolled backed changes now, tests passed and error has gone

bindings/cs/MongoDB.Crypt.Test/BasicTests.cs
DmitryLukyanov: it's a default format from `vs` via `ctrl` + `k` + `d`. I guess it can be configured somehow or these lines can be manually reverted. Not sure that it worth it?

DmitryLukyanov:
ready for rereview.

bindings/cs/MongoDB.Crypt.Test/BasicTests.cs
vincentkam: Hm, if we're going to make style changes, how about we drop the braces that surround the contents of each "case" in the switch statement? That should make the autoformatter drop the spaces in front of each line.

vincentkam: Discussed offline, let's just leave it, I don't mind 😆

bindings/cs/MongoDB.Crypt/CryptOptions.cs
jyemin: No.

vincentkam: If it's illegal to have more than one credential of a particular type, should we throw an exception in the case that this happens? Or is it specified somewhere that we should just quietly replace the old value with the new value? @jyemin @DmitryLukyanov

DmitryLukyanov: As I see in `java` implementation, we just overwrite the previous value: https://github.com/mongodb/libmongocrypt/blob/master/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptOptions.java#L88 via a builder logic

DmitryLukyanov: So, maybe it's enough to have the current version?

bindings/cs/MongoDB.Crypt/CryptOptions.cs
vincentkam: Do we really need to expose a writeable dictionary? How about `IReadOnlyList<>`?

bindings/cs/MongoDB.Crypt/CryptOptions.cs
```diff
@@ -14,28 +14,47 @@
...
+    /// Options to configure mongocrypt with.
     /// </summary>
     public class CryptOptions
     {
-        public IKmsCredentials KmsCredentials { get; }
+        public Dictionary<KmsType, IKmsCredentials> KmsCredentialsOptions { get; } = new Dictionary<KmsType, IKmsCredentials>();
         public byte[] Schema { get; }
 
-        public CryptOptions(IKmsCredentials kmsCredentials)
+        public CryptOptions(IEnumerable<IKmsCredentials> kmsCredentialsOptions) : this(kmsCredentialsOptions, null)
```
vincentkam: Since it's invalid to have more than one type of kmsCredential,  how do you feel about the structuring our constructors to reflect this? One constructor for local credentials, one constructor for aws credentials, and one with both.

bindings/cs/MongoDB.Crypt/CryptOptions.cs
vincentkam: Right, Java does overwrite the previous value. They are using the builder pattern though, which makes it very explicit that you're "re-setting" an option if you specify more than one local or aws credential. Here, we're taking in a list of credentials, and it's non-obvious IMHO that only 1 of each type is allowed.  How do you feel about making this restriction more explicit? (see my comments for the constructors)

bindings/cs/MongoDB.Crypt/CryptOptions.cs
vincentkam: Alternatively, we could not expose a list at all and just expose two public properties: one for local credentials and one for AWS credentials

bindings/cs/MongoDB.Crypt/CryptOptions.cs
DmitryLukyanov: I thought about this solution, my idea was to avoid straight references on `IKmsCredentials` implementations. It will help us to leave this class without any changes in case if we need to add some other implementation of `IKmsCredentials`, but maybe it's not a big advantage, since in any case, we will need to make changes in this(wrapper) solution (because I don't think that we will add a new possible `IKmsCredentials` implementation in `driver` itself)

bindings/cs/MongoDB.Crypt/CryptOptions.cs
vincentkam: Per offline discussion, if we keep the dictionary: How about `KmsCredentialsMap`?

bindings/cs/MongoDB.Crypt/CryptOptions.cs
vincentkam: Resolved per offline discussion

bindings/cs/MongoDB.Crypt/MongoDB.Crypt.csproj
vincentkam: Resolved per offline discussion

PR #60, opened 2019-08-15
PYTHON-1963 Make pymongocrypt compatible with manylinux2010 releases

ShaneHarvey:
I've manually tested that the manylinux2010 wheel can be installed and used on RHEL 6.2, ubuntu 14.04, and amazon linux.

### DIFF
--- a/bindings/python/build-manylinux-wheel.sh
+++ b/bindings/python/build-manylinux-wheel.sh
@@ -1,0 +1,16 @@
+#!/bin/bash -ex
+cd /python
+
+# Compile wheel
+# https://github.com/pypa/manylinux/issues/49
+rm -rf build
+/opt/python/cp37-cp37m/bin/python setup.py bdist_wheel --universal
+
+# Audit wheels and write manylinux2010 tag
+for whl in dist/*.whl; do
+    # Skip already built manylinux2010 wheels.
+    if [[ "$whl" != *"manylinux2"* ]]; then
+        auditwheel repair $whl -w dist
+        rm $whl
+    fi
+done
--- a/bindings/python/pymongocrypt/binding.py
+++ b/bindings/python/pymongocrypt/binding.py
@@ -918,14 +918,16 @@ mongocrypt_setopt_crypto_hooks (mongocrypt_t *crypt,
 # build without relying on platform specific library path environment
 # variables, like LD_LIBRARY_PATH. For example:
 # export PYMONGOCRYPT_LIB='/path/to/libmongocrypt.so'
+# If the PYMONGOCRYPT_LIB is not set then load the embedded library and
+# fallback to the relying on a system installed library.
+_base = os.path.dirname(os.path.realpath(__file__))
 if sys.platform == 'win32':
-    # On windows the dll is named mongocrypt.dll
-    _path = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)), 'mongocrypt')
+    _path = os.path.join(_base, 'mongocrypt.dll')
+elif sys.platform == 'darwin':
+    _path = os.path.join(_base, 'libmongocrypt.dylib')
 else:
-    # On *nix the so is named libmongocrypt.so or libmongocrypt.dylib
-    _path = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)), 'libmongocrypt')
+    _path = os.path.join(_base, 'libmongocrypt.so')
+
 _PYMONGOCRYPT_LIB = os.environ.get('PYMONGOCRYPT_LIB')
 try:
     if _PYMONGOCRYPT_LIB:
--- a/bindings/python/release.sh
+++ b/bindings/python/release.sh
@@ -18,7 +18,6 @@ REVISION=latest
 # Build the manylinux2010 wheel
 rm -rf ./libmongocrypt
 curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/rhel-62-64-bit/master/${REVISION}/libmongocrypt.tar.gz
-curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/rhel-62-64-bit/master/latest/libmongocrypt.tar.gz
 mkdir libmongocrypt
 tar xzf libmongocrypt.tar.gz -C ./libmongocrypt
 NOCRYPTO_SO=libmongocrypt/nocrypto/lib64/libmongocrypt.so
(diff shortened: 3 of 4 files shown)


bindings/python/pymongocrypt/binding.py
```diff
@@ -918,14 +918,16 @@
...
-    _path = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)), 'mongocrypt')
+    _path = os.path.join(_base, 'mongocrypt.dll')
+elif sys.platform == 'darwin':
+    _path = os.path.join(_base, 'libmongocrypt.dylib')
 else:
-    # On *nix the so is named libmongocrypt.so or libmongocrypt.dylib
-    _path = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)), 'libmongocrypt')
+    _path = os.path.join(_base, 'libmongocrypt.so')
```
ShaneHarvey: We have to specify the full file name on Linux and macOS or else dlopen() does not find the library. For consistency I changed windows to use the full filename too.

bindings/python/setup.py
```diff
@@ -1,21 +1,32 @@
...
+# 	libmongocrypt.so
+# The wheel has to be platlib compliant in order to be repaired by auditwheel.
 try:
     from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
-    class bdist_wheel(_bdist_wheel):
 
-        def finalize_options(self):
-            _bdist_wheel.finalize_options(self)
-            self.root_is_pure = False
+    if sys.platform in ('win32', 'darwin'):
```
behackett: Why not put the entire try block inside of this if statement? You only need to import from wheel on those two platforms, nowhere else.

bindings/python/release.sh
```diff
@@ -0,0 +1,49 @@
...
+# Build the source dist first
+rm -rf build
+python3.7 setup.py sdist
+
+# The libmongocrypt git revision release to embed in our wheels.
+REVISION=latest
+
+# Build the manylinux2010 wheel
+rm -rf ./libmongocrypt
+curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/rhel-62-64-bit/master/${REVISION}/libmongocrypt.tar.gz
```
behackett: Why do you need to do this? Isn't the next line just going to overwrite it?

bindings/python/release.sh
```diff
@@ -0,0 +1,49 @@
...
+curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/rhel-62-64-bit/master/latest/libmongocrypt.tar.gz
+mkdir libmongocrypt
+tar xzf libmongocrypt.tar.gz -C ./libmongocrypt
+NOCRYPTO_SO=libmongocrypt/nocrypto/lib64/libmongocrypt.so
+chmod +x ${NOCRYPTO_SO}
+cp ${NOCRYPTO_SO} pymongocrypt/
+rm -rf ./libmongocrypt libmongocrypt.tar.gz
+
+docker run --rm -v `pwd`:/python quay.io/pypa/manylinux2010_x86_64 /python/build-manylinux-wheel.sh
```
behackett: What advantage does building in manylinux2010 provide over just running python setup.py bdist_wheel? I assume it has something to do with auditwheel?

bindings/python/build-manylinux-wheel.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -ex
+cd /python
+
+# Compile wheel
+# https://github.com/pypa/manylinux/issues/49
+rm -rf build
+/opt/python/cp37-cp37m/bin/python setup.py bdist_wheel --universal
```
behackett: cp37-cp37m? I guess they renamed things for 2010? Is there a non-'m' version?

bindings/python/release.sh
ShaneHarvey: Whoops, that's some leftover cruft. Fixed.

bindings/python/setup.py
ShaneHarvey: Done.

bindings/python/release.sh
ShaneHarvey: One advantage is that it allows us to create the manylinux wheel on macOS. Another is that we don't need to install auditwheel. But most importantly, I believe it's required to use the container. See this quote from [PEP-571](https://www.python.org/dev/peps/pep-0571/#compilation-of-compliant-wheels) (emphasis mine):
> Like manylinux1, the auditwheel tool adds manylinux2010 platform tags to linux wheels built by pip wheel or bdist_wheel **in a manylinux2010 Docker container**.

bindings/python/build-manylinux-wheel.sh
ShaneHarvey: These are the available versions:
```
+ ls -la /opt/python/
total 8
drwxr-xr-x 2 root root 4096 Aug 15 06:40 .
drwxr-xr-x 1 root root 4096 Aug 15 06:26 ..
lrwxrwxrwx 1 root root   34 Aug 15 06:28 cp27-cp27m -> /opt/_internal/cpython-2.7.16-ucs2
lrwxrwxrwx 1 root root   34 Aug 15 06:30 cp27-cp27mu -> /opt/_internal/cpython-2.7.16-ucs4
lrwxrwxrwx 1 root root   29 Aug 15 06:33 cp34-cp34m -> /opt/_internal/cpython-3.4.10
lrwxrwxrwx 1 root root   28 Aug 15 06:35 cp35-cp35m -> /opt/_internal/cpython-3.5.7
lrwxrwxrwx 1 root root   28 Aug 15 06:38 cp36-cp36m -> /opt/_internal/cpython-3.6.9
lrwxrwxrwx 1 root root   28 Aug 15 06:40 cp37-cp37m -> /opt/_internal/cpython-3.7.4
```

Note that since our package doesn't have any extensions I don't think it matters which python we use.

bindings/python/pymongocrypt/binding.py
behackett: Ack

bindings/python/build-manylinux-wheel.sh
behackett: Agreed. Thanks for clarifying.

bindings/python/release.sh
behackett: Cool. Thanks for clarifying.

PR #63, opened 2019-10-24
MONGOCRYPT-196 Allow compile.sh to run on Windows dev machines

vincentkam:
https://jira.mongodb.org/browse/MONGOCRYPT-196
~~https://evergreen.mongodb.com/version/5db2176dc9ec445c66adb3de~~
https://evergreen.mongodb.com/version/5db222b82a60ed51bcee3e9b

### DIFF
--- a/cmake/GetVersion.cmake
+++ b/cmake/GetVersion.cmake
@@ -39,7 +39,7 @@ function (GetVersion OUTVAR)
 
     # Otherwise, append our custom suffix -<date>-git<short hash>
     execute_process (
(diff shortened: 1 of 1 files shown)


vincentkam:
Update: So while the single quotes work locally, EG is distinctly unhappy. So I'm trying the second approach, which is to remove the offending `^{commit}` from the git command. If this isn't acceptable, no worries, I'll either use the CURRENT_VERSION file locally or I'll use a tweaked cmake file. 😸

vincentkam:
@rcsanchez97 What do you think? 😃

vincentkam:
Rebased, squashed, date amended and pushed!

PR #64, opened 2019-10-28
fix(bson): respect user bson options when using autoEncryption

daprahamian:
Adds a field where we can pass in user BSON serialization
options like promoteValues. We need to respect these options
when it comes to user data.

Fixes NODE-2278

~TODO: Find a way to test this.~ Tests in https://github.com/mongodb/node-mongodb-native/pull/2194

### DIFF
--- a/bindings/node/lib/stateMachine.js
+++ b/bindings/node/lib/stateMachine.js
@@ -69,8 +69,7 @@ module.exports = function(modules) {
    */
   class StateMachine {
     constructor(options) {
-      options = options || {};
-      this.userBsonOptions = options.userBsonOptions || undefined;
+      this.options = options || {};
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bindings/node/lib/stateMachine.js
```diff
@@ -68,6 +68,11 @@ module.exports = function(modules) {
    * @class StateMachine
    */
   class StateMachine {
+    constructor(options) {
```
mbroadst: why not just store `options` as `this.options` rather than making a sub document for `userBsonOptions`? I think there's value to a future where a flat options object is resolved from the connection string + MongoClient options, frozen, and then just passed all the way down the chain. Also this might eventually have options that aren't specifically bson options, you would have to have `userDecryptOptions`, `userEncryptOptions`, etc. I say it's turtles all the way down

PR #69, opened 2019-11-04
MONGOCRYPT-201 document installing on macOS/Windows

kevinAlbs:


README.md
```diff
@@ -35,6 +35,40 @@ make
...
+
+Or use cmake's `find_package`:
+```
+find_package (mongocrypt)
+# Then link against mongo::mongocrypt
+```
+
+## Installing libmongocrypt on Windows ##
+For Windows, there is a fixed URL to download the DLL and includes directory:
+https://s3.amazonaws.com/mciuploads/libmongocrypt/windows/latest_release/libmongocrypt.tar.gz
```
kevinAlbs: This is currently a prerelease. After the next tagged release this will be updated, and subsequently only tagged releases will update this.

kevinAlbs:
@divjotarora when you get a chance, can you confirm that the macOS/Windows instructions work okay for the Go driver?

README.md
```diff
@@ -35,6 +35,40 @@ make
 
 This builds libmongocrypt.dylib and test-libmongocrypt, in the cmake-build directory. Note, the `CMAKE_PREFIX_PATH` must include the path to the BSON library installation directory if it was not the defaults.  Also note that if your project will also dynamically link to the BSON library, you will need to add `-DENABLE_SHARED_BSON=ON` to the `cmake` command line.
 
+## Installing libmongocrypt on macOS ##
+First install [Homebrew according to its own instructions](https://brew.sh/). Using Homebrew, install the following dependencies.
+```
+brew install mongo-c-driver cmake
```
divjotarora: I had to run `brew link mongo-c-driver` as well but that may have been because I already had an older version of libmongoc installed.

README.md
```diff
@@ -35,6 +35,40 @@ make
...
+```
+git clone git@github.com:mongodb/libmongocrypt.git
+cd libmongocrypt
+cmake -DENABLE_SHARED_BSON=ON .
+cmake --build . --target install
+```
+
+Then, libmongocrypt can be used with pkg-config:
+```
+pkg-config libmongocrypt --libs --cflags
```
divjotarora: I assume this will install into a known location so `PKG_CONFIG_PATH` doesn't have to be updated?

README.md
```diff
@@ -35,6 +35,40 @@ make
...
+```
+
+Install the XCode Command Line Tools:
+```
+xcode-select --install
+```
+
+Then clone and build libmongocrypt:
+```
+git clone git@github.com:mongodb/libmongocrypt.git
```
rcsanchez97: I recall that during last week's meeting someone (I forget who) pointed out that cloning over https would prevent certain users getting errors if there was no ssh key configured in GitHub.  That point was made about the compile script that cloned and build libbson.  However, I think that for documentation meant for general public consumption (as opposed to active GitHub contributors), it is good practice to keep to https URLs.

README.md
kevinAlbs: My rough understanding is that homebrew won't overwrite files it didn't create. So, if you had installed libmongoc before (either by copying files, or building/installing from source). Then homebrew requires the explicit `link` after install to overwrite those files.

README.md
kevinAlbs: Yes, the default path it puts the pkg-config files in (`/usr/local/lib/pkgconfig/`) is searched for by default by pkg-config.

README.md
kevinAlbs: Good point, done.

kevinAlbs:
Updated.

ShaneHarvey:
Any reason we recommend building from source on macOS instead of downloading the prebuilt binary like we document for Windows?

kevinAlbs:
We decided to have users build libmongocrypt from source on macOS as a temporary measure until we have libmongocrypt in homebrew (https://jira.mongodb.org/browse/MONGOCRYPT-185). Since we don't have a package management solution for Windows, we decided to simply offer a fixed URL of the most recent release for the tarball.

ShaneHarvey:
I see. Are we still going to support the `macos` mciuploads urls? That's what we use in python to obtain the library to embedded in our release: https://github.com/mongodb/libmongocrypt/blob/pymongocrypt-0.1b2/bindings/python/release.sh#L54

behackett:
This change for the Go driver can't break the existing distribution model for other drivers.

kevinAlbs:
Yes, other drivers can continue to rely on the URLs to download artifacts from Evergreen tasks. This commit does not modify or remove any. It adds one upload task for Windows that is aimed for users to download from (we strip out everything except just the necessary .dll and headers).

PR #70, opened 2019-11-04
MONGOCRYPT-197 Evergreen task should create release mode dll

vincentkam:
Per offline discussions with Kevin, I've applied Kevin's changes and made the necessary changes to fix the C# bindings as well:

https://jira.mongodb.org/browse/MONGOCRYPT-197
https://jira.mongodb.org/browse/MONGOCRYPT-198
MONGOCRYPT-198 Update C# binding to account for changes in output structure
https://evergreen.mongodb.com/version/5dc080e030661573c0d12f9c

### DIFF
--- a/bindings/cs/MongoDB.Libmongocrypt/MongoDB.Libmongocrypt.csproj
+++ b/bindings/cs/MongoDB.Libmongocrypt/MongoDB.Libmongocrypt.csproj
@@ -28,28 +28,26 @@
 
   <PropertyGroup>
     <LibMongoCryptBinaries>../../../cmake-build/$(Configuration)</LibMongoCryptBinaries>
-    <LibMongoCryptBinaries Condition="'$(Configuration)'=='Debug' And '$(OS)' != 'Windows_NT'">../../../debug/</LibMongoCryptBinaries>
-    <LibMongoCryptBinaries Condition="'$(Configuration)'=='Release' And '$(OS)' != 'Windows_NT'">../../../release/</LibMongoCryptBinaries>
     <LibMongoCryptBinaries Condition="'$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT'">../../../cmake-build/RelWithDebInfo/</LibMongoCryptBinaries>
   </PropertyGroup>
 
   <ItemGroup>
-    <Content Include="$(LibMongoCryptBinaries)/libmongocrypt.so" Condition="'$(Configuration)'=='Debug' And Exists('$(LibMongoCryptBinaries)/libmongocrypt.so')">
+    <Content Include="$(LibMongoCryptBinaries)/libmongocrypt.so" Condition="Exists('$(LibMongoCryptBinaries)/libmongocrypt.so')">
       <Link>libmongocrypt.so</Link>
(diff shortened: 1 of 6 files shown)


vincentkam:
Ready for review!

bindings/cs/MongoDB.Libmongocrypt/MongoDB.Libmongocrypt.csproj
```diff
@@ -28,28 +28,26 @@
 
   <PropertyGroup>
     <LibMongoCryptBinaries>../../../cmake-build/$(Configuration)</LibMongoCryptBinaries>
-    <LibMongoCryptBinaries Condition="'$(Configuration)'=='Debug' And '$(OS)' != 'Windows_NT'">../../../debug/</LibMongoCryptBinaries>
```
vincentkam: I've verified that building locally in debug mode still results in the proper binaries being copied over (on Windows)

bindings/cs/MongoDB.Libmongocrypt/Package.include.template.csproj
```diff
@@ -27,6 +27,12 @@
       <PackagePath>x64/native/windows/mongocrypt.dll</PackagePath>
       <Pack>true</Pack>
     </Content>
+
+    <Content Include="@mongocrypt_BINARY_DIR@/RelWithDebInfo/mongocrypt.dll" Condition="Exists('@mongocrypt_BINARY_DIR@/RelWithDebInfo/mongocrypt.dll')">
```
vincentkam: This change is needed because for the build to succeed on Evergreen

bindings/cs/MongoDB.Libmongocrypt/MongoDB.Libmongocrypt.csproj
kevinAlbs: IIUC by default LibMongoCryptBinaries refers to `../../../cmake-build/$(Configuration)` unless Configuration is Release, then it refers to `../../../cmake-build/RelWithDebInfo`? And I assume that's needed because there's no notion of a RelWithDebInfo configuration?

bindings/cs/MongoDB.Libmongocrypt/Package.include.template.csproj
kevinAlbs: Should there also be a `Condition="'$(Configuration)'=='Release'"` and `Condition="'$(Configuration)'!='Release'"` one line 25 in case both dlls were built? With this change Evergreen will only be building RelWithDebInfo. But if we started building Debug in addition, perhaps this would choose the Debug build?

vincentkam: Hm, nice future proofing idea Kevin! 😸 I've added two extra commits needed for this idea to work (prior to these commits the C# binding was building and testing in debug mode with the RelWithDebInfo `mongocrypt.dll`

vincentkam:
High level LGTM received offline from Dima as well :)

vincentkam:
Rebased, date amended and pushed!

PR #71, opened 2019-11-05
NODE-2298: update documentation to synergize with driver docs

daprahamian:
Updates the documentation to correct a few items, and to
set up types so that they can be leveraged by the main
driver docs

Part of NODE-2298

### DIFF
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -10,40 +10,57 @@ module.exports = function(modules) {
   const cryptoCallbacks = require('./cryptoCallbacks');
 
   /**
-   * @typedef AutoEncrypter~AutoEncryptionExtraOptions
-   * @prop {string} [mongocryptdURI] overrides the uri used to connect to mongocryptd
-   * @prop {boolean} [mongocryptdBypassSpawn=false] if true, autoEncryption will not spawn a mongocryptd
-   * @prop {string} [mongocryptdSpawnPath] the path to the mongocryptd executable
-   * @prop {string[]} [mongocryptdSpawnArgs] command line arguments to pass to the mongocryptd executable
+   * Configuration options for a automatic client encryption.
+   *
+   * **NOTE**: Support for client side encryption is in beta. Backwards-breaking changes may be made before the final release.
+   *
+   * @typedef {Object} AutoEncrypter~AutoEncryptionOptions
+   * @property {MongoClient} [keyVaultClient] A `MongoClient` used to fetch keys from a key vault
+   * @property {string} [keyVaultNamespace] The namespace where keys are stored in the key vault
+   * @property {KMSProviders} [kmsProviders] Configuration options that are used by specific KMS providers during key generation, encryption, and decryption.
+   * @property {object} [schemaMap] A map of namespaces to a local JSON schema for encryption
+   * @property {boolean} [bypassAutoEncryption] Allows the user to bypass auto encryption, maintaining implicit decryption
+   * @property {AutoEncrypter~logger} [options.logger] An optional hook to catch logging messages from the underlying encryption engine
+   * @property {AutoEncrypter~AutoEncryptionExtraOptions} [extraOptions] Extra options related to the mongocryptd process
    */
 
   /**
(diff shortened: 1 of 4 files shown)


bindings/node/lib/autoEncrypter.js
```diff
@@ -10,40 +10,57 @@ module.exports = function(modules) {
...
 
   /**
-   * @typedef AutoEncrypter~AutoEncryptionExtraOptions
-   * @prop {string} [mongocryptdURI] overrides the uri used to connect to mongocryptd
-   * @prop {boolean} [mongocryptdBypassSpawn=false] if true, autoEncryption will not spawn a mongocryptd
-   * @prop {string} [mongocryptdSpawnPath] the path to the mongocryptd executable
-   * @prop {string[]} [mongocryptdSpawnArgs] command line arguments to pass to the mongocryptd executable
+   * Configuration options for a automatic client encryption.
+   *
+   * **NOTE**: Support for client side encryption is in beta. Backwards-breaking changes may be made before the final release.
```
mbroadst: I think we probably want to remove this comment about being in "beta" since we're going GA

bindings/node/lib/autoEncrypter.js
```diff
@@ -10,40 +10,57 @@ module.exports = function(modules) {
...
+   * @property {boolean} [mongocryptdBypassSpawn=false] If true, autoEncryption will not attempt to spawn a mongocryptd before connecting
+   * @property {string} [mongocryptdSpawnPath] The path to the mongocryptd executable on the system
+   * @property {string[]} [mongocryptdSpawnArgs] Command line arguments to use when auto-spawning a mongocryptd
+   */
+
+  /**
+   * @callback AutoEncrypter~logger
+   * @description A callback that is invoked with logging information from
+   * the underlying C++ Bindings.
+   * @param {AutoEncrypter~logLevel} level The level of logging. Valid values are 0 (Fatal Error), 1 (Error), 2 (Warning), 3 (Info), 4 (Trace)
```
mbroadst: nit: "Valid values.." seems redundant given you've provided a table below

bindings/node/lib/clientEncryption.js
```diff
@@ -196,12 +196,12 @@ module.exports = function(modules) {
...
     /**
-     * Explicitly encrypt a provided value. Note that either `options.dataKey` or `options.keyAltName` must
-     * be specified. Specifying both `options.dataKey` and `options.keyAltName` is considered an error.
+     * Explicitly encrypt a provided value. Note that either `options.keyId` or `options.keyAltName` must
+     * be specified. Specifying both `options.keyId` and `options.keyAltName` is considered an error.
      *
      * @param {*} value The value that you wish to serialize. Must be of a type that can be serialized into BSON
      * @param {object} options
-     * @param {ClientEncryption~dataKey} [options.dataKey] The Binary dataKey to use for encryption
+     * @param {ClientEncryption~dataKey} [options.keyId] The Binary dataKey to use for encryption
```
mbroadst: should this become `ClientEncryption~keyId` instead of `dataKey`? Also, what is the impetus for the change from `dataKey` => `keyId`?

daprahamian: The motivation was originally b/c the documentation incorrectly identified the field as `dataKey` instead of `keyId`. So at a bare minimum we need to properly document that the field is called `keyId`.

I'd be on board with renaming `ClientEncryption~dataKey` to `ClientEncryption~dataKeyId`

mbroadst: I think it looks strange to have `clientEncryption.createDataKey` return something called a `keyId`. I would expect it to return a "data key"

daprahamian: It creates a `dataKey`, and inserts that `dataKey` into the key vault namespace, and then returns the `_id` of that `dataKey`. It is explicitly not the `dataKey` b/c you could not insert that into the keyVaultNamespace and have it work.

mbroadst: ah right right. Okay, this LGTM as long as you make everything consistent (`dataKeyId` sounds good to me too).

daprahamian:
Updated to incorporated feedback and rebase

PR #76, opened 2019-11-15
NODE-2213: Document that using library directly requires injection

daprahamian:
Fixes NODE-2213

Ignore `README.md` as it is the compiled documentation and just represents the changes in `lib`.

Note that I used raw markdown for the examples on how to do injection, b/c the `@example` tag was not working when we attempted to use it in the greater API docs.

### DIFF
--- a/bindings/node/README.md
+++ b/bindings/node/README.md
@@ -29,14 +29,43 @@ npm test
 
 <dl>
 <dt><a href="#AutoEncrypter">AutoEncrypter</a></dt>
-<dd><p>An internal class to be used by the driver for auto encryption
-<strong>NOTE</strong>: Not meant to be instantiated directly, this is for internal use only.</p>
+<dd><p>An internal class to be used by the driver for auto encryption</p>
+<p><strong>NOTE</strong>: Not meant to be instantiated directly, this is for internal use only.</p>
 </dd>
 <dt><a href="#ClientEncryption">ClientEncryption</a></dt>
 <dd><p>The public interface for explicit client side encryption</p>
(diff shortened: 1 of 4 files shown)


bindings/node/README.md
```diff
@@ -165,6 +195,24 @@ the underlying C++ Bindings.
 ## ClientEncryption
 The public interface for explicit client side encryption
 
+**NOTE:** This class must be instantiated directly from `mongodb-client-encryption`.
+To do this, you must "inject" the `mongodb` package into `mongodb-client-encryption`:
+
+```js
+// Simply importing mongodb-client-encryption returns
```
mbroadst: I feel like this documentation is a bit obtuse. Why not just give them an example of using it:

```
const mongodb = require('mongodb');
const mdbce = require('mongodb-client-encryption')(mongodb);
const ClientEncryption = mdbce.ClientEncryption;
```

Also, I think we should update the injection to decorate the `mongodb` package, this is more in line with a future where we have something like `mongodb.use(require('client-encryption'))`. What about:

```
const mongodb = require('mongodb');
require('mongob-client-encryption')(mongodb);

const ClientEncryption = mongodb.ClientEncryption;
```

daprahamian: I think that is a neater API. My hesitations are:

1. It makes it confusing why we need to do that for `ClientEncryption` but not `AutoEncryption`
2. Hypothetically, ESM requires that module exports be immutable. So idk the repercussions of trying to augment the exported module either when running in ESM mode or when using a transpiler that enforces this.

mbroadst: How about if we introduce `mongodb.use` now then?

daprahamian:
Closing b/c we did #79 instead

PR #77, opened 2019-11-18
fix static builds on windows

mbroadst:
Since kms_message is built as a static library, it needs to be
included on the link line for all statically compiled targets that
need to include it. Additionally, `KMS_MSG_STATIC` needed to be
defined during `mongocrypt_static` compilation in order to avoid
prefixing symbols which assumed a dynamic build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ target_link_libraries (mongocrypt PRIVATE kms_message_static)
 
 generate_export_header(mongocrypt EXPORT_FILE_NAME src/mongocrypt-export.h BASE_NAME mongocrypt )
 
-add_library (mongocrypt_static STATIC ${MONGOCRYPT_SOURCES} $<TARGET_OBJECTS:kms_message_obj>)
+add_library (mongocrypt_static STATIC ${MONGOCRYPT_SOURCES})
 target_include_directories (mongocrypt_static PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/kms-message/src")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -178,15 +178,15 @@ target_link_libraries (mongocrypt PRIVATE kms_message_static)
...
 
-add_library (mongocrypt_static STATIC ${MONGOCRYPT_SOURCES} $<TARGET_OBJECTS:kms_message_obj>)
+add_library (mongocrypt_static STATIC ${MONGOCRYPT_SOURCES})
 target_include_directories (mongocrypt_static PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/kms-message/src")
 target_include_directories (mongocrypt_static PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>)
 target_include_directories (mongocrypt_static PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/src")
 target_include_directories (mongocrypt_static PRIVATE ${BSON_INCLUDES})
 target_compile_definitions (mongocrypt_static PRIVATE ${BSON_DEFINITIONS})
 target_link_libraries (mongocrypt_static PRIVATE ${BSON_TARGET})
 target_link_libraries (mongocrypt_static PRIVATE ${CMAKE_THREAD_LIBS_INIT})
```
markbenvenuto: Add a line here so that kms_message_static is included in mongocrypt_static:
`target_link_libraries (mongocrypt_static PRIVATE kms_message_static)`
Then you will only have to include mongocrypt_static in consumers.

Then you can remove the separate inclusion of `kms_message_static` below.

mbroadst: done

PR #78, opened 2019-11-18
MONGOCRYPT-211 support linking against windows static runtime

mbroadst:
This allows dependent projects to optionally enable overriding the
default MSVC runtime (`/MD` for the dynamic mulitthreaded) to use
the static runtime (`/MT`)


kevinAlbs:
Committed in https://github.com/mongodb/libmongocrypt/commit/e48bacef2f12fea011ef549b3297cfa8219aad13

PR #79, opened 2019-11-18
NODE-2339: change APIn to remove mandatory injection

daprahamian:
Rather than require users inject mongodb into
mongodb-client-encryption we assume that require('mongodb')
resolves to the user mongodb package. If users need to ensure
their mongodb version is used, they can use the extension function
to achieve the old injection effect.

Note: Uses getters to avoid a circular load call.

### DIFF
--- a/bindings/node/index.js
+++ b/bindings/node/index.js
@@ -1,102 +1,8 @@
 'use strict';
 
-/**
- * @module mongodbClientEncryption
- * @description An extension that can be installed to enable MongoDB Client-Side Encryption.
- * @example
- * // Using AutoEncryption
- * 
- * const mongodb = require('mongodb');
- * 
- * // Simply create a MongoClient with the option `autoEncryption`. If
- * // mongodb-client-encryption is installed, it will be automatically
- * // loaded. Your encrypted client will automatically encrypt and decrypt
- * // according to the schema provided by the server.
- * const encryptedClient = new mongodb.MongoClient('mongodb://localhost:27017', {
- *   useNewUrlParser: true,
- *   useUnifiedTopology: true,
- *   autoEncryption: {
- *     kmsProviders: {
- *       aws: {
- *         accessKeyId: AWS_ACCESS_KEY,
- *         secretAccessKey: AWS_SECRET_KEY
- *       }
- *     }
- *   }
- * });
- * 
- * main();
- * 
- * async function main() {
- *   try {
- *     await client.connect();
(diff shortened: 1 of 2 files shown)


bindings/node/index.js
```diff
@@ -106,6 +12,30 @@ module.exports = function(mongodb) {
   return {
     AutoEncrypter: modules.autoEncrypter.AutoEncrypter,
     ClientEncryption: modules.clientEncryption.ClientEncryption,
-    MongoCryptError: require('./lib/common').MongoCryptError
+    MongoCryptError
   };
-};
+}
+let _module;
+function loadModule() {
```
mbroadst: how is `loadModule` different than `memo(() => extension(require('mongodb'))`?

bindings/node/index.js
```diff
@@ -15,27 +23,20 @@ function extension(mongodb) {
...
-  get: memo(() => loadModule().AutoEncrypter)
-});
-Object.defineProperty(exports, 'ClientEncryption', {
-  enumerable: true,
-  get: memo(() => loadModule().ClientEncryption)
-});
+module.exports = {
+  extension,
+  MongoCryptError,
+  get AutoEncrypter() {
```
mbroadst: cool, didn't know this worked outside of a `class` definition. This seems relatively complicated with all the deletion, why not just have the getter returned a cached copy? It seems like you already cache it, so can't the getter just be:
```
const m = loadDefaultModule();
return m.AutoEncrypter;
```

daprahamian: Eh, I wanted to get rid of the getter. Did a benchmark and it is *technically* better to reset it to a prop:

```js

{ prop:
   { moe: 1.0856058264241745e-10,
     rme: 1.589729254384141,
     sem: 5.5388052368580335e-11,
     deviation: 5.2252984098655e-10,
     mean: 6.828872422334184e-9,
     variance: 2.730374347214292e-19 },
  getter:
   { moe: 9.04029618171838e-11,
     rme: 0.87392302730372,
     sem: 4.6123960110808064e-11,
     deviation: 4.448033630226029e-10,
     mean: 1.034449934293418e-8,
     variance: 1.9785003175621743e-19 } }
```

but honestly thats a micro-optimization. We can just leave them as getters if you want.

mbroadst: the getter feels less complicated to me, my priority is readability for the future so I'd prefer to just leave them as getters

PR #80, opened 2019-11-21
Release pymongocrypt 0.1b3

ShaneHarvey:


bindings/python/release.sh
```diff
@@ -13,7 +13,7 @@ set -o xtrace   # Write all commands first to stderr
 set -o errexit  # Exit the script with error if any of the commands fail
 
 # The libmongocrypt git revision release to embed in our wheels.
-REVISION=$(git rev-parse 1.0.0-beta5)
+REVISION=$(git rev-list -n 1 1.0.0)
```
ShaneHarvey: rev-parse returns the SHA of the tag object for an annotated tag which is not what we want. `rev-list -n 1` returns the SHA of the commit for both annotated and unannotated tags. I noticed this because the latest tag (1.0.0) is annotated and previous tags were not.

ShaneHarvey:
Released: https://pypi.org/project/pymongocrypt/0.1b3/

PR #82, opened 2019-11-29
MONGOCRYPT-212 add null check for `kms_request_get_canonical`

mbroadst:
This method can return `NULL`, so a null check is required in order to prevent segfaults.

**NOTE:** sorry for the kind of cheesy test case. I think the unit test belongs in `kms_message` but there didn't seem to be existing tests there outside of the AWS test suite, and this was much easier for me to add.


mbroadst:
@kevinAlbs are you fine with me merging this then?

kevinAlbs:
@mbroadst Yes

PR #86, opened 2020-01-14
Java: Added OSGI support

rozza:
Updated the java build to include the bnd plugin
and configured so the pom produces a META-INF/MANIFEST.MF

JAVA-3575

https://evergreen.mongodb.com/version/5e1dd218d6d80a6b51b895aa

```
$ cat META-INF/MANIFEST.MF
Manifest-Version: 1.0
Automatic-Module-Name: com.mongodb.crypt.capi
Bnd-LastModified: 1579015065704
Build-Version: 1.0.0-beta3-235-g9f6fd7c-dirty
Bundle-ManifestVersion: 2
Bundle-Name: MongoCrypt
Bundle-SymbolicName: com.mongodb.crypt.capi
Bundle-Version: 1.0.0.beta3-235-g9f6fd7c-dirty
Created-By: 11.0.5 (Private Build)
Export-Package: com.mongodb.crypt.capi;uses:="org.bson";version="1.0.0
 .beta3-235-g9f6fd7c-dirty"
Import-Package: org.bson;version="[3.10,5.0)",org.bson.assertions;vers
 ion="[3.10,5.0)",org.bson.codecs;version="[3.10,5.0)",org.bson.codecs
 .configuration;version="[3.10,5.0)",org.bson.io;version="[3.10,5.0)"
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Tool: Bnd-4.3.1.201911131708

```

### DIFF
--- a/bindings/java/mongocrypt/build.gradle.kts
+++ b/bindings/java/mongocrypt/build.gradle.kts
@@ -37,6 +37,7 @@ plugins {
     `maven-publish`
     signing
     id("de.undercouch.download").version("3.4.3")
+    id("biz.aQute.bnd.builder").version("4.3.1")
 }
 
 repositories {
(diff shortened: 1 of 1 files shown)


bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -211,14 +227,7 @@ publishing {
 
             artifact(tasks["sourcesJar"])
             artifact(tasks["javadocJar"])
-            versionMapping {
```
rozza: This removes the resolved version mapping that was the cause of the pom.xml containing the resolved issue. Because it now uses the default declared version resolution, there is no need to modify the pom.xml.

https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:resolved_dependencies

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -240,20 +249,6 @@ publishing {
                     connection.set("scm:https://github.com/mongodb/libmongocrypt")
                     developerConnection.set("scm:git@github.com:mongodb/libmongocrypt")
                 }
-                withXml {
```
rozza: This modification is no longer needed due to using the declared version rather than the resolved version(s).

rozza:
Thanks for the reviews - looks like we will need a change to the java driver for driver-core to work as expected see: rozza/mongo-java-driver#351

I've added @jstewart-mongo to the reviewers for this change as well.

PR #93, opened 2020-06-04
update gitignore

kevinAlbs:
Sync up with mongo-c-driver's .gitignore. This should prevent build artifacts from getting tracked in git in case of an in-source build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,23 @@ Win32/*
 cmake_install.cmake
 x64/*
 Testing/
+CPackConfig.cmake
+CPackSourceConfig.cmake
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -22,6 +22,23 @@ Win32/*
...
+.vs
+install_manifest.txt
+
+# Build artifacts.
+*.a
+*.dylib
+*.gcda
+*.gcno
+*.gz
+*.o
```
jmikola, line 38: Missing `*.lo` :)

kevinAlbs: 🤦

PR #95, opened 2020-06-17
MONGOCRYPT-258 Simplify python binding test setup on Windows

ShaneHarvey:
https://jira.mongodb.org/browse/MONGOCRYPT-258

We can simplify the python binding test setup on Windows now that we test with windows-64-vsMulti-small (PYTHON-2169) using a real Windows toolchain. In particular we no longer need to use "venv" or manually upgrade pip/setuptools.

### DIFF
--- a/bindings/python/.evergreen/test.sh
+++ b/bindings/python/.evergreen/test.sh
@@ -11,55 +11,12 @@ MONGOCRYPT_DIR="$MONGOCRYPT_DIR"
 if [ "Windows_NT" = "$OS" ]; then # Magic variable in cygwin
     PYMONGOCRYPT_LIB=${MONGOCRYPT_DIR}/nocrypto/bin/mongocrypt.dll
     export PYMONGOCRYPT_LIB=$(cygpath -m $PYMONGOCRYPT_LIB)
-    # We need to create virtualenvs to install pymongocrypt's
-    # cffi, cryptography, and pymongo dependencies correctly from
-    # wheels. Otherwise, setup.py attempts and fails to install
-    # from the source distributions (.tar.gz).
-    VIRTUALENV="/cygdrive/c/python/Python27/python -m virtualenv"
-    $VIRTUALENV -p C:/python/Python27/python.exe venv27
-    PYTHON27="$(pwd)/venv27/Scripts/python"
-    # Upgrade pip to install the cryptography wheel and not the tar.
-    # <20.1 because 20.0.2 says a future release may drop support for 2.7.
-    $PYTHON27 -m pip install --upgrade 'pip<20.1'
-    # Upgrade setuptools because cryptography requires 18.5+.
-    # <45 because 45.0 dropped support for 2.7.
-    $PYTHON27 -m pip install --upgrade 'setuptools<45'
-    $PYTHON27 -m pip install . pymongo
-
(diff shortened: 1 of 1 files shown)


bindings/python/.evergreen/test.sh
```diff
@@ -11,55 +11,12 @@ MONGOCRYPT_DIR="$MONGOCRYPT_DIR"
...
-    $PYTHON38 -m pip install . pymongo
-
-    PYTHONS=("python" \
-             "$PYTHON27" \
-             "$PYTHON34" \
-             "$PYTHON35" \
-             "$PYTHON36" \
-             "$PYTHON37" \
-             "$PYTHON38")
+    PYTHONS=("C:/python/Python27/python.exe" \
```
prashantmital, line 14: So now we just install everything in the base python environment? Shouldn't we use a virtualenv for this? (I almost feel like I saw a ticket about this recently).

prashantmital:
Also, why do we only run the Python tests on such few platforms? Is there something preventing us from testing this on Ubuntu 18?

bindings/python/.evergreen/test.sh
ShaneHarvey: We don't install anything on the system. We just let `python setup.py test` fetch and locally install the dependencies automagically.

prashantmital: Right. But I think this does mean that the deps are installed system-wide and _might_ still be present when/if this instance gets picked from the pool again to run the test-suite. Previously, we used a Python binary that resided inside a venv so cryptography and cffi were being installed in that virtualenv which would get wiped out with the rest of the test-run data (as each run gets a unique directory under /data/mci/... in which to run).

In any case, I doubt this is an actual problem but I'm just worried that we might run into some issues that are hard to debug/reproduce.

prashantmital: Right. But I think this does mean that the deps are installed system-wide and _might_ still be present when/if this instance gets picked from the pool again to run the test-suite. Previously, we used a Python binary that resided inside a venv so cryptography and cffi were being installed in that virtualenv which would get wiped out with the rest of the test-run data (as each run gets a unique directory under /data/mci/... in which to run).

In any case, I doubt this is an actual problem but I'm just worried that we might run into some issues that are hard to debug/reproduce.

prashantmital: I think if you'd like to ensure everything runs in a virtualenv, we can define the `$VIRTUALENV` variable for each platform and the `$BIN_DIR` variable (`Scripts` for WINDOWS and `bin` for other platforms) in the if...elif block and then modify the last codeblock as follows:

```
for PYTHON_BINARY in "${PYTHONS[@]}"; do
    # Clear cached eggs for different python versions.
    rm -rf .eggs
    rm -rf venv
    $VIRTUALENV -p $PYTHON_BINARY venv
    venv/$BIN_DIR/python -c 'import sys; print(sys.version)'
    venv/$BIN_DIR/python setup.py test
```

ShaneHarvey: I don't understand how polluting the system would be a concern with `python setup.py test` because it does not install anything on the system. If a library is needed it is fetched and cached in the local directory. If you think we should use a virtualenv for some other reason please open a new ticket.

prashantmital:
LGTM (in case you were waiting for it).

PR #98, opened 2020-08-05
MONGOCRYPT-261 Build tests on musl

markbenvenuto:
stdout is a readonly with musl libc which is used by Alpine Linux.

There is no define check for MUSL (i.e. there is no equivalent to __GLIBC__). As a result, I defined this test as only running on APPLE and GLIBC. This will miss Freebsd, etc but we will still get coverage on our core platforms.

### DIFF
--- a/test/test-mongocrypt-log.c
+++ b/test/test-mongocrypt-log.c
@@ -96,7 +96,7 @@ _test_trace_log (_mongocrypt_tester_t *tester)
    mongocrypt_destroy (crypt);
 }
 
-#ifndef _WIN32
+#if defined(__GLIBC__) || defined(__APPLE__)
 static void
 _test_no_log (_mongocrypt_tester_t *tester)
 {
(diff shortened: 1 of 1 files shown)


behackett:
@markbenvenuto how are you testing with musl?

behackett:
Could we get an image in Evergreen with one of these distros configured with musl?  https://wiki.musl-libc.org/projects-using-musl.html#Linux-distros-shipping-musl-as-an-optional-package

markbenvenuto:
To test, I was using podman and a Alpine Linux container locally. I did not port over the build scripts (there is some gnu-isms in the build scripts).

Most of these packages are only needed for building the server.
```
FROM alpine:latest
RUN apk add py3-pip gcc scons python3 git py3-psutil py3-yaml g++ openssl-dev curl-dev py3-cheetah linux-headers make cmake openssh

RUN apk add bash

# libmongocrypt compile scripts use git+ssh to grab c-driver source
COPY id_rsa /root/.ssh/
COPY id_rsa.pub /root/.ssh/

RUN apk add xz-dev netsnmp-dev cyrus-sasl-dev openldap-dev krb5-dev
```

PR #101, opened 2020-09-08
make build work with space in source tree path

rcsanchez97:
Based on: https://developer.mongodb.com/community/forums/t/mongo-c-driver-error-when-executing-a-build/8852

Full Evergreen patch build: https://spruce.mongodb.com/version/5f57aa6c61837d5a554d1f87/tasks


kevinAlbs (approved):
LGTM! Similarly, I suggest noting this in a Jira ticket as well.

rcsanchez97:
@kevinAlbs The Jira ticket is MONGOCRYPT-280.

PR #102, opened 2020-09-09
MONGOCRYPT-252 package for Debian

rcsanchez97:
Full Evergreen patch build (including new Linux Distro Packaging variant with debian-package-build task): https://spruce.mongodb.com/version/5f52482657e85a7177aab7de/tasks

Note that my plan is that once this merged, I will push a new branch with the same changes.  The new branch will be called `debian/1.0.4` and will start from the `1.0.4` tag.


.evergreen/config.yml
```diff
@@ -115,6 +115,11 @@ functions:
           pkg_version=$mongocrypt_version
           tar -zcv -C install --transform="s|^libmongocrypt/|libmongocrypt-$pkg_version/|" --exclude=nocrypto --exclude=sharedbson -f libmongocrypt-$pkg_version.tar.gz libmongocrypt
           cd libmongocrypt/etc/
+          # The files from libmongocrypt/debian/ are the official maintainer scripts,
+          # but libmongocrypt/etc/debian/ contains a a few custom scripts that are
```
kevinAlbs, line 119:
```suggestion
          # but libmongocrypt/etc/debian/ contains a few custom scripts that are
```

.evergreen/debian_package_build.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/sh
+
+#
+# Test mongocxx's Debian packaging scripts.
```
kevinAlbs, line 4:
```suggestion
# Test libmongocrypt's Debian packaging scripts.
```

debian/control
```diff
@@ -0,0 +1,37 @@
+Source: libmongocrypt
+Priority: optional
+Maintainer: Mongo C Driver Team <mongo-c-driver-debian@googlegroups.com>
+Uploaders: Kevin Albertson <kevin.albertson@mongodb.com>,
+           Samantha Ritter <samantha.ritter@mongodb.com>,
+           Roberto C. Sanchez <roberto@connexer.com>
```
kevinAlbs, line 6: Can you also add Clyde?

etc/debian/control
```diff
@@ -1,8 +1,10 @@
 Source: libmongocrypt
 Priority: optional
-Maintainer: Roberto C. Sanchez <roberto@connexer.com>
-Build-Depends:
-Standards-Version: 4.1.3
+Maintainer: Mongo C Driver Team <mongo-c-driver-debian@googlegroups.com>
+Uploaders: Kevin Albertson <kevin.albertson@mongodb.com>,
+           Samantha Ritter <samantha.ritter@mongodb.com>,
+           Roberto C. Sanchez <roberto@connexer.com>
```
kevinAlbs, line 6: Here too, add Clyde so we have all of the C/C++ team listed.

kevinAlbs (commented):
Looks good from what I can tell, just a couple minor things.

.evergreen/config.yml
rcsanchez97: Done

debian/control
rcsanchez97: Done

etc/debian/control
rcsanchez97: Done

PR #105, opened 2020-09-18
Update suse12-test to suse12-sp5-small

kevinAlbs:
Per the email "Evergreen SUSE/SLES 12 Distro update to SP5" (I just forwarded it to you), this updates the distro.


rcsanchez97 (approved):
LGTM

PR #106, opened 2020-09-21
MONGOCRYPT-265 Add RSA signing with openssl

kevinAlbs:


kms-message/test/test_kms_request.c
```diff
@@ -1025,6 +1025,93 @@ kms_request_validate_test (void)
...
+ * test-private-key.pem generated as follows: openssl genpkey -out
+ * test-private-key.pem -algorithm RSA -pkeyopt rsa_keygen_bits:2048
+ *
+ * It was used to generate a test signature of the string "data to sign" with:
+ * openssl dgst -sign test-private-key.pem -sha256 signme.txt | base64
+ */
+static void
+kms_signature_test (void)
+{
+   const char *private_key_b64 =
```
kevinAlbs, line 1040: In addition to this example key I generated, I tested signing a JWT I created with my own GCP credentials.

kms-message/src/kms_crypto_libcrypto.c
```diff
@@ -91,4 +91,48 @@ kms_sha256_hmac (void *unused_ctx,
...
+   ctx = EVP_MD_CTX_new ();
+   pkey = d2i_PrivateKey (EVP_PKEY_RSA,
+                          NULL,
+                          (const unsigned char **) &private_key,
+                          private_key_len);
+   if (!pkey) {
+      goto cleanup;
+   }
+
+   ret = EVP_DigestSignInit (ctx, NULL, EVP_sha256 (), NULL /* engine */, pkey);
```
markbenvenuto, line 116: Looks like the default is RSA_PKCS1_PADDING which made this really easy.

markbenvenuto (approved):

kms-message/src/kms_crypto_libcrypto.c
kevinAlbs: Yeah, if we needed a different padding, I think I could set the padding type as a property of the EVP_PKEY with `EVP_PKEY_CTX_set_rsa_padding`.

PR #109, opened 2020-09-25
MONGOCRYPT-282 support chunked transfer encoding

kevinAlbs:
GCP HTTP responses used `Transfer-Encoding: chunked`. This adds support in the tiny kms-message HTTP parser.

### DIFF
--- a/kms-message/src/kms_response_parser.c
+++ b/kms-message/src/kms_response_parser.c
@@ -57,6 +57,7 @@ kms_response_parser_wants_bytes (kms_response_parser_t *parser, int32_t max)
    case PARSING_CHUNK_LENGTH:
       return max;
    case PARSING_CHUNK:
+      /* add 2 for trailing \r\n */
       return (parser->chunk_size + 2) -
              ((int) parser->raw_response->len - parser->start);
    case PARSING_BODY:
--- a/kms-message/src/kms_response_parser.c
+++ b/kms-message/src/kms_response_parser.c
@@ -6,6 +6,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "hexlify.h"
+
 /* destroys the members of parser, but not the parser itself. */
 static void
 _parser_destroy (kms_response_parser_t *parser)
@@ -50,6 +52,11 @@ kms_response_parser_wants_bytes (kms_response_parser_t *parser, int32_t max)
    case PARSING_STATUS_LINE:
    case PARSING_HEADER:
       return max;
+   case PARSING_CHUNK_LENGTH:
+      return max;
+   case PARSING_CHUNK:
+      return (parser->chunk_size + 2) -
+             ((int) parser->raw_response->len - parser->start);
    case PARSING_BODY:
       KMS_ASSERT (parser->content_length != -1);
       return parser->content_length -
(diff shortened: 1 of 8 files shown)


kms-message/test/example-chunked-response.bin
```diff
@@ -0,0 +1,20 @@
...
+X-Frame-Options: SAMEORIGIN
+X-Content-Type-Options: nosniff
+Alt-Svc: h3-Q050=":443"; ma=2592000,h3-29=":443"; ma=2592000,h3-27=":443"; ma=2592000,h3-T051=":443"; ma=2592000,h3-T050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"
+Accept-Ranges: none
+Vary: Origin,Accept-Encoding
+Connection: close
+Transfer-Encoding: chunked
+
+105
+{"access_token":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","expires_in":3599,"token_type":"Bearer"}
```
kevinAlbs, line 18: This is the form of a real GCP response, with the access token bytes replaced.

kms-message/test/test_kms_request.c
```diff
@@ -964,6 +937,83 @@ kms_response_parser_test (void)
...
 }
 
+typedef struct {
+   const char *filepath;
+   const char *expected_body;
+   int max_to_read;
+   int expected_status;
+} parser_testcase_t;
+
+/* File should have \r\n line endings (use /etc/rewrite.py) */
```
markbenvenuto, line 947: Who is supposed to call rewrite.py? You can use git attributes to control how the file is checked out - https://www.git-scm.com/docs/gitattributes

kms-message/src/kms_response_parser.c
```diff
@@ -254,6 +298,29 @@ kms_response_parser_feed (kms_response_parser_t *parser,
...
+            if (!parser->response->body) {
+               parser->response->body = kms_request_str_new ();
+            }
+            kms_request_str_append_chars (parser->response->body,
+                                          raw->str + parser->start,
+                                          parser->chunk_size);
+            curr = parser->start + parser->chunk_size + 2;
+            parser->start = curr;
+            if (parser->chunk_size == 0) {
+               printf ("done\n");
```
markbenvenuto, line 314: Remove

kms-message/src/kms_response_parser.c
```diff
@@ -101,6 +111,16 @@ _parse_int_from_view (const char *str, int start, int end, int *result)
    return ret;
 }
 
+static bool
+_parse_hex_from_view (const char *str, int start, int end, int *result)
```
markbenvenuto, line 115: Just take length instead of (start, end)

kms-message/test/test_kms_request.c
kevinAlbs: It's in etc, so just a development script for producing `\r\n` on non Windows. The default seems to preserve the `\r\n` (at least for .bin files).

kms-message/src/kms_response_parser.c
kevinAlbs: Gah, done.

kms-message/src/kms_response_parser.c
kevinAlbs: Done.

markbenvenuto (approved):

PR #111, opened 2020-09-25
MONGOCRYPT-277 & MONGOCRYPT-278

rcsanchez97:
Full Evergreen patch build: https://spruce.mongodb.com/version/5f6dff15850e61660141fae1/tasks


divjotarora:
I'll set aside some time to test static linking for the Go driver using this branch next Monday.

.evergreen/pkgconfig-tests.sh
```diff
@@ -0,0 +1,103 @@
...
+cd $pkgconfig_tests_root
+rm -rf libmongocrypt-cmake-build install/libmongocrypt
+mkdir -p libmongocrypt-cmake-build
+
+# Build libmongocrypt, dynamic linking against libbson
+cd $pkgconfig_tests_root/libmongocrypt-cmake-build
+PREFIX_PATH="$(system_path $pkgconfig_tests_root/install)"
+INSTALL_PATH="$(system_path $pkgconfig_tests_root/install/libmongocrypt)"
+SRC_PATH="$(system_path $libmongocrypt_root)"
+$CMAKE -DENABLE_SHARED_BSON=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS -DCMAKE_C_FLAGS="-fPIC" -DCMAKE_PREFIX_PATH="$PREFIX_PATH" -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" "$SRC_PATH"
```
kevinAlbs, line 84: If this is dynamically linking against libbson, this should be `-DENABLE_SHARED_BSON=ON`, right?

CMakeLists.txt
```diff
@@ -323,19 +329,22 @@ set (MONGOCRYPT_BUILD_VERSION ${BUILD_VERSION})
 configure_file (src/mongocrypt.h.in src/mongocrypt.h)
 
 set (PROJECT_VERSION "${BUILD_VERSION}")
-# If libbson is statically linked, don't require a dependency.
+set (PROJECT_DESCRIPTION "The libmongocrypt client-side field level encryption library.")
+set (PKG_CONFIG_REQUIRES "libbson-static-1.0")
+set (PKG_CONFIG_STATIC_REQUIRES "libbson-static-1.0")
```
kevinAlbs, line 334: Ah, I think this was my initial understanding of static/dynamic dependencies for the four cases of linking to libmongocrypt.

1. static link libmongocrypt, and libmongocrypt static links libbson
2. static link libmongocrypt, and libmongocrypt dynamic links libbson
3. dynamic link libmongocrypt, and libmongocrypt static links libbson
4. dynamic link libmongocrypt, and libmongocrypt dynamic links libbson

For (1), a consumer only needs to link to libmongocrypt-static.a.
For (2), a consumer needs to link to libmongocrypt-static.a and libbson.so (since static libraries don't have relocation info).
For (3) and (4), a consumer only needs to link to libmongocrypt.so.

But given the report and this change, that seems incorrect. If libmongocrypt static links libbson, consumers that are static linking libmongocrypt still need to static link libbson?

.evergreen/pkgconfig-tests.sh
rcsanchez97: Quite right.  Looks like a copy pasta error on my part.

CMakeLists.txt
rcsanchez97: Your (1) is incorrect.  Regardless of whether libmongocrypt was told to dynamically or statically link libbson, consumers of libmongocrypt will also need to specify linkage to libbson (even if the consumer does not directly use libbson).  A static library is nothing more than a collection of object files in a single archive.  It knows nothing of how to link to any other libraries, thus requiring any artifact that links to that static library to also link to anything else that the static library may need.  That is really the fundamental reason why MONGOCRYPT-277 happened in the first place; the pkg-config script did not specify the other things needed by the static libmongocrypt so that consumers of static libmongocrypt could tell the linker about them.

Your (2) is right, but for the wrong reason.  The comments above apply here as well; it is not the absence of relocation info, but rather the absence of linkage info.

For (3) and (4), you are correct.

kevinAlbs: Ah gotcha, thanks for the explanation! For some reason I had thought if libmongocrypt.a was built statically linking to libbson.a, it'd be pulling in libbson's object files.

rcsanchez97: No problem.  Actually, if you think back to when we merged the kms_message project into libmonogcrypt to form a single project it was precisely because prior to the merger we would have to ship libmongocrypt.a and libkms_message.a and the consumer would have to specify both when linking.  By combining the projects, we can produce a libmongocrypt.a that contains all of the libmongocrypt object files and all of the kms_message object files as well.  That is the only way to eliminate the need for the consumer to specify linkage to an upstream dependency of a static library.

PR #113, opened 2020-09-30
MONGOCRYPT-272 support createDataKey with GCP

kevinAlbs:


kms-message/src/kms_gcp_request.c
```diff
@@ -84,11 +84,12 @@ kms_gcp_request_oauth_new (const char *host,
    req->crypto.sign_rsaes_pkcs1_v1_5 = kms_sign_rsaes_pkcs1_v1_5;
    if (opt->crypto.sign_rsaes_pkcs1_v1_5) {
       req->crypto.sign_rsaes_pkcs1_v1_5 = opt->crypto.sign_rsaes_pkcs1_v1_5;
+      req->crypto.sign_ctx = opt->crypto.sign_ctx;
```
kevinAlbs, line 87: This was a bug discovered when testing the callbacks. The signing context was not being copied from the `kms_request_opt` to the `kms_request_t`.

src/mongocrypt-kms-ctx.c
```diff
@@ -438,10 +439,11 @@ _ctx_done_aws (mongocrypt_kms_ctx_t *kms, const char *json_field)
    return ret;
 }
 
-/* An Azure oauth KMS context has received full response. Parse out the bearer
+/* A Azure/GCP oauth KMS context has received full response. Parse out the
+ * bearer
```
kevinAlbs, line 443: I'll fix the weird newline break in the next commit.

src/mongocrypt-kms-ctx.c
```diff
@@ -595,6 +599,114 @@ _ctx_done_azure_wrapkey_unwrapkey (mongocrypt_kms_ctx_t *kms)
    return ret;
 }
 
+/* An AWS KMS context has received full response. Parse out the result or error.
```
kevinAlbs, line 602: Oops, "AWS" should say "gcp" here. I'll fix in the next commit.

src/mongocrypt.c
```diff
@@ -124,6 +124,17 @@ mongocrypt_new (void)
...
+
+   if (0 != _mongocrypt_once (_mongocrypt_do_init) ||
+       !(_native_crypto_initialized)) {
+      mongocrypt_status_t *status = crypt->status;
+
+      CLIENT_ERR ("failed to initialize");
+      /* Return crypt with failure status so caller can obtain error when
+       * calling mongocrypt_init */
+   }
+
```
kevinAlbs, line 137: base64 decoding requires `_mongocrypt_do_init` to be called first. And now we base64 decode the GCP private key in `mongocrypt_setopt_kms_providers` before `mongocrypt_init`. So this now does the necessary global initialization when allocating a `mongocrypt_t`.

src/mongocrypt-ctx-private.h
```diff
@@ -54,6 +63,7 @@ typedef struct __mongocrypt_ctx_opts_t {
    _mongocrypt_key_alt_name_t *key_alt_names;
    mongocrypt_encryption_algorithm_t algorithm;
    _mongocrypt_ctx_opts_azure_kek_t azure_kek;
+   _mongocrypt_ctx_opts_gcp_kek_t gcp_kek;
```
markbenvenuto, line 66: We should use a union here for the azure and gcp options. You could a file a bug to do the same for AWS in the future.

src/mongocrypt-opts-private.h
```diff
@@ -55,6 +62,9 @@ typedef struct {
    void *log_ctx;
    _mongocrypt_buffer_t schema_map;
    _mongocrypt_opts_kms_provider_azure_t kms_provider_azure;
```
markbenvenuto, line 64: Same note as with ctx, use a union for azure and gcp options

src/mongocrypt-private.h
```diff
@@ -83,6 +83,7 @@ struct _mongocrypt_t {
    /* A counter, protected by mutex, for generating unique context ids */
    uint32_t ctx_counter;
    _mongocrypt_cache_oauth_t *cache_oauth_azure;
```
markbenvenuto, line 85: use union

src/mongocrypt-kms-ctx.c
```diff
@@ -595,6 +598,114 @@ _ctx_done_azure_wrapkey_unwrapkey (mongocrypt_kms_ctx_t *kms)
...
+         CLIENT_ERR ("Error in KMS response. HTTP status=%d", http_status);
+         goto fail;
+      }
+      /* GCP error responses include a JSON message, like { "message":
+       * "error", "code": <num> } */
+      bson_destroy (&body_bson);
+      if (!bson_init_from_json (&body_bson, body, body_len, &bson_error)) {
+         bson_init (&body_bson);
+      } else {
+         const char *msg = "";
```
markbenvenuto, line 645: Where do you use msg?

markbenvenuto (changes requested):

src/mongocrypt-opts-private.h
kevinAlbs: These store KMS provider options. A user can configure multiple concurrent KMS providers and, say, do auto encryption with keys encrypted by different providers. Though the KMS provider options for creating a datakey should be represented with a union.

src/mongocrypt-private.h
kevinAlbs: Similarly, we could be authenticating against Azure and GCP concurrently, so we need separate caches.

src/mongocrypt-ctx-private.h
kevinAlbs: Good point, I noted in https://jira.mongodb.org/browse/MONGOCRYPT-283 to do this for AWS.

markbenvenuto (approved):

PR #114, opened 2020-10-06
MONGOCRYPT-274 support GCP auto/explicit en/decrypt

kevinAlbs:
Final portion of GCP support.


src/mongocrypt-key.c
```diff
@@ -140,7 +142,10 @@ _parse_masterkey (bson_iter_t *iter,
       }
 
       if (0 == strcmp ("keyVaultEndpoint", field) ||
-          0 == strcmp ("keyName", field) || 0 == strcmp ("keyVersion", field)) {
+          0 == strcmp ("keyName", field) || 0 == strcmp ("keyVersion", field) ||
```
markbenvenuto, line 145: keyName and keyVersion are listed twice

markbenvenuto (approved):

src/mongocrypt-key.c
kevinAlbs: Fixed.

PR #115, opened 2020-10-06
MONGOCRYPT-264 Add crypto RSAES-PKCS1 SHA256 signing for Windows

markbenvenuto:
Add support for RSA signing with SHA-256 on Windows. This supports PKCS#1 certificates which are PEM blobs titled "PRIVATE KEY".


kevinAlbs (approved):
Fantastic, LGTM!

PR #117, opened 2020-10-09
MONGOCRYPT-284 Wrapped new api in the Java binding

rozza:
Wrapped:
* mongocrypt_setopt_kms_providers - to support new KMS provider options.
* mongocrypt_ctx_setopt_key_encryption_key - to support new data key options.
* mongocrypt_setopt_crypto_hook_sign_rsaes_pkcs1_v1_5 - to support signature callback for GCP.

MONGOCRYPT-284

### DIFF
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
@@ -321,6 +321,12 @@ public class CAPI {
                                    mongocrypt_hash_fn sha_256,
                                    Pointer ctx);
 
+    public static native boolean
+    mongocrypt_setopt_crypto_hook_sign_rsaes_pkcs1_v1_5(
+            mongocrypt_t crypt,
+            mongocrypt_hmac_fn sign_rsaes_pkcs1_v1_5,
+            Pointer sign_ctx);
+
     /**
      * Set a handler to get called on every log message.
      *
@@ -354,6 +360,18 @@ public class CAPI {
     public static native boolean
     mongocrypt_setopt_kms_provider_local(mongocrypt_t crypt,
                                          mongocrypt_binary_t key);
+
+    /**
+     * Configure KMS providers with a BSON document.
+     *
+     * @param crypt The @ref mongocrypt_t object.
+     * @param kms_providers A BSON document mapping the KMS provider names to credentials.
+     * @return A boolean indicating success. If false, an error status is set.
+     * @since 1.1
+     */
+    public static native boolean
+    mongocrypt_setopt_kms_providers(mongocrypt_t crypt,
+                                    mongocrypt_binary_t kms_providers);
 
     /**
      * Set a local schema map for encryption.
@@ -522,6 +540,18 @@ public class CAPI {
     public static native boolean
     mongocrypt_ctx_setopt_masterkey_local (mongocrypt_ctx_t ctx);
 
+    /**
+     * Set key encryption key document for creating a data key.
+     * Currently only applies to Azure.
+     *
+     * @param ctx The @ref mongocrypt_ctx_t object.
+     * @param keyDocument BSON representing the key encryption key document.
+     * @return A boolean indicating success. If false, and error status is set.
+     * @since 1.1
+     */
+    public static native boolean
+    mongocrypt_ctx_setopt_key_encryption_key(mongocrypt_ctx_t ctx,
+                                             mongocrypt_binary_t keyDocument);
 
     /**
      * Initialize a context to create a data key.
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
@@ -542,7 +542,6 @@ public class CAPI {
 
     /**
      * Set key encryption key document for creating a data key.
-     * Currently only applies to Azure.
      *
      * @param ctx The @ref mongocrypt_ctx_t object.
      * @param keyDocument BSON representing the key encryption key document.
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
@@ -25,11 +25,13 @@ import com.mongodb.crypt.capi.CAPI.mongocrypt_t;
 import com.sun.jna.Pointer;
 import org.bson.BsonDocument;
 import org.bson.BsonString;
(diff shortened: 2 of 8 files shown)


bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
```diff
@@ -217,7 +235,7 @@ public MongoCryptContext createDataKeyContext(final String kmsProvider, final Mo
         } else if (kmsProvider.equals("local")) {
             success = mongocrypt_ctx_setopt_masterkey_local(context);
         } else {
-            throw new IllegalArgumentException("Unsupported KMS provider " + kmsProvider);
+            success = mongocrypt_ctx_setopt_key_encryption_key(context, toBinary(options.getMasterKey()).getBinary());
```
jyemin, line 238: The spec says:

> This [mongocrypt_ctx_setopt_key_encryption_key] accepts a BSON document exactly matching the data key options, but expects the BSON document to also have a "provider" field set to either "azure" or "gcp". 

I don't see where the provider field has been added.

kevinAlbs:
Regarding the few failing `build-and-test-java` tasks on the patch build, my guess is that the mappings in https://github.com/mongodb/libmongocrypt/blob/master/bindings/java/mongocrypt/build.gradle.kts#L123:L136 are out-of-date (and consequently they're fetching an older libmongocrypt without the new API, resulting the linker error).

E.g. https://evergreen.mongodb.com/task/libmongocrypt_macos_build_and_test_and_upload_patch_ccfc731a75f77b3a0fa498a93c8a08b025e9c3cf_5f802000d6d80a2339321610_20_10_09_08_32_02 uploads to 
https://mciuploads.s3.amazonaws.com/libmongocrypt/macos/master/latest/5f802000d6d80a2339321610/libmongocrypt.tar.gz

The mapping indicates it is looking for "darwin" in the URL.

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
kevinAlbs: +1

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
```diff
@@ -139,6 +149,15 @@
             }
         }
 
+        if (options.getKmsProviderOptions() != null) {
+            try (BinaryHolder binaryHolder = toBinary(options.getKmsProviderOptions())) {
+                success = mongocrypt_setopt_kms_providers(wrapped, binaryHolder.getBinary());
```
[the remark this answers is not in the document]
rozza: Currently the driver passes on the options as set by the user. I didn't want to add too much logic in the driver / capi bindings.  Initially I coded a `get(Azure|Gcp)KmsProviderOptions` mandating the individual options for each KMS. However, I felt that was over kill and that the validation would be handled by the `mongocrypt_setopt_kms_providers` - However happy to revisit that if we want more validation in Java code.

Once #116 is merged do you want this API to stop calling `mongocrypt_setopt_kms_provider_local` and `mongocrypt_setopt_kms_provider_aws` altogether ?

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
rozza: That was a little unclear to me as what we expect from the user, we seem to use the following format in the database:

```
"masterKey": {
    "provider": "azure",
    "keyVaultEndpoint": "my-key-vault.vault.azure.net",
    "keyName": "test-key"
}
```

And I had some expectations that the field was explicit.

I was expecting users to use that format eg:

```
{
"azure": {
    "provider": "azure",
    "keyVaultEndpoint": "my-key-vault.vault.azure.net",
    "keyName": "test-key"
}}
```

I'll update and treat the provider as implicit if not explicitly provided.

```
"azure": {
    "keyVaultEndpoint": "my-key-vault.vault.azure.net",
    "keyName": "test-key"
}
```

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
```diff
@@ -522,6 +540,18 @@ boolean hmac(Pointer ctx, mongocrypt_binary_t key, mongocrypt_binary_t in, mongo
     public static native boolean
     mongocrypt_ctx_setopt_masterkey_local (mongocrypt_ctx_t ctx);
 
+    /**
+     * Set key encryption key document for creating a data key.
+     * Currently only applies to Azure.
```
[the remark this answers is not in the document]
rozza: Updated.

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
rozza: Updated

rozza:
I'll look into evergreen - the mapping is for JNA - its looking to download the macos version and it maps to darwin arch.

I noticed this error for Ubuntu:  

>  Error looking up function 'mongocrypt_setopt_crypto_hook_sign_rsaes_pkcs1_v1_5': /usr/lib/x86_64-linux-gnu/libmongocrypt.so: undefined symbol: mongocrypt_setopt_crypto_hook_sign_rsaes_pkcs1_v1_5

Looks like its picked up libmongocrypt from `/usr/lib/` which looks to be the cause there.

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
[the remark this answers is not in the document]
kevinAlbs: Sounds good to me. I'll leave it up to you whether you want to consolidate when #116 is merged. The other API will continue to work for backwards compatibility.

rozza:
All java tests now passing.

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/MongoCryptImpl.java
jyemin: @kevinAlbs how should this be tested?

kevinAlbs: libmongocrypt verifies that "provider" is present. If the driver omits it, the prose tests for creating a data key for Azure or GCP will fail.

jyemin (approved):
LGTM!

kevinAlbs (approved):

PR #119, opened 2020-10-16
MONGOCRYPT-275 mock test oauth requests

kevinAlbs:
Adds a couple small tests of mock error responses for GCP and Azure oauth.

### DIFF
--- a/test/data/kms-tests.json
+++ b/test/data/kms-tests.json
@@ -124,5 +124,79 @@
       "\r\n"
     ],
     "expect": "Could not parse Status-Code"
+  },
+  {
+    "description": "Successful oauth request",
+    "ctx": ["azure_oauth_datakey", "azure_oauth_decrypt", "gcp_oauth_datakey", "gcp_oauth_decrypt"],
+    "http_reply": [
+      "HTTP/1.1 200 OK\r\n",
+      "Cache-Control: no-store, no-cache\r\n",
+      "Pragma: no-cache\r\n",
+      "Content-Type: application/json; charset=utf-8\r\n",
+      "Expires: -1\r\n",
+      "Strict-Transport-Security: max-age=31536000; includeSubDomains\r\n",
+      "X-Content-Type-Options: nosniff\r\n",
+      "P3P: CP=\"DSP CUR OTPi IND OTRi ONL FIN\"\r\n",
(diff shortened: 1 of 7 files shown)


test/util/util.c
```diff
@@ -569,6 +569,13 @@ connect_with_tls (const char *endpoint,
       mongoc_stream_destroy (stream);
       stream = NULL;
    }
+
+   if (!mongoc_stream_tls_handshake_block (
```
kevinAlbs, line 573: Fix for the csfle CLI tool. Omitting this explicit call to handshake the TLS stream was not an issue for OpenSSL, since it seemed to do the handshake on the first write. But it was a problem for secure transport.

test/data/kms-tests.json
```diff
@@ -124,5 +124,79 @@
       "\r\n"
     ],
     "expect": "Could not parse Status-Code"
+  },
+  {
+    "description": "Successful oauth request",
+    "ctx": ["azure_oauth_datakey", "azure_oauth_decrypt", "gcp_oauth_datakey", "gcp_oauth_decrypt"],
```
kevinAlbs, line 130: This tests both datakey and decryption, since they go through different code paths for creating KMS requests (encryption/decryption goes through mongocrypt-key-broker.c, creating a datakey in mongocrypt-ctx-datakey.c).

test/test-mongocrypt-kms-responses.c
```diff
@@ -65,6 +65,31 @@ _test_one_kms_response (_mongocrypt_tester_t *tester, bson_t *test)
          ctx = mongocrypt_ctx_new (crypt);
          mongocrypt_binary_destroy (bin);
          bin = _mongocrypt_tester_encrypted_doc (tester);
+         tester->key_file_path = "./test/example/key-document.json";
+         ASSERT_OK (mongocrypt_ctx_decrypt_init (ctx, bin), ctx);
+      } else if (0 == strcmp ("azure_oauth_datakey", ctx_type)) {
+         ctx = mongocrypt_ctx_new (crypt);
```
markbenvenuto, line 71: Hoist this up above the if/else since it just copy/paste

markbenvenuto (approved):

test/test-mongocrypt-kms-responses.c
kevinAlbs: Done.

PR #118, opened 2020-10-14
PYTHON-2371 Add Azure and GCP support for CSFLE

prashantmital:
Tested this locally and it works for both Azure and GCP. No changes are needed in PyMongo.

Still working on getting some tests into Evergreen. Testing suggestions for pymongoctypt welcome @ShaneHarvey.

PR #122, opened 2020-10-20
add local key length check

kevinAlbs:
Adds the same check from `mongocrypt_setopt_kms_provider_local` to `mongocrypt_setopt_kms_providers` for the local provider.


prashantmital (approved):
Tested in #118 and it works!

PR #123, opened 2020-10-20
MONGOCRYPT-287 fix error handling of signature cb

kevinAlbs:


src/mongocrypt-key-broker.c
```diff
@@ -570,7 +572,8 @@ _mongocrypt_key_broker_add_doc (_mongocrypt_key_broker_t *kb,
                    &kb->crypt->log,
                    &kb->crypt->opts,
                    key_doc->kek.provider.gcp.endpoint)) {
-               mongocrypt_kms_ctx_status (&key_returned->kms, kb->status);
+               mongocrypt_kms_ctx_status (&kb->auth_request_gcp.kms,
+                                          kb->status);
```
kevinAlbs, line 576: This was the issue. If creating the oauth request failed when encrypting/decrypting, this attempted to copy the KMS status from the wrong request (`key_returned->kms` instead of `kb->auth_request_gcp.kms`).

markbenvenuto (approved):

DmitryLukyanov (approved):
I've checked the c# code with this implementation and error handling behavior has been solved. 
LGTM

PR #125, opened 2020-10-22
PYTHON-2371 Document expected masterKey fields for new KMS providers

prashantmital:
I overlooked adding these in #118 

@kevinAlbs I did the best I could to add a brief description for each of these fields. It would be helpful if you could give it a quick read and give your feedback.

### DIFF
--- a/bindings/python/pymongocrypt/explicit_encrypter.py
+++ b/bindings/python/pymongocrypt/explicit_encrypter.py
@@ -54,8 +54,8 @@ class DataKeyOpts(object):
             If the `kms_provider` is "azure" it is required and has the
             following fields::
 
-              - `keyVaultEndpoint` (string): Host with optional port, e.g.
-                 "example.vault.azure.net".
+              - `keyVaultEndpoint` (string): Required. Host with optional
+                 port, e.g. "example.vault.azure.net".
               - `keyName` (string): Required. Key name in the key vault.
               - `keyVersion` (string): Optional. Version of the key to use.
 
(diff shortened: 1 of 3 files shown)


bindings/python/pymongocrypt/explicit_encrypter.py
```diff
@@ -49,6 +51,26 @@ def __init__(self, master_key=None, key_alt_names=None):
                 requests to. May include port number, e.g.
                 "kms.us-east-1.amazonaws.com:443".
 
+            If the `kms_provider` is "azure" it is required and has the
+            following fields::
+
+              - `keyVaultEndpoint` (string): Host with optional port, e.g.
+                 "example.vault.azure.net".
```
ShaneHarvey, line 58: Nit: suggest moving keyVaultEndpoint to the end to be consistent with aws "endpoint" and gcp "endpoint". Also it's a bit odd that this field is named keyVaultEndpoint instead of just "endpoint". Is there a reason for that?

ShaneHarvey (commented):
Can you update MongoCryptOptions's `kms_providers` documentation as well?

bindings/python/pymongocrypt/explicit_encrypter.py
prashantmital: I have not placed it as you have suggested because for azure this is a required field, whereas for other KMS providers it is optional (generally we list required options before listing the optional ones). 
As for why the field is named `keyVaultEndpoint` and not `endpoint`, I just used the same name that we intend to pass to `libmongocrypt` - this is also the name for this option in the spec. @kevinAlbs should be able to shed some light on why it was so named.

ShaneHarvey: Okay, can you add "Required." for this option?

ShaneHarvey (approved):
LGTM with a minor suggestion.

bindings/python/pymongocrypt/explicit_encrypter.py
prashantmital: Done.

PR #126, opened 2020-10-27
PYTHON-2409 Allow local.key KMS provider option to be specified as either bytes or a base64-encoded string

prashantmital:
Turns out there were some changes needed beyond just adding the testing. These will need to be merged before CI can pass on https://github.com/mongodb/mongo-python-driver/pull/506

### DIFF
--- a/bindings/python/pymongocrypt/mongocrypt.py
+++ b/bindings/python/pymongocrypt/mongocrypt.py
@@ -18,7 +18,7 @@ import copy
 from pymongocrypt.binary import (MongoCryptBinaryIn,
                                  MongoCryptBinaryOut)
 from pymongocrypt.binding import ffi, lib, _to_string
-from pymongocrypt.compat import str_to_bytes, unicode_type, PY3
+from pymongocrypt.compat import safe_bytearray_or_base64, str_to_bytes
 from pymongocrypt.errors import MongoCryptError
 from pymongocrypt.state_machine import MongoCryptCallback
 
@@ -41,10 +41,12 @@ class MongoCryptOptions(object):
             differ by provider:
               - `aws`: Map with "accessKeyId" and "secretAccessKey" as strings.
               - `azure`: Map with "clientId" and "clientSecret" as strings.
-              - `gcp`: Map with "email" and "privateKey" as strings.
+              - `gcp`: Map with "email" as a string and "privateKey" as
+                a byte array or a base64-encoded string. On Python 2,
+                base64-encoded strings must be passed as unicode literals.
               - `local`: Map with "key" as a 96-byte array or the equivalent
                 base64-encoded string. On Python 2, base64-encoded strings
-                must be passed as a unicode literal.
+                must be passed as unicode literals.
           - `schema_map`: Optional map of collection namespace ("db.coll") to
             JSON Schema.  By default, a collection's JSONSchema is periodically
             polled with the listCollections command. But a JSONSchema may be
@@ -59,6 +61,9 @@ class MongoCryptOptions(object):
             automatic encryption for client side encryption. Other validation
             rules in the JSON schema will not be enforced by the driver and
             will result in an error.
+
+        .. versionadded:: 1.1
+           Support for "azure" and "gcp" kmsProviders.
 
         .. versionchanged:: 1.1
            For kmsProvider "local", the "key" field can now be specified
@@ -93,6 +98,13 @@ class MongoCryptOptions(object):
                 raise ValueError("kms_providers['gcp'] must contain "
                                  "'email' and 'privateKey'")
 
+            key = kms_providers['gcp']['privateKey']
+            safe_key = safe_bytearray_or_base64(
+                key, "kms_providers['gcp']['privateKey']")
+            if key != safe_key:
+                kms_providers = copy.deepcopy(kms_providers)
+                kms_providers['gcp']['privateKey'] = safe_key
+
         if 'local' in kms_providers:
             local = kms_providers['local']
             if not isinstance(local, dict):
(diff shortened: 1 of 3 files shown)


prashantmital:
See the relevant SPEC change here: https://github.com/mongodb/specifications/pull/869/files?file-filters%5B%5D=.rst#diff-de01d398114f63480af23798e09ffca599aa758efd04f15a9beec89b4566041bR324

bindings/python/test/test_mongocrypt.py
```diff
@@ -64,15 +66,46 @@ def test_mongocrypt_binary_out(self):
...
+                else:
+                    self.assertEqual(opts.kms_providers[provider_name], map)
+            self.assertEqual(opts.schema_map, schema_map)
+
+    @unittest.skipIf(not PY3, 'python 3 only')
+    def test_mongocrypt_options_local_kms_py3(self):
+        valid = [
+            ({'local': {'key': b'1'*96}}, None),
+            ({'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'foo'},
+              'local': {'key': b'1'*96}}, None)]
```
ShaneHarvey, line 108: Shouldn't we test base64 encoding on Python 3 too?

bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -42,7 +42,9 @@ def __init__(self, kms_providers, schema_map=None):
               - `aws`: Map with "accessKeyId" and "secretAccessKey" as strings.
               - `azure`: Map with "clientId" and "clientSecret" as strings.
               - `gcp`: Map with "email" and "privateKey" as strings.
```
ShaneHarvey, line 44: The spec says that  ["privateKey" can also be a `bytes` or base64 encoded string](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#kmsproviders):
```
gcp: {
   email: String,
   privateKey: byte[] or String, // May be passed as a base64 encoded string.
   endpoint: Optional<String> // Defaults to oauth2.googleapis.com
}
```

I think we should test this behavior as well.

bindings/python/test/test_mongocrypt.py
```diff
@@ -64,15 +66,46 @@ def test_mongocrypt_binary_out(self):
...
             ({'aws': {'accessKeyId': '', 'secretAccessKey': ''}}, schema_map),
-            ({'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'foo'}}, None),
+            ({'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'foo'}}, None)]
+        for kms_providers, schema_map in valid:
+            opts = MongoCryptOptions(kms_providers, schema_map)
+            self.assertEqual(opts.kms_providers, kms_providers)
+            self.assertEqual(opts.schema_map, schema_map)
+
+    @unittest.skipIf(PY3, 'python 2 only')
+    def test_mongocrypt_options_local_kms_py2(self):
```
ShaneHarvey, line 80: Can we also test that `{'key': bson.binary.Binary(b'1'*96)}` works correctly on both Python 2 and 3?

ShaneHarvey (changes requested):

bindings/python/test/test_mongocrypt.py
prashantmital: Good point. Done.

bindings/python/test/test_mongocrypt.py
prashantmital: Done.

bindings/python/pymongocrypt/mongocrypt.py
prashantmital: Documented and tested!

bindings/python/test/test_mongocrypt.py
```diff
@@ -134,6 +193,36 @@ def test_mongocrypt_validation(self):
...
 
+    @unittest.skipUnless(not PY3, 'bytes are encoded to BSON binary on Py 3')
+    def test_setopt_local_kms_py2(self):
+        callback = MockCallback()
+        kms_dict = {'local': {'key': 'foo'}}
+
+        # Case 1: pass key as string containing bytes (valid)
+        kms_dict['local']['key'] = '\x00' * 96
+        options = MongoCryptOptions(kms_dict)
+        MongoCrypt(options, callback)
```
ShaneHarvey, line 204: It's preferable to call MongoCrypt.close() explicitly rather than relying on `__del__`. One reason is described here: https://cffi.readthedocs.io/en/latest/using.html?highlight=__del__#memory-pressure-pypy :
```
MongoCrypt(options, callback).close()
# or 
mc = MongoCrypt(options, callback)
mc.close()
```

I added `__del__` as a additional layer of safety but I don't think we should get into the habit of relying on it in practice.

bindings/python/test/test_mongocrypt.py
```diff
@@ -134,6 +193,36 @@ def test_mongocrypt_validation(self):
                 MongoCryptError, "local key must be 96 bytes"):
             MongoCrypt(invalid_key_len_opts, callback)
 
+    @unittest.skipUnless(not PY3, 'bytes are encoded to BSON binary on Py 3')
```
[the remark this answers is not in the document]
prashantmital: Done.

bindings/python/test/test_mongocrypt.py
```diff
@@ -134,6 +157,47 @@ def test_mongocrypt_validation(self):
                 MongoCryptError, "local key must be 96 bytes"):
             MongoCrypt(invalid_key_len_opts, callback)
 
+    # @unittest.skipUnless(not PY3, 'bytes are encoded to BSON binary on Py 3')
```
ShaneHarvey, line 160: No longer needed.

ShaneHarvey (approved):
LGTM with one nit.

PR #129, opened 2020-10-30
CSHARP-3142: Fix csharp binding targets file

MikalaiMazurenka:


bindings/cs/MongoDB.Libmongocrypt/MongoDB.Libmongocrypt.targets
```diff
@@ -1,9 +1,10 @@
...
-      <Visible>false</Visible>
-      <Link>mongocrypt.dll</Link>
-    </Content>
-  </ItemGroup>
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemGroup Condition="$([MSBuild]::IsOsPlatform('Windows'))">
+        <Content Include="$(MSBuildThisFileDirectory)../runtimes/win/native/mongocrypt.dll">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+            <Visible>false</Visible>
```
DmitryLukyanov, line 6: let's mimic the same approach as we have in the driver. There is no real reason to make any different approach now. So I suggest to remove `Visisble` field,

MikalaiMazurenka: I'd say that I'm not very happy with making this change, because if we remove remove the `Visible` tag, net framework users will see a `mongocrypt.dll` hanging in their solution.
Done.

MikalaiMazurenka (commented):
Ready for review!

DmitryLukyanov (approved):
LGTM

JamesKovacs (approved):
LGTM

DmitryLukyanov:
Merged.

PR #130, opened 2020-11-03
PYTHON-2400 Add libmongocrypt version check

prashantmital:
I think the error message needs to be improved and ideally, I'd like to do this without `distutils.Version.LooseVersion`.

### DIFF
--- a/bindings/python/CHANGELOG.rst
+++ b/bindings/python/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 1.1.0b0
+--------------------------
+
+- Add support for Azure and GCP KMS providers.
+- Bundle libmongocrypt 1.1.0-beta1 in release wheels.
+
 Changes in Version 1.0.1
 ------------------------
 
--- a/bindings/python/pymongocrypt/version.py
+++ b/bindings/python/pymongocrypt/version.py
@@ -12,4 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from distutils.version import LooseVersion as Version
+
 __version__ = '1.0.2.dev0'
+__min_libmongocrypt_version__ = '1.1.0beta1'
--- a/bindings/python/pymongocrypt/version.py
+++ b/bindings/python/pymongocrypt/version.py
@@ -14,5 +14,5 @@
 
 from distutils.version import LooseVersion as Version
 
-__version__ = '1.0.2.dev0'
+__version__ = '1.1.0.dev0'
 __min_libmongocrypt_version__ = '1.1.0beta1'
--- a/bindings/python/pymongocrypt/version.py
+++ b/bindings/python/pymongocrypt/version.py
@@ -12,7 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from distutils.version import LooseVersion as Version
+try:
+    from pkg_resources import parse_version as _parse_version
+except ImportError:
+    from distutils.version import LooseVersion as _LooseVersion
+
+    def _parse_version(version):
+        return _LooseVersion(version)
+
 
 __version__ = '1.1.0.dev0'
-__min_libmongocrypt_version__ = '1.1.0beta1'
+
+_MIN_LIBMONGOCRYPT_VERSION = '1.1.0beta1'
--- a/bindings/python/pymongocrypt/version.py
+++ b/bindings/python/pymongocrypt/version.py
@@ -12,15 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-try:
-    from pkg_resources import parse_version as _parse_version
-except ImportError:
-    from distutils.version import LooseVersion as _LooseVersion
-
-    def _parse_version(version):
-        return _LooseVersion(version)
-
-
 __version__ = '1.1.0.dev0'
 
 _MIN_LIBMONGOCRYPT_VERSION = '1.1.0beta1'
--- a/bindings/python/pymongocrypt/binding.py
+++ b/bindings/python/pymongocrypt/binding.py
@@ -19,6 +19,7 @@ import sys
 import cffi
 
 from pymongocrypt.compat import PY3
+from pymongocrypt.version import Version, __min_libmongocrypt_version__
 
 
 ffi = cffi.FFI()
@@ -1052,3 +1053,10 @@ else:
 def libmongocrypt_version():
     """Returns the version of libmongocrypt."""
     return _to_string(lib.mongocrypt_version(ffi.NULL))
+
+
+# Check the libmongocrypt version.
+if Version(libmongocrypt_version()) < Version(__min_libmongocrypt_version__):
+    raise RuntimeError(
+        "Expected libmongocrypt version %s or greater, found %s" % (
+            __min_libmongocrypt_version__, libmongocrypt_version()))
--- a/bindings/python/pymongocrypt/binding.py
+++ b/bindings/python/pymongocrypt/binding.py
@@ -19,7 +19,7 @@ import sys
 import cffi
 
 from pymongocrypt.compat import PY3
-from pymongocrypt.version import Version, __min_libmongocrypt_version__
+from pymongocrypt.version import _MIN_LIBMONGOCRYPT_VERSION, _parse_version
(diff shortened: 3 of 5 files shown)


bindings/python/pymongocrypt/version.py
```diff
@@ -12,4 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '1.0.2.dev0'
+from distutils.version import LooseVersion as Version
+
+__version__ = '1.1.0.dev0'
+__min_libmongocrypt_version__ = '1.1.0beta1'
```
ShaneHarvey, line 18: I think this would be more readable without the dunder naming for `min_libmongocrypt_version`.

ShaneHarvey (commented):
Hmm, it's a little unfortunate that we need to parse the version here. From some googling distutils.version is not documented so it's a bit of a grey area to use it. Is that right? 

Another option is that we could catch/re-raise this exception when calling `mongocrypt_setopt_kms_providers`:
```python
  File "/data/mci/d4522601bbd42a0aefd8bcd86342b2e9/src/test/test_encryption.py", line 126, in test_kwargs
    client = MongoClient(auto_encryption_opts=opts, connect=False)
  File "/data/mci/d4522601bbd42a0aefd8bcd86342b2e9/src/pymongo/mongo_client.py", line 755, in __init__
    self._encrypter = _Encrypter.create(
  File "/data/mci/d4522601bbd42a0aefd8bcd86342b2e9/src/pymongo/encryption.py", line 324, in create
    return _Encrypter(io_callbacks, opts)
  File "/data/mci/d4522601bbd42a0aefd8bcd86342b2e9/src/pymongo/encryption.py", line 250, in __init__
    self._auto_encrypter = AutoEncrypter(io_callbacks, MongoCryptOptions(
  File "/opt/python/3.9/lib/python3.9/site-packages/pymongocrypt/auto_encrypter.py", line 31, in __init__
    self.mongocrypt = MongoCrypt(mongo_crypt_opts, callback)
  File "/opt/python/3.9/lib/python3.9/site-packages/pymongocrypt/mongocrypt.py", line 142, in __init__
    self.__init()
  File "/opt/python/3.9/lib/python3.9/site-packages/pymongocrypt/mongocrypt.py", line 153, in __init
    if not lib.mongocrypt_setopt_kms_providers(
  File "/opt/python/3.9/lib/python3.9/site-packages/cffi/api.py", line 912, in __getattr__
    make_accessor(name)
  File "/opt/python/3.9/lib/python3.9/site-packages/cffi/api.py", line 908, in make_accessor
    accessors[name](name)
  File "/opt/python/3.9/lib/python3.9/site-packages/cffi/api.py", line 838, in accessor_function
    value = backendlib.load_function(BType, name)
AttributeError: function/symbol 'mongocrypt_setopt_kms_providers' not found in library 'libmongocrypt.so.0': /usr/lib/x86_64-linux-gnu/libmongocrypt.so.0: undefined symbol: mongocrypt_setopt_kms_providers
```

We could re-raise this as:
```
AttributeError: function/symbol 'mongocrypt_setopt_kms_providers' not found in library 'libmongocrypt.so.0': /usr/lib/x86_64-linux-gnu/libmongocrypt.so.0: undefined symbol: mongocrypt_setopt_kms_providers (Expected libmongocrypt version 1.1 or greater, found 1.0.4)
```

prashantmital:
Yes you are correct, it is somewhat of a grey area. That said, I am not a fan of relying on that exception as a way to confirm whether we are talking to the correct version of the library. Maybe we use `pkg_resources.parse_version` (part of `setuptools` since 8.something - see [commit](https://github.com/pypa/setuptools/commit/18fc31b1516b4493ff1925497d4a6b8bd0110809)) to get the version and run the check when available. If not, we can revert to using `distutils`?

If that is not desirable either, we could do what you suggest with manually catching the exception, but this might end up being a moving target going forward as other endpoints might stop working and we'd have to add checks for those as well. Also, for this kind of an error, I feel it is much nicer to error at import time rather than erroring out at runtime.

ShaneHarvey:
> Also, for this kind of an error, I feel it is much nicer to error at import time rather than erroring out at runtime.

I strongly disagree and here's why: pymongo always attempts to import pymongocrypt. So if we have a misconfigured pymongocrypt installation this PR will break pymongo even if the app is not using FLE. Any approach we take should cause an error when pymongocrypt is (first) used, not imported.

prashantmital:
What about using `pkg_resources.parse_version` ? We seem to rely on setuptools in `setup.py` so there seems to be no reason to not use that instead?

ShaneHarvey:
> What about using pkg_resources.parse_version?

I don't think `pkg_resources` is guaranteed to exist since it's not part of the standard library. `distutils` is: https://docs.python.org/3/distutils/apiref.html#module-distutils.version

This would be a good compromise:
```python
try:
    from pkg_resources import parse_version as _parse_version
except ImportError:
    from distutils.version import LooseVersion as _LooseVersion

    def _parse_version(v):
        return _LooseVersion(v)

```

prashantmital:
This is what this looks like in action (generated by bumping the min version): 
```
======================================================================
ERROR: test_mongocrypt_new (test.test_binding.TestBinding)
  File "/Users/pmital/Developer/libmongocrypt/bindings/python/test/test_binding.py", line 49, in test_mongocrypt_new
    data = lib.mongocrypt_new()
  File "/Users/pmital/Developer/libmongocrypt/bindings/python/pymongocrypt/binding.py", line 1044, in __getattr__
    raise self._error
  File "/Users/pmital/.pyenv/versions/3.7.5/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/pmital/.pyenv/versions/3.7.5/lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/Users/pmital/Developer/libmongocrypt/bindings/python/test/test_binding.py", line 54, in test_mongocrypt_binary_new
    data = lib.mongocrypt_binary_new()
  File "/Users/pmital/Developer/libmongocrypt/bindings/python/pymongocrypt/binding.py", line 1044, in __getattr__
    raise self._error
RuntimeError: Expected libmongocrypt version 2.1.0beta1 or greater, found 1.1.0beta1+20201102git80202647fc
```

prashantmital:
This is hard to test because the check only runs when the library source is initially read. Patching the version number afterward doesn't reproduce the exception.

bindings/python/CHANGELOG.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 1.1.0b0
+--------------------------
+
+- Add support for Azure and GCP KMS providers.
+- Bundle libmongocrypt 1.1.0-beta1 in release wheels.
```
ShaneHarvey, line 8: You'l need to update the release scrip for this to be true: https://github.com/mongodb/libmongocrypt/blob/68a55e519b52b34dcc3ec56f776765eca01367ea/bindings/python/release.sh#L16

bindings/python/pymongocrypt/version.py
```diff
@@ -12,4 +12,15 @@
...
 # limitations under the License.
 
-__version__ = '1.0.2.dev0'
+try:
+    from pkg_resources import parse_version as _parse_version
+except ImportError:
+    from distutils.version import LooseVersion as _LooseVersion
+
+    def _parse_version(version):
+        return _LooseVersion(version)
```
ShaneHarvey, line 21: It's a little odd to put code in here since this is the single-source version file that's read in setup.py. Can we move these imports to binding.py?

ShaneHarvey:
> This is hard to test because the check only runs when the library source is initially read. Patching the version number afterward doesn't reproduce the exception.

Right if we want to test this we should test it for real by linking to a pre-1.1.0beta version.

Can we also add some tests to make sure _parse_version can parse the various libmongocrypt version strings (releases, betas, and dev versions)?

bindings/python/CHANGELOG.rst
prashantmital: Done.

prashantmital:
Opened https://jira.mongodb.org/browse/PYTHON-2417 to add a test at a later date. 
Also added some testing for `_parse_version` but this can likely be improved.

ShaneHarvey (approved):

PR #132, opened 2020-11-04
check for beta in GetVersion.cmake

kevinAlbs:
Running `cmake -P ./cmake/GetVersion.cmake` was returning:
`1.1.0beta1+20201102git80202647fc` though the pre-release tag was `1.1.0-beta1`.

After this change, it has the form `1.1.0-pre2+20201104gitf6765152f6`

This fixes the `publish-packages` tasks.


rcsanchez97 (approved):
LGTM

PR #133, opened 2020-11-05
Release pymongocrypt 1.1.0b0

prashantmital:
Once merged, I will tag the release on master and push the tag. Bumping version to `1.1.0b1.dev0` will be done in a separate PR.

### DIFF
--- a/bindings/python/README.rst
+++ b/bindings/python/README.rst
@@ -144,36 +144,19 @@ Linux::
   1.0.4
 
 
-Installing a Beta or Release Candidate
---------------------------------------
+Installing the Azure and GCP beta release
+-----------------------------------------
 
-Installing Beta and RC releases of PyMongoCrypt follows much the same process as
-`Installing from source`_ as we do not publish wheels for these releases
-to PyPI. Instead, these releases are available as
-`tags <https://git-scm.com/book/en/v2/Git-Basics-Tagging>`_ in the project's
-`GitHub repository <https://github.com/mongodb/libmongocrypt/>`_. Tags are
-usually named as follows::
+To install the beta release, first checkout the ``pymongocrypt-1.1.0b0`` tag
+and install it::
 
-  pymongocrypt-<version>
-
-where ``<version>`` is the Beta/RC version string (e.g. ``1.1.0b0``).
-
-Start by identifying the tag corresponding to the release you want to install
-(in this example, we install ``pymongocrypt-1.1.0b0``) and exporting it as an
-environment variable::
-
-  $ export PYMONGOCRYPT_TAG='pymongocrypt-1.1.0b0'
-
-Now, install PyMongoCrypt from source::
-
-  $ git clone --branch $PYMONGOCRYPT_TAG git@github.com:mongodb/libmongocrypt.git
+  $ git clone --branch 1.1.0b0 git@github.com:mongodb/libmongocrypt.git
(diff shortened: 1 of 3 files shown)


bindings/python/pymongocrypt/binding.py
```diff
@@ -19,7 +19,7 @@
 import cffi
 
 from pymongocrypt.compat import PY3
-from pymongocrypt.version import _MIN_LIBMONGOCRYPT_VERSION
+from pymongocrypt.version import MIN_LIBMONGOCRYPT_VERSION
```
prashantmital, line 22: I made this public since we use it in the instructions above. I don't think it should be a problem, but do let me know what you think.

ShaneHarvey (commented):
I'd prefer if MIN_LIBMONGOCRYPT_VERSION was kept private. I don't think the installation instructions need to be generic based on it. A pitfall here is that importing pymongocrypt might fail, so the user might not even be able to access MIN_LIBMONGOCRYPT_VERSION. (for example if the user has set PYMONGOCRYPT_LIB to libmongocryp 1.0.4).

bindings/python/pymongocrypt/version.py
```diff
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '1.1.0.dev0'
+__version__ = '1.1.0b0'
 
-_MIN_LIBMONGOCRYPT_VERSION = '1.1.0beta1'
+MIN_LIBMONGOCRYPT_VERSION = '1.1.0-beta1'
```
prashantmital, line 17: The old version worked for the version comparison since the parser essentially ignores the hyphens within a patch version but this still needed to be fixed to give the user a convenient way to get the libmongocrypt tag.

bindings/python/README.rst
```diff
@@ -143,6 +143,46 @@ Linux::
...
+
+  $ git clone --branch $PYMONGOCRYPT_TAG git@github.com:mongodb/libmongocrypt.git
+  $ python -m pip install ./libmongocrypt/bindings/python
+
+Next, we identify the required version of libmongocrypt, download the corresponding
+tarball, and extract it::
+
+  $ export LIBMONGOCRYPT_TAG=$(python -c "import pymongocrypt; print(pymongocrypt.version._MIN_LIBMONGOCRYPT_VERSION)")
+  $ export LIBMONGOCRYPT_REVISION=$(git -C ./libmongocrypt rev-list -n 1 $LIBMONGOCRYPT_TAG)
+  $ curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/all/master/$LIBMONGOCRYPT_REVISION/libmongocrypt-all.tar.gz
```
ShaneHarvey, line 177: I'm okay approving this once we change this to hardcode the url to the 1.1.0-beta1 release:
```

  $ curl -O https://s3.amazonaws.com/mciuploads/libmongocrypt/all/master/5d40c56b75f96df4d0e8ba431a7b3ca3218c0891/libmongocrypt-all.tar.gz
```

ShaneHarvey (changes requested):

prashantmital:
I've opened https://jira.mongodb.org/browse/PYTHON-2419 to rollback the breaking changes for the 1.1.0 release.

ShaneHarvey (approved):

PR #139, opened 2020-11-18
feat: support Azure and GCP KMS providers [node.js]

mbroadst:
This patch set updates the Node.js bindings for libmongocrypt to support Azure and GCP as KMS providers for CSFLE. The changes should be straightforward, following the migration guide for this version of libmongocrypt. Notably, the KMS client code introduced a dependency on `BufferList` in order to simplify the logic of reading in chunks, and forwarding those chunks to libmongocrypt.

NODE-2825

### DIFF
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -49,11 +49,20 @@ module.exports = function(modules) {
   /**
    * @typedef {object} KMSProviders
    * @description Configuration options that are used by specific KMS providers during key generation, encryption, and decryption.
-   * @prop {object} [aws] Configuration options for using 'aws' as your KMS provider
-   * @prop {string} [aws.accessKeyId] The access key used for the AWS KMS provider
-   * @prop {string} [aws.secretAccessKey] The secret access key used for the AWS KMS provider
-   * @prop {object} [local] Configuration options for using 'local' as your KMS provider
-   * @prop {Buffer} [local.key] The master key used to encrypt/decrypt data keys. A 96-byte long Buffer.
+   * @property {object} [aws] Configuration options for using 'aws' as your KMS provider
+   * @property {string} [aws.accessKeyId] The access key used for the AWS KMS provider
+   * @property {string} [aws.secretAccessKey] The secret access key used for the AWS KMS provider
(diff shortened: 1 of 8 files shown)


nbbeeken (approved):
Just a suggestion, lgtm!

bindings/node/lib/stateMachine.js
```diff
@@ -216,10 +217,11 @@ module.exports = function(modules) {
     kmsRequest(request) {
       const parsedUrl = request.endpoint.split(':');
       const port = parsedUrl[1] != null ? Number.parseInt(parsedUrl[1], 10) : HTTPS_PORT;
-      const options = { host: parsedUrl[0], port };
+      const options = { host: parsedUrl[0], port, rejectUnauthorized: false };
       const message = request.message;
 
       return new Promise((resolve, reject) => {
+        const buffer = new BufferList();
```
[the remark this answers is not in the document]
mbroadst: `Buffer.concat` is a static method which creates a new buffer by combining two existing ones, and I believe it guarantees a copy (the docs don't indicate left or right on this).  The goal of using `bl` is to minimize copies as much as possible, I think we'll probably want to eventually roll our own but we already have this dependency in the driver so seemed okay for today.

emadum (approved):
LGTM

bindings/node/lib/clientEncryption.js
```diff
@@ -167,7 +206,8 @@ module.exports = function(modules) {
 
       const bson = this._bson;
       options = sanitizeDataKeyOptions(bson, options);
-      const context = this._mongoCrypt.makeDataKeyContext(provider, options);
+      const dataKeyBson = bson.serialize(Object.assign({ provider }, options.masterKey));
```
scriste-sv, line 209: Due to changes on this line all the other options like for example `keyAltNames`, besides `options.masterKey`, are not being serialized and therefore not passed to the context.
Was this something intentional?

kmahar: hi  @scriste-sv, thanks for pointing this out! we've opened [NODE-3118](https://jira.mongodb.org/browse/NODE-3118) and will be implementing a fix soon

PR #140, opened 2020-11-18
MONGOCRYPT-279 disable test-python on RHEL 6.2

kevinAlbs:
Disabling this task for now, so patch builds can succeed until MONGOCRYPT-279 is resolved.


ShaneHarvey:
We're disabling the broken tests in a different way in: https://github.com/mongodb/libmongocrypt/pull/141

PR #143, opened 2020-11-24
Add notes for development packages

kevinAlbs:
In preparation for driver beta releases, users will need to obtain the latest unstable libmongocrypt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ Install the latest release of libmongocrypt with the following.
 brew install mongodb/brew/libmongocrypt
 ```
 
+To install the latest unstable development version of libmongocrypt, use `brew install mongodb/brew/libmongocrypt --HEAD`. Do not use the unstable version of libmongocrypt in a production environment.
+
 ## Building libmongocrypt from source on macOS ##
 First install [Homebrew according to its own instructions](https://brew.sh/). Using Homebrew, install the following dependencies.
 ```
@@ -77,6 +79,7 @@ https://s3.amazonaws.com/mciuploads/libmongocrypt/windows/latest_release/libmong
 
 To download the latest unstable release, download from this URL:
 https://s3.amazonaws.com/mciuploads/libmongocrypt/windows/latest_release/libmongocrypt_unstable.tar.gz
+Do not use the unstable version of libmongocrypt in a production environment.
 
 ### Testing ###
 `test-mongocrypt` mocks all I/O with files stored in the `test/data` and `test/example` directories. Run `test-mongocrypt` from the source directory:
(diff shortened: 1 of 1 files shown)


divjotarora (approved):
LGTM. I'll add a sentence to the Go Driver release notes for the beta next week to say that CSFLE users will need to upgrade to a development version of libmongocrypt.

README.md
```diff
@@ -127,6 +130,9 @@ Do the following when releasing:
 ## Installing libmongocrypt From Distribution Packages ##
 Distribution packages (i.e., .deb/.rpm) are built and published for several Linux distributions.  The installation of these packages for supported platforms is documented here.
 
+### Unstable Development Distribution Packages ###
+To install the latest unstable development package, change `1.0` to `development` in the package URLs listed in the subsequent instructions. For example, `https://libmongocrypt.s3.amazonaws.com/apt/ubuntu <release>/libmongocrypt/1.0` in the instructions would become `https://libmongocrypt.s3.amazonaws.com/apt/ubuntu <release>/libmongocrypt/development`. Do not use the unstable version of libmongocrypt in a production environment.
+
```
rcsanchez97, line 135: My concern here is that we do not necessarily guarantee that the package with the highest version number is the latest for any given release cycle.  The algorithm we use to calculate the version number, for instance, will break down around "beta" versions.  Prior to the first beta tag, the version number will be of the form `x.y.z~dev+<date><git-hash>-0`.  Then when the beta tag is made, it goes to `x.y.z~beta1-0` (or whatever the `beta#` tag is).  This will actually sort as a lower version number than all the `~dev...` versions that preceded it.  On the next commit after the `x.y.z~beta1-0` tag, then we switch to `x.y.z~pre+<date><git-hash>-0`.  The `~pre...` versions will all sort higher than the `~dev....` and `~beta...` versions.  If there is a second beta tag, then the packages produced from it will have the version `x.y.z~beta2-0`, which will again be lower than all the `~dev...` and `~pre...` versions.  Also, if the user specifically installs a `~beta...` version and then later does an upgrade of installed packages, the latest `~dev...` or `~pre...` will be installed over the `~beta...` unless special action is taken.

The two key pieces of information that the user would need to ensure the correct version is installed and does not get inadvertently overridden are:

1. Install the package with the explicit version number: `apt-get install libmongocrypt-dev=1.1.0~beta1-0`
1. Set the package to "hold": `echo "libmongocrypt-dev hold" | sudo dpkg --set-selections` and `echo "libmongocrypt0 hold" | sudo dpkg --set-selections`

Of course, this also means that the user must then specifically upgrade the libmongocrypt-dev and libmongocrypt0 packages later when a newer version is desired.

The part I am not sure about is whether this is important enough to "clutter" the README with it.  Perhaps it might help to describe these additional issues in a separate document and then simply make this section a reference to the separate document.

kevinAlbs: Thanks for writing this up. Per discussion offline, this won't be a problem unless we do a beta2 release. A beta2 release is currently not planned. If we do end up making a beta2 release, we should include these instructions (either as part of the release announcement, or in the README). I created https://jira.mongodb.org/browse/MONGOCRYPT-296 with a 1.1.0-beta2 fixVersion so we don't lose track. If a 1.1.0-beta2 release does not happen, then we'll close that as "Gone Away".

rcsanchez97 (approved):
LGTM

PR #145, opened 2020-12-01
MONGOCRYPT-297 fix publish-packages on Debian and Ubuntu

rcsanchez97:
Note that the patch build does not actually exercise this change, since we don't publish for patch builds.  That said, I did perform an additional patch build to verify that the Python venv sequence worked correctly.  However, the curator invocation still failed because the necessary credentials were not available.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -563,6 +563,13 @@ tasks:
             exit 0
           fi
 
+          # Need requests and poster for notary-client.py
+          python -m virtualenv venv
+          cd venv
+          . bin/activate
+          ./bin/pip install requests
+          ./bin/pip install poster
+          cd ..
           pkg_version=$mongocrypt_version
           CURATOR_RELEASE=${curator_release|"latest"}
           curl -L -O http://boxes.10gen.com/build/curator/curator-dist-rhel70-$CURATOR_RELEASE.tar.gz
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -716,7 +723,7 @@ buildvariants:
   - build-and-test-node
   - name: publish-packages
     distros:
-    - ubuntu1804-test
+    - ubuntu2004-large
```
kevinAlbs, line 726: Was this needed for an upgrade of python versions?

I wonder if this is related to other failures I observed as part of the C++ driver on Ubuntu 18.04 giving an error about virtualenv not being defined: https://evergreen.mongodb.com/task/cxx_driver_mongohouse_ubuntu_test_mongohouse_bd929b61d61289383ef47ed3fd3493eabe9b6af9_20_12_01_16_18_35

kevinAlbs (approved):
LGTM, just had one question.

.evergreen/config.yml
rcsanchez97: Actually, it is needed because for some reason the `ubuntu1804-*` images don't seem to have the virtualenv Python module installed on them.  It seemed simpler to just update to the more recent Ubuntu rather than write a BUILD ticket, especially since the underlying version of the OS is not important to the actions of the publish task.

PR #149, opened 2020-12-14
fix: copy output buffers to libmongocrypt in Node.js bindings

addaleax:
Fixes intermittent crashes on Node.js 14.x and above
when `Buffer::New()` happens to be called with a pointer that
has already used for another `Buffer::New()` call and the object
returned by the first call has not been garbage-collected yet.

NODE-2964

### DIFF
--- a/bindings/node/src/mongocrypt.cc
+++ b/bindings/node/src/mongocrypt.cc
@@ -52,16 +52,19 @@ v8::Local<v8::Object> BufferFromBinary(mongocrypt_binary_t* binary) {
     return scope.Escape(buffer);
 }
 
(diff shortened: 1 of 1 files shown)


emadum (approved):

nbbeeken (approved):

addaleax:
Just in case it isn’t obvious, I don’t have merge permissions here. :)

nbbeeken:
Yep, sorry about that, I'll handle merging. I'm just looking into the Amazon Linux issue, it doesn't seem related at all to this so I'll probably merge soon anyway but trying to learn a bit more first. 🙂

PR #150, opened 2020-12-28
fix: include bson-static-1.0.lib in Windows prebuilts

addaleax:
[MONGOCRYPT-301](https://jira.mongodb.org/browse/MONGOCRYPT-301)


rcsanchez97 (approved):
I'm not sure how we've overlooked this for so long.  Thanks for catching it @addaleax .  LGTM!

kevinAlbs:
Thanks!

kevinAlbs (approved):

PR #151, opened 2021-01-04
MONGOCRYPT-257 propagate errors from KMS message hooks

kevinAlbs:
The KMS message crypto hooks (`kms_sha256`, `kms_sha256_hmac`, `kms_sign_rsaes_pkcs1_v1_5`) did not expose an out parameter for an error message, so an error in one of those hooks would be returned as a generic error message through libmongocrypt like `error constructing KMS message`.

The changes here wrap the passed context with a `mongocrypt_status_t` so the error message is returned through libmongocrypt.

### DIFF
--- a/src/mongocrypt-status.c
+++ b/src/mongocrypt-status.c
@@ -151,14 +151,14 @@ mongocrypt_status_destroy (mongocrypt_status_t *status)
 
 
 void
-_mongocrypt_status_wrap (mongocrypt_status_t *status,
-                         mongocrypt_status_t *to_wrap)
+_mongocrypt_status_append (mongocrypt_status_t *status,
+                           mongocrypt_status_t *to_append)
 {
    char *orig = status->message;
 
-   if (mongocrypt_status_ok (to_wrap)) {
+   if (mongocrypt_status_ok (to_append)) {
       return;
    }
-   status->message = bson_strdup_printf ("%s: %s", orig, to_wrap->message);
+   status->message = bson_strdup_printf ("%s: %s", orig, to_append->message);
    bson_free (orig);
 }
--- a/src/mongocrypt-status.c
+++ b/src/mongocrypt-status.c
@@ -148,3 +148,17 @@ mongocrypt_status_destroy (mongocrypt_status_t *status)
    bson_free (status->message);
    bson_free (status);
 }
+
+
+void
+_mongocrypt_status_wrap (mongocrypt_status_t *status,
+                         mongocrypt_status_t *to_wrap)
+{
+   char *orig = status->message;
+
+   if (mongocrypt_status_ok (to_wrap)) {
+      return;
+   }
+   status->message = bson_strdup_printf ("%s: %s", orig, to_wrap->message);
+   bson_free (orig);
+}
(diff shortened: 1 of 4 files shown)


test/test-mongocrypt-crypto-hooks.c
```diff
@@ -80,7 +80,7 @@ _aes_256_cbc_encrypt (void *ctx,
    bson_string_append_printf (call_history, "ret:%s\n", BSON_FUNC);
    if (0 == strcmp ((char *) ctx, "error_on:aes_256_cbc_encrypt")) {
       mongocrypt_status_set (
-         status, MONGOCRYPT_STATUS_ERROR_CLIENT, 1, "error message", -1);
+         status, MONGOCRYPT_STATUS_ERROR_CLIENT, 1, (char *) ctx, -1);
```
kevinAlbs, line 83: This now checks that the error message returned corresponds to the hook. The error message will contain a message specific to that hook. In this case it is: `error_on:aes_256_cbc_encrypt`.

kevinAlbs:
@prashantmital I tested this with pymongo and verified the full traceback is included when an error occurs in the signature callback loading the PKCS8 key: https://gist.github.com/kevinAlbs/dc07eac2e125c82d6c58b49cfecec42e.

src/mongocrypt-status.c
```diff
@@ -148,3 +148,17 @@ mongocrypt_status_destroy (mongocrypt_status_t *status)
...
+                         mongocrypt_status_t *to_wrap)
+{
+   char *orig = status->message;
+
+   if (mongocrypt_status_ok (to_wrap)) {
+      return;
+   }
+   status->message = bson_strdup_printf ("%s: %s", orig, to_wrap->message);
+   bson_free (orig);
+}
```
samantharitter, line 164: nit, newline

src/mongocrypt-status.c
```diff
@@ -148,3 +148,17 @@ mongocrypt_status_destroy (mongocrypt_status_t *status)
    bson_free (status->message);
    bson_free (status);
 }
+
+
+void
+_mongocrypt_status_wrap (mongocrypt_status_t *status,
```
samantharitter, line 154: So, at the end of this method, `to_wrap` is unchanged, and `status` contains a message like `status's original message: to_wraps' message`

Could you add a comment explaining this?  It's a little difficult to parse reading it.  Maybe changing the names of the parameters could help, too?  What do you think of calling this "append" instead of "wrap" ?

test/test-mongocrypt-crypto-hooks.c
samantharitter: Nice!

samantharitter (commented):
This seems much better!  I have a couple of small comments but otherwise LGTM

src/mongocrypt-status.c
kevinAlbs: Good idea. "append" seems much clearer. Updated and added a function comment.

src/mongocrypt-status.c
kevinAlbs: Done.

samantharitter (approved):
LGTM!

prashantmital (approved):
LGTM!

PR #152, opened 2021-01-06
First fuzzer to be run on oss-fuzz.

DavidKorczynski:
First fuzzer, which will be integrated to be run on oss-fuzz.

Cross referencing:
- mongo-c-driver fuzzers https://github.com/mongodb/mongo-c-driver/pull/706
- OSS-fuzz mongo PR https://github.com/google/oss-fuzz/pull/4808


kevinAlbs:
Thank you @DavidKorczynski!

PR #154, opened 2021-01-14
chore(node-binding): Add initial typings MONGOSH-527

rose-m:
This PR adds TypeScript definitions to the `mongodb-client-encryption` binding package.

The type definitions were created based on the generated README information from JSDoc as well as going in the source code, as not all comments were entirely precise.

Please have a thorough look.

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -1,0 +1,225 @@
+import type { Binary } from 'bson';
+
+export type ClientEncryptionDataKeyProvider = 'aws' | 'local';
+
+/**
+ * An error indicating that something went wrong specifically with MongoDB Client Encryption
+ */
+export class MongoCryptError extends Error {
+}
+
+export interface ClientEncryptionCreateDataKeyCallback {
+  /**
+   * @param error If present, indicates an error that occurred in the creation of the data key
+   * @param dataKeyId If present, returns the id of the created data key
+   */
+  (error?: Error, dataKeyId?: Binary): void;
+}
+
+export interface ClientEncryptionEncryptCallback {
+  /**
+   * @param error If present, indicates an error that occurred in the process of encryption
+   * @param result If present, is the encrypted result
+   */
+  (error?: Error, result?: Binary): void;
+}
+
+export interface ClientEncryptionDecryptCallback {
+  /**
+   * @param error If present, indicates an error that occurred in the process of decryption
+   * @param result If present, is the decrypted result
+   */
+  (error?: Error, result?: any): void;
+}
+
+/**
+ * Configuration options that are used by specific KMS providers during key generation, encryption, and decryption.
+ */
+export interface KMSProviders {
+  /**
+   * Configuration options for using 'aws' as your KMS provider
+   */
+  aws?: {
+    /**
+     * The access key used for the AWS KMS provider
+     */
+    accessKeyId?: string | undefined;
+
+    /**
+     * The secret access key used for the AWS KMS provider
+     */
+    secretAccessKey?: string | undefined;
+  };
+
+  /**
+   * Configuration options for using 'local' as your KMS provider
+   */
+  local?: {
+    /**
+     * The master key used to encrypt/decrypt data keys. A 96-byte long Buffer.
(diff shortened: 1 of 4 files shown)


addaleax (approved):
LGTM but it would be nice to have somebody else take a look as well :)

addaleax:
@mbroadst ?

bindings/node/index.d.ts
```diff
@@ -0,0 +1,225 @@
...
+}
+
+/**
+ * Options to provide when creating a new data key.
+ */
+export interface ClientEncryptionCreateDataKeyProviderOptions {
+  /**
+   * Idenfities a new KMS-specific key used to encrypt the new data key. If the kmsProvider is "aws" it is required.
+   */
+  masterKey?: {
```
aherlihy, line 87: See this discussion over types/future type for masterKey: https://docs.google.com/document/d/1v0UUDvK4Ch8WM03XAFw1Ci-hem84vrgIw07RWCwmyJ4/edit?disco=AAAALAO7L-U I'm not sure if this has happened yet

aherlihy: Just reread that and it's possible he mean to comment about masterKey and not kms....really not sure

markbenvenuto:
Are you going to add the type information for GCP and Azure in this PR or a subsequent PR?

addaleax:
@markbenvenuto Just to make sure I’m not missing anything – even though the Node.js binding’s tests only cover the local and AWS KMS providers, and the Node.js driver documentation/typescript definitions also reference only those two, using GCP or Azure should still work fine as drop-in replacements for those because that is all handled by libmongocrypt on the C layer anyway?

markbenvenuto:
@addaleax Yes, the Node bindings support GCP and Azure. Support was added in 873b8365f4164087e5728305133654eaaf7147d8/#139 and shipped in 1.1.1-beta.0 version of the bindings (f35071a8a999d63b3cb6dd5970c54ac180de5cc2).

GCP and Azure do not work with an earlier version of the bindings even if you have a newer version of libmongocrypt installed. The API was generalized for GCP and Azure and the bindings needed to be updated to use the new API.

durran (approved):
LGTM

rose-m:
@markbenvenuto Thanks for that question. Took a look in the source code and of course it's there... 😅 I've regenerated the README locally, will update the typings and then push both again!

bindings/node/index.d.ts
markbenvenuto: @aherlihy Yes, I put the comment on the wrong part of the type signature.

bindings/node/index.d.ts
```diff
@@ -0,0 +1,340 @@
...
+}
+
+/**
+ * Configuration options for making an AWS encryption key
+ */
+export interface AWSEncryptionKeyOptions {
+  /**
+   * The AWS region of the KMS
+   */
+  region?: string | undefined;
```
markbenvenuto, line 142: Since I am new to Typescript, if I understand this correctly, this type definition means I could have a object of type `AWSEncryptionKeyOptions` without `region` set?

I ask this because `region` and `key` are required for an AWS key. The `endpoint` field is optional.

rose-m: Yes, your understanding is correct.

Right now the types are very "unrestrictive" as I was looking at the documentation of the library when creating that. Do you know what the best place is to grab the exact details on which parameters are required and which are optional? Can I rely on what is defined in https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#user-facing-api?

nbbeeken (approved):

bindings/node/index.d.ts
markbenvenuto: Yes, the best place for the grab to the exact details is in that spec.
https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#kmsproviders
https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#masterkey

emadum (approved):

PR #155, opened 2021-01-19
MONGOCRYPT-304 Fix base64 URL safe padding

markbenvenuto:
The functions that convert regular base64 to url safe base64 terminate their loops on '=' which trims the '=' padding from the final output. While the Azure base64 decoder tolerates the missing padding, python does not accept it and was discovered while testing against a python mock server.

Also fixed a sprintf warning from clang.


kevinAlbs (approved):
Thanks for the fix!

PR #161, opened 2021-02-09
fix: loosen peerDependencies requirement on mongodb

addaleax:
See [NODE-3108](https://jira.mongodb.org/browse/NODE-3108).

### DIFF
--- a/bindings/node/package.json
+++ b/bindings/node/package.json
@@ -47,7 +47,7 @@
     "standard-version": "^8.0.2"
   },
   "peerDependencies": {
-    "mongodb": "^3.4.0"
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
LGTM, I wonder if we want to make this `>=3.6` to suggest a supported driver version.
Maybe that's not the place for that though, I'll let the team chime in.
Does npm 7 now install peerDeps by default?

addaleax:
> Does npm 7 now install peerDeps by default?

Yes, that’s what happens here :+1:

durran (approved):

emadum (approved):

PR #163, opened 2021-02-12
PYTHON-2539 Ensure sessionToken is supported for temp AWS credentials

ShaneHarvey:
We already support `sessionToken` since we use `mongocrypt_setopt_kms_providers`. This change verifies that sessionToken can be passed without error.


behackett (approved):
Neat! FYI @techbelle

PR #165, opened 2021-02-19
pymongocrypt 1.1.0

ShaneHarvey:
Bundle libmongocrypt 1.2.0 in release wheels.
PYTHON-2419 Document that libmongocrypt 1.0 is no longer supported.


bindings/python/CHANGELOG.rst
```diff
@@ -1,11 +1,13 @@
...
 =========
 
-Changes in Version 1.1.0b0
---------------------------
+Changes in Version 1.1.0
+------------------------
 
 - Add support for Azure and GCP KMS providers.
-- Bundle libmongocrypt 1.1.0-beta1 in release wheels.
+- Add support for temporary AWS credentials via the "sessionToken" option.
```
prashantmital, line 8: I think this entry should reflect the fact that `sessionToken` is only supported when binding libmongocrypt 1.2 or later even though pymongocrypt 1.1 supports libmongocrypt 1.1+:
```
- Add support for temporary AWS credentials via the "sessionToken" option when using libmongocrypt 1.2 or later.
```
OR
```
- Add support for temporary AWS credentials via the "sessionToken" option (requires libmongocrypt 1.2+).
```

prashantmital (commented):
Thanks for taking care of this! 

Looking at this PR I now realize that dealing with https://jira.mongodb.org/browse/PYTHON-2583 is not going to be as simple as bumping the pymongocrypt dependency to version 1.1. This is because temp AWS creds will only be supported when using libmongocrypt 1.2+. It seems like introducing a runtime version check on the libmongocrypt binary might be unavoidable?

ShaneHarvey:
Updated to require pymongocrypt 1.2. I also wonder if we can ship wheels for manylinux2014_aarch64, manylinux2014_ppc64le, and manylinux2014_s390x. We may be able to get away using the [Ubuntu 16.04 arm64](https://evergreen.mongodb.com/task/libmongocrypt_ubuntu1604_arm64_build_and_test_and_upload_19ac16abcdf6256b6b981fd7b6d947e09b91e769_21_02_16_11_40_31), [RHEL 7.1 ppc64el](https://evergreen.mongodb.com/task/libmongocrypt_rhel_71_ppc64el_build_and_test_and_upload_19ac16abcdf6256b6b981fd7b6d947e09b91e769_21_02_16_11_40_31), and [RHEL 6.7 s390x](https://evergreen.mongodb.com/task/libmongocrypt_rhel_67_s390x_build_and_test_and_upload_19ac16abcdf6256b6b981fd7b6d947e09b91e769_21_02_16_11_40_31) libmongocrypt builds respectively.

prashantmital (approved):

ShaneHarvey:
I'm going to wait until Monday to release this. Rather not release on a Friday :)

PR #166, opened 2021-03-04
fix: wrap all mongocrypt_binary_t in smart pointers

mbroadst:
This fixes two places where we weren't cleaning up memory after
creating binary instances

NODE-2985

### DIFF
--- a/bindings/node/src/mongocrypt.cc
+++ b/bindings/node/src/mongocrypt.cc
@@ -37,11 +37,7 @@ std::string StringFromBinary(mongocrypt_binary_t* binary) {
 mongocrypt_binary_t* BufferToBinary(v8::Local<v8::Object> node_buffer) {
     uint8_t* buffer = (uint8_t*)node::Buffer::Data(node_buffer);
     size_t buffer_len = node::Buffer::Length(node_buffer);
-
-    uint8_t* buffer_copy = new uint8_t[buffer_len];
-    memcpy(buffer_copy, buffer, buffer_len);
-
-    return mongocrypt_binary_new_from_data(buffer_copy, buffer_len);
+    return mongocrypt_binary_new_from_data(buffer, buffer_len);
 }
 
 v8::Local<v8::Object> BufferFromBinary(mongocrypt_binary_t* binary) {
--- a/bindings/node/src/mongocrypt.cc
+++ b/bindings/node/src/mongocrypt.cc
@@ -121,7 +121,6 @@ std::string errorStringFromStatus(mongocrypt_ctx_t* context) {
     return errorMessage;
 }
 
-Nan::Persistent<v8::Function> MongoCrypt::constructor;
 NAN_MODULE_INIT(MongoCrypt::Init) {
     v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(New);
     tpl->SetClassName(Nan::New("MongoCrypt").ToLocalChecked());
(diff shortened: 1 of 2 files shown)


addaleax (approved):
Looks good, although judging by the valgrind output this only fixes this partially (I think?)

mbroadst:
I was assuming the leaks from new `MongoCrypt` and `MongoCryptContext` were from an ungraceful termination (maybe you ctrl+c'd out of the inspector), this looked like the most obvious low hanging fruit. I threw this together because I had a few minutes this morning, maybe if you have some repro steps it could help the next person?

addaleax:
> I was assuming the leaks from new `MongoCrypt` and `MongoCryptContext` 

Yeah, so, these are harder to fix (without using N-API) and probably also not particularly urgent, because they’re tied to GC. But the ones in `MakeDecryptionContext`, `AddMongoOperationResponse`, `MakeEncryptionContext` seem more problematic.

> were from an ungraceful termination  (maybe you ctrl+c'd out of the inspector)

It’s really just the script mentioned in the ticket, no Ctrl+C or inspector involved, the script exited like it always does

> maybe if you have some repro steps it could help the next person?

I followed the README instructions in https://github.com/mongodb-university/csfle-guides/tree/master/nodejs (except with the latest mongodb-client-encryption version) and then ran `valgrind --leak-check=full node clients.js` instead of just `node clients.js`.

emadum (approved):
LGTM

mbroadst:
@addaleax thanks, I found another bug. I think `mongocrypt_binary_t` used to be an "owning view" so we were making copies of buffers to pass into libmongocrypt, so that's been fixed here as well. I did some further analysis to see if we had leaks with the various context types, and can confirm that they aren't actual leaking anything it's just that node doesn't explicit GC on app close, leaving it up to the kernel to do the job (see [here](https://github.com/nodejs/abi-stable-node-addon-examples/issues/17#issuecomment-378532484) for more detail). Applying the following patch to `clients.js` and running your valgrind line with `node --expose-gc` will confirm the leaks aren't there:

addaleax (approved):
LGTM, and yeah, makes sense that changing `mongocrypt_binary_new_from_data` semantics would lead to problems here :)

> and can confirm that they aren't actual leaking anything it's just that node doesn't explicit GC on app close, leaving it up to the kernel to do the job (see here for more detail). 

Yeah, I can also provide a bit more context around this :) Ben's comment is correct, of course, but Node.js has done quite a bit of work towards cleaning up its own resources before process exit. The two main motivation for that are a) Worker threads, where a Node.js instance *has* to fully clean up before exit, and b) embedded scenarios where the Node.js main thread instance does not necessarily have the same lifetime as the process itself.

This extends to addons in the sense that, if users want to be able to use addons from environments like those, these addons also have to perform proper cleanup (and, usually, also be threadsafe).

In order to support that, all supported versions of Node.js (incl. v10.x) have an `AddEnvironmentCleanupHook()` API that allows registering hooks which are run before a Node.js instance shuts down, regardless of GC. In Nan-based native addons, that is the only way to clean up e.g. persistent handles that refer to JS objects (because Node.js has no insight into their existence or ownership by addons). N-API addons, on the other hand, perform this cleanup automatically as far as JS-bound resources are concerned.

Long story short, there's probably not much of a point in adding that extra work here unless it's a) part of a migration to N-API (which I hope/assume is something we'd like to do at some point down the road?) or b) there's reasons to work on thread safety/Worker compatibility (e.g. because a customer needs it), in which case the end-of-addon-lifetime-cleanup probably just comes as a side effect of that transition.

nbbeeken (approved):
LGTM, integration tests with driver pass

PR #167, opened 2021-03-08
MONGOCRYPT-298: Add a publish script.

DmitryLukyanov:
Expected steps to make release:
1. Specify income variables
2. Run the script
3. Validate and push the created tag to the upstream manually
4. Validate and push the created .nupkg file to nuget org via:
>dotnet nuget push "*.nupkg" --source https://api.nuget.org/v3/index.json --api-key <Nuget_Api_Key>


bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+# 4. artifacts. After all, the generated .nupkg files will be placed here.
+#
+# Input data:
+# $libmongocryptAllUrl - this is the link on EG patch that contains all required platform-specific C libraries.
+
+packagingFolder="c:/build"
+bindingsReleaseVersion="1.2.1"
+fork="https://github.com/DmitryLukyanov/libmongocrypt.git"
+bracnhName="csharp-v1.2.1_release"
+libmongocryptAllUrl="https://mciuploads.s3.amazonaws.com/libmongocrypt/all/master/508e21f4abff9f5519e0357a63a4ad30d2c24692/libmongocrypt-all.tar.gz"
```
DmitryLukyanov, line 15: this link is taken from this EG patch: https://evergreen.mongodb.com/task/libmongocrypt_publish_snapshot_upload_all_508e21f4abff9f5519e0357a63a4ad30d2c24692_21_01_22_14_28_35. Originally I thought that we can trigger getting these libs automatically somehow, but this way might be ok also at least as the first step

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+# nocrypto
+cp ${packagingFolder}/${libmongocryptAllFolder}/macos/nocrypto/lib/libmongocrypt.dylib ${packagingFolder}/${downloadedMongocryptFolder}
+cp -r ${packagingFolder}/${downloadedMongocryptFolder}/* ${packagingFolder}/${csharpGitTagName}/libmongocrypt/cmake-build/RelWithDebInfo
+
+echo "Building.."
+cd ../../bindings/cs/MongoDB.Libmongocrypt
+dotnet build MongoDB.Libmongocrypt.csproj --configuration Release
+echo "Running tests.."
+cd ../MongoDB.Libmongocrypt.Test
+# net452 tests will fail with the expected error `System.PlatformNotSupportedException : MongoDB.Libmongocrypt needs to be run in a 64-bit process.`
```
DmitryLukyanov, line 66: I didn't find a way how to run `dotnet test` in x64 mode for net452 (don't test doesn't support runtime option), so this case fails with the expected error all the time. I'm open to suggestions here

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+# 3. downloaded_mongocrypt. This folder contains the used in the libmongocrypt platform-specific C libraries.
+# 4. artifacts. After all, the generated .nupkg files will be placed here.
+#
+# Input data:
+# $libmongocryptAllUrl - this is the link on EG patch that contains all required platform-specific C libraries.
+
+packagingFolder="c:/build"
+bindingsReleaseVersion="1.2.1"
+fork="https://github.com/DmitryLukyanov/libmongocrypt.git"
+bracnhName="csharp-v1.2.1_release"
```
JamesKovacs, line 14: bracnhName => branchName

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+# 2. csharp-v{$bindingsReleaseVersion}. After all, this folder will containts the fully ready source code.
+# 3. downloaded_mongocrypt. This folder contains the used in the libmongocrypt platform-specific C libraries.
+# 4. artifacts. After all, the generated .nupkg files will be placed here.
+#
+# Input data:
+# $libmongocryptAllUrl - this is the link on EG patch that contains all required platform-specific C libraries.
+
+packagingFolder="c:/build"
+bindingsReleaseVersion="1.2.1"
+fork="https://github.com/DmitryLukyanov/libmongocrypt.git"
```
JamesKovacs, line 13: This should point to the main repo by default.

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+# 1. libmongocrypt-all. This folder contains all platform-specific C libraries.
+# 2. csharp-v{$bindingsReleaseVersion}. After all, this folder will containts the fully ready source code.
+# 3. downloaded_mongocrypt. This folder contains the used in the libmongocrypt platform-specific C libraries.
+# 4. artifacts. After all, the generated .nupkg files will be placed here.
+#
+# Input data:
+# $libmongocryptAllUrl - this is the link on EG patch that contains all required platform-specific C libraries.
+
+packagingFolder="c:/build"
+bindingsReleaseVersion="1.2.1"
```
JamesKovacs, line 12: Any variables that change version to version (e.g. `bindingsReleaseVersion`, `branchName`, and `libmonogcryptAllUrl`) should be required parameters rather than hard-coded values.

JamesKovacs: For `libmongocryptAllUrl`, we might be able to use the `latest` URL, which should point to the latest successful build. We should confirm this will always contain the desired unmanaged dependencies. If so, we can hard-code this URL in the script.
https://mciuploads.s3.amazonaws.com/libmongocrypt/all/master/latest/libmongocrypt-all.tar.gz

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+libmongocryptAllUrl="https://mciuploads.s3.amazonaws.com/libmongocrypt/all/master/508e21f4abff9f5519e0357a63a4ad30d2c24692/libmongocrypt-all.tar.gz"
+libmongocryptAllFolder="libmongocrypt-all"
+downloadedMongocryptFolder=downloaded_mongocrypt
+csharpGitTagName=csharp-v$bindingsReleaseVersion
+
+# Start
+cd $packagingFolder
+#Download libs
+echo "Download libs"
+mkdir $libmongocryptAllFolder
```
JamesKovacs, line 24: We should use absolute paths for these commands. Then you don't have to rely on `cd <<CORRECT_DIRECTORY>>`. This makes the script easier to understand as you don't have to keep track of your current working directory.

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+cd $libmongocryptAllFolder
+curl --url $libmongocryptAllUrl --output libmongocrypt-all.tar.gz
+echo "Unzipping.."
+tar xzvf libmongocrypt-all.tar.gz
```
JamesKovacs, line 28: You can use `-C <<directory>>` to specify which directory to extract into.

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+cd ../
+read -p "Press any key to proceed with libmongocrypt cloning.."
+
+# Clone libmongocrypt
+echo "Clone Libmongocrypt"
+
+mkdir $csharpGitTagName
+cd $csharpGitTagName
+git clone --branch $bracnhName $fork
```
JamesKovacs, line 37: $bracnchName => $branchname

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+read -p "Press any key to push this tag to origin.."
+git push origin $csharpGitTagName
+# Make pushing this tag to the upstream manually
+read -p "Press any key to proceed with building and testing.."
+
+# Create cmake-build
+mkdir cmake-build
+cd cmake-build
+mkdir RelWithDebInfo
+cd RelWithDebInfo
```
JamesKovacs, line 52: You can simplify this to `mkdir -p cmake-build/RelWithDebInfo`. `-p` creates all intervening directories as needed. As well, consider using absolute paths as mentioned previously.

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "Running tests.."
+cd ../MongoDB.Libmongocrypt.Test
+# net452 tests will fail with the expected error `System.PlatformNotSupportedException : MongoDB.Libmongocrypt needs to be run in a 64-bit process.`
+dotnet test MongoDB.Libmongocrypt.Test.csproj --configuration Release
+read -p "Press any key to proceed with packing.."
+
+# packing
+cd ../MongoDB.Libmongocrypt
+dotnet pack MongoDB.Libmongocrypt.csproj --configuration Release -p:PackageVersion=${bindingsReleaseVersion} -p:NoWarn=NU1605 -p:NoWarn=NU5100 --output ../../../../../artifacts
+read -p "Done. Don't forget to push a new tag to the upstream and publish the .nupkg file"
```
JamesKovacs, line 73: Since the script is marked with `read -p` to ensure that each step in the build is checked, why not include the publishing of the .nupkg file as a last step?

JamesKovacs (changes requested):
Great start at automating the build. Given the steps and dependencies between them, wouldn't it make more sense to automate this with Cake rather than Bash? Something to consider, but I'm OK with this being Bash. Added some suggestions to clean up the script and make it easier to follow and refactor.

DmitryLukyanov:
I thought about cake and actually, I prefer it too, I didn't use it because we don't use cake in the libmongocrypt solution. Maybe it makes sense to do it, if at some moment we will merge libmongocrypt with the driver, it will be helpful to have the same format.

DmitryLukyanov:
We also can convert it to the cake format in different ticket later

bindings/cs/Scripts/Create-and-pack-nuget.sh
DmitryLukyanov: agreed about the first. As for the second, I think we cannot rely on the above link. For example the latest libmongocrypt release, for now, is `1.2.1`, but we changed the code only for `1.2.0`. Idk whether it's problematic to use the latest library in the next release without our code changes, but even if not so, we can have such problems later, but I will confirm this understanding

JamesKovacs:
> I thought about cake and actually, I prefer it too, I didn't use it because we don't use cake in the libmongocrypt solution. Maybe it makes sense to do it, if at some moment we will merge libmongocrypt with the driver, it will be helpful to have the same format.

Cake is just a build scripting language. It is an easy-enough add and doesn't force any other bindings to use it. I think it worth refactoring to use Cake if you're going to modify the Bash script to any great extent. More consistent with the .NET/C# driver and easier to document the build tasks and dependencies between them.

bindings/cs/Scripts/Create-and-pack-nuget.sh
DmitryLukyanov: agreed

bindings/cs/Scripts/Create-and-pack-nuget.sh
DmitryLukyanov: the good catch, will change it

bindings/cs/Scripts/Create-and-pack-nuget.sh
DmitryLukyanov: Given that we cannot republish the nuget with the same version, I would double-check that everything is correct before the last confirmation, do you think it's overcautious?

JamesKovacs: It's a good idea to make this a separate task. If you rewrite this as a Cake script, then you can have a `PushToNuGet` task that ensures the packages are built and then publishes them to NuGet as we do with the .NET/C# driver. That way you ensure that the push to NuGet is scripted properly without having to manually copy/paste the command line every time.

bindings/cs/Scripts/Create-and-pack-nuget.sh
JamesKovacs: My intent was to avoid having to specify a SHA1 hash as a command line argument. Maybe ask if the official `libmongocrypt` builds could be symlinked to the SHA1 hashes automatically. For example:
https://mciuploads.s3.amazonaws.com/libmongocrypt/all/v1.2.0/libmongocrypt-all.tar.gz

would be symlinked to:

https://mciuploads.s3.amazonaws.com/libmongocrypt/all/master/508e21f4abff9f5519e0357a63a4ad30d2c24692/libmongocrypt-all.tar.gz

I would rather remember a tag name than a SHA1. Plus tag names are easier to visually validate than a SHA1. (e.g. Did I copy the entire value? Did I leave off a digit?)

DmitryLukyanov: yeah, I will check what we can do here

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+var libmongocryptCsDirectory = csharpBindingsDirectory.Combine("bindings").Combine("cs");
+var libmongocryptSolutionDirectory = libmongocryptCsDirectory.Combine("MongoDB.Libmongocrypt");
+var libmongocryptSolutionFile = libmongocryptSolutionDirectory.CombineWithFilePath("MongoDB.Libmongocrypt.csproj");
+var libmongocryptTestsSolutionDirectory = libmongocryptCsDirectory.Combine("MongoDB.Libmongocrypt.Test");
+var artifactsDirectory = buildDirectory.Combine("artifacts");
+
+Task("Prepare")
+    .Does(() =>
+    {
+        if (DirectoryExists(buildDirectory))
```
DmitryLukyanov, line 37: >EnsureDirectoryExists
>CleanDirectory

the above didn't work for me, because I could not remove a git folder from cloned libmongocrypt repo

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+var gitVersion = GitVersion();
+
+var buildDirectory = MakeAbsolute(Directory(GetSettingValue("buildDirectory", "c:\\build")));
+var libmongocryptAllDirectory=buildDirectory.Combine("libmongocrypt-all");
+var downloadedMongocryptDirectory=buildDirectory.Combine("downloadedMongocryptDirectory");
+var localReleaseVersion = "local-0.0.0";
+var releaseVersion = GetSettingValue("releaseVersion", localReleaseVersion);
+var fork = GetSettingValue("fork", "https://github.com/mongodb/libmongocrypt.git");
+var branch = GetSettingValue("branch", "master"); 
+var libmongocryptAllUrl = GetSettingValue("url", "https://mciuploads.s3.amazonaws.com/libmongocrypt/all/master/508e21f4abff9f5519e0357a63a4ad30d2c24692/libmongocrypt-all.tar.gz");
```
DmitryLukyanov, line 24: this is still a question

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+        Information("Downloading native libs..");
+        EnsureDirectoryExists(libmongocryptAllDirectory);
+        var nativeLibrariesArchive = libmongocryptAllDirectory.CombineWithFilePath("libmongocrypt-all.tar");
+        DownloadFile(libmongocryptAllUrl, nativeLibrariesArchive);
+        
+        Information("Unzipping..");
+        UncompressToTheCurrentDirectory(nativeLibrariesArchive);
+
+        Information("Cloning the libmongocrypt repo..");
+        GitClone(
```
DmitryLukyanov, line 62: should we consider some debugging case when we didn't clone the whole repo but copy the current working directory?

DmitryLukyanov: in this case we can use `gitVersion.LegacySemVer` instead of `var localReleaseVersion = "local-0.0.0";`

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+        };
+        DotNetCorePack(
+            projectFullPath,
+            settings);            
+    });
+    
+Task("NugetPush")
+    .Does(() =>
+    {
+        ThrowIfLocalRelease();
```
DmitryLukyanov, line 138: just in case

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+
+        Information("Show origin:");
+        Git(libmongocryptSolutionDirectory, "remote -v");
+        Git(libmongocryptSolutionDirectory, $"tag -a {csharpBindingsGitTagName} -m {csharpBindingsGitTagName}"); 
+        Git(libmongocryptSolutionDirectory, $"push origin {csharpBindingsGitTagName}"); 
+    });
+
+RunTarget(target);
+
+string GetSettingValue(string commandArgumentName, string defaultValue)
```
DmitryLukyanov, line 165: this method allows us to get the required arguments from the initial command like:

> .\build.ps1 -ScriptArgs '--releaseVersion="1"'

or via the EnvironmentVariable with naming like: `LIBMONGOCRYPT_PACKAGING_RELEASEVERSION`

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+    });
+    
+Task("GreateGitTag")
+    .Does(() =>
+    {
+        ThrowIfLocalRelease();
+
+        Information("Show origin:");
+        Git(libmongocryptSolutionDirectory, "remote -v");
+        Git(libmongocryptSolutionDirectory, $"tag -a {csharpBindingsGitTagName} -m {csharpBindingsGitTagName}");
```
DmitryLukyanov, line 159: it's possible to use pre-defined cake methods here like described here: https://stackoverflow.com/questions/44484889/create-tag-to-remote-repository-using-cake-build. But I had troubles matching the existed wrappers to the particular required command. I think the current way is also good-looking

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+                ApiKey = nugetApi,
+                Source = "https://api.nuget.org/v3/index.json"
+            });
+        Information(gitVersion.FullSemVer.ToString());
+    });
+    
+Task("GreateGitTag")
+    .Does(() =>
+    {
+        ThrowIfLocalRelease();
```
DmitryLukyanov, line 155: Just in case

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+    
+Task("NugetPush")
+    .Does(() =>
+    {
+        ThrowIfLocalRelease();
+
+        var nugetApi = GetSettingValue("NugetApiKey", null);
+        var packageFilePath = artifactsDirectory.CombineWithFilePath($"{libmongocryptSolutionFile.GetFilename().ToString()}.{releaseVersion}.nupkg");
+        Information(packageFilePath);
+        NuGetPush(
```
DmitryLukyanov, line 143: I didn't check this, so we can do it during the next release

bindings/cs/Scripts/build.ps1
```diff
@@ -0,0 +1,275 @@
+##########################################################################
```
DmitryLukyanov, line 1: the below files are copied from the driver solution without any changes

bindings/cs/Scripts/Create-and-pack-nuget.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
```
DmitryLukyanov, line 1: this will be removed

bindings/cs/.gitignore
```diff
@@ -7,3 +7,6 @@ obj/
 
 # VS settings
 .vs/
+
+# Other
+/Scripts/Tools
```
DmitryLukyanov, line 12: this needs to hide the cake temp files

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,217 @@
...
+        Git(libmongocryptSolutionDirectory, $"tag -a {csharpBindingsGitTagName} -m {csharpBindingsGitTagName}"); 
+        Git(libmongocryptSolutionDirectory, $"push origin {csharpBindingsGitTagName}"); 
+    });
+
+RunTarget(target);
+
+string GetSettingValue(string commandArgumentName, string defaultValue)
+{
+    var optionValue = Argument(commandArgumentName, (string)null);
+    if (optionValue != null)
```
DmitryLukyanov, line 168: the command-line argument has priority

bindings/cs/Scripts/Create-and-pack-nuget.sh
DmitryLukyanov: Kevin is out this week, I think we can move this logic to a separate ticket. I will discuss it with him when he will be back

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,214 @@
...
+                RecurseSubmodules = true
+            });
+
+        EnsureDirectoryExists(libmongocryptRelWithDebInfoDirectory);
+        EnsureDirectoryExists(downloadedMongocryptDirectory);
+        CopyFile(
+            libmongocryptAllDirectory.Combine("windows-test").Combine("bin").CombineWithFilePath("mongocrypt.dll"),
+            downloadedMongocryptDirectory.CombineWithFilePath("mongocrypt.dll"));
+        CopyFile(
+            libmongocryptAllDirectory.Combine("ubuntu1804-64").Combine("nocrypto").Combine("lib").CombineWithFilePath("libmongocrypt.so"),
```
JamesKovacs, line 76: What are these `nocrypto` directories? It sounds like versions of the binaries that expose the desired API but do not perform any crypto. Shouldn't we be packaging up the same binaries but without `nocrypto` in the pathname?

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,214 @@
...
+            downloadedMongocryptDirectory.CombineWithFilePath("libmongocrypt.so"));
+        CopyFile(
+            libmongocryptAllDirectory.Combine("macos").Combine("nocrypto").Combine("lib").CombineWithFilePath("libmongocrypt.dylib"),
```
JamesKovacs, line 79: Same as above. Why `nocrypto`?

bindings/cs/Scripts/build.cake
JamesKovacs: Excellent idea.

bindings/cs/Scripts/build.cake
JamesKovacs: Agreed.

bindings/cs/Scripts/build.cake
JamesKovacs: GreateGitTag => CreateGitTag

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,214 @@
...
+        return optionValue;
+    }
+
+    var environmentVariableName = $"LIBMONGOCRYPT_PACKAGING_{commandArgumentName.ToUpper()}";
+    var environmentVariable = Environment.GetEnvironmentVariable(environmentVariableName);
+    if (environmentVariable == null)
+    {
+        if (defaultValue == null)
+        {
+            throw new Exception($"Neither {commandArgumentName} command argument or {environmentVariableName} environmentVariable have not been configured.");
```
JamesKovacs, line 176: Grammar fix:
$"Neither {commandArgumentName} command argument nor {environmentVariableName} environmentVariable have been configured."

JamesKovacs (changes requested):
A few minor comments and fixes. Overall looking good.

bindings/cs/Scripts/build.cake
DmitryLukyanov: because in local development (and as the result in a local solution), we don't have control (via git branches) over which native libraries are copied into the `cmake-build` folder, so if we're working on some ticket, then if we checkout on the master, the native libraries in `cmake-build` folder won't be changed, we can always clean them in each release, but then it can broke the local ticket

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,218 @@
...
+
+var gitVersion = GitVersion();
+
+var buildDirectory = MakeAbsolute(Directory(GetSettingValue("buildDirectory", "c:\\build")));
+var libmongocryptAllDirectory=buildDirectory.Combine("libmongocrypt-all");
+var downloadedMongocryptDirectory=buildDirectory.Combine("downloadedMongocryptDirectory");
+var localReleaseVersion = "local-0.0.0";
+var releaseVersion = GetSettingValue("releaseVersion", localReleaseVersion);
+var fork = GetSettingValue("fork", "https://github.com/mongodb/libmongocrypt.git");
+var branch = GetSettingValue("branch", "master");
```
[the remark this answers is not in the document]
DmitryLukyanov: in this case, we will need to create a tag manually before making a release. It also can work, my idea was to delegate this to the script. So we can say like make a release from this commit, and the related tag will be created automatically

bindings/cs/Scripts/build.cake
```diff
@@ -0,0 +1,214 @@
...
+                BranchName = branch,
+                Checkout = true,
+                IsBare = false,
+                RecurseSubmodules = true
+            });
+
+        EnsureDirectoryExists(libmongocryptRelWithDebInfoDirectory);
+        EnsureDirectoryExists(downloadedMongocryptDirectory);
+        CopyFile(
+            libmongocryptAllDirectory.Combine("windows-test").Combine("bin").CombineWithFilePath("mongocrypt.dll"),
```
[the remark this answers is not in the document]
DmitryLukyanov: agreed, but we have what we have

bindings/cs/Scripts/build.cake
DmitryLukyanov: nope, this related to this ticket (and macos one): https://jira.mongodb.org/browse/MONGOCRYPT-86. In two words, in Linux and MacOS, we configure crypto callbacks like here https://github.com/mongodb/libmongocrypt/blob/master/bindings/cs/MongoDB.Libmongocrypt/CryptClientFactory.cs#L49, so the C level calls these hooks to make signing/hashing/encryption and so on steps. But in windows, this stuff is built-in to the C library

bindings/cs/Scripts/build.cake
DmitryLukyanov: done

bindings/cs/Scripts/build.cake
DmitryLukyanov: done

bindings/cs/Scripts/build.cake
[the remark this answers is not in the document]
JamesKovacs: Your way works too. Resolving.

bindings/cs/Scripts/build.cake
JamesKovacs: Fair enough.

bindings/cs/Scripts/build.cake
JamesKovacs: Just wanted to make sure we were packing the right binaries. You're much more familiar with the details. So I'll trust you on this.

JamesKovacs (approved):
LGTM. Don't forget to delete the outdated `Create-and-pack-nuget.sh` script.

DmitryLukyanov:
> 
> 
> LGTM. Don't forget to delete the outdated `Create-and-pack-nuget.sh` script.

done

JamesKovacs (approved):
LGTM

JamesKovacs (approved):
LGTM

bindings/cs/Scripts/Create-and-pack-nuget.sh
kevinAlbs: > Idk whether it's problematic to use the latest library in the next release without our code changes, but even if not so, we can have such problems later, but I will confirm this understanding

Ideally it shouldn't. If in the unlikely event we do a backwards breaking release, we'd have to ensure drivers opt-in to consuming it. But, specifying the tag guarantees you don't get an unwanted upgrade.

I agree having a URL available via a tag seems like a good idea. I filed https://jira.mongodb.org/browse/MONGOCRYPT-315 to make that change.

PR #168, opened 2021-03-19
fix(NODE-2995): add metadata client usage

nbbeeken:
Automatic client side encryption needs to perform metadata look ups
like listCollections. In situations where the connection pool size
is constrained or in full use it can be impossible for an operation
to proceed. Adding a separate client in these situations permits the
metadata look ups to proceed unblocking operations.

Related to: https://github.com/mongodb/node-mongodb-native/pull/2760

### DIFF
--- a/bindings/node/binding.gyp
+++ b/bindings/node/binding.gyp
@@ -21,11 +21,7 @@
       'src/mongocrypt.cc'
     ],
     'xcode_settings': {
-      'MACOSX_DEPLOYMENT_TARGET': '10.12',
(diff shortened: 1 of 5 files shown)


emadum (commented):
Should this be targeting `4.0` rather than `master`?

nbbeeken:
> Should this be targeting `4.0` rather than `master`?

This is the libmongocrypt repo 🙂  not the driver

emadum (approved):

PR #175, opened 2021-06-07
fix(NODE-3326): binding.gyp missing from files list in package.json

nbbeeken:
- Add test to ensure package contains exactly the files listed in an allow list

### DIFF
--- a/bindings/node/package.json
+++ b/bindings/node/package.json
@@ -11,7 +11,8 @@
     "lib",
     "src",
     "build",
-    "index.d.ts"
+    "index.d.ts",
+    "binding.gyp"
   ],
   "directories": {
     "lib": "lib"
@@ -53,7 +54,8 @@
     "segfault-handler": "^1.3.0",
     "sinon": "^4.5.0",
     "sinon-chai": "^3.6.0",
-    "standard-version": "^8.0.2"
+    "standard-version": "^9.3.0",
+    "tar": "^6.1.0"
   },
   "peerDependencies": {
     "mongodb": ">=3.4.0"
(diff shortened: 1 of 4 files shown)


bindings/node/package.json
```diff
@@ -10,8 +10,8 @@
     "CHANGELOG.md",
     "lib",
     "src",
-    "build",
```
nbbeeken, line 13: @mbroadst @addaleax Hey, does this seem right to y'all? It's my understanding that the goal of prebuild is to either download or compile the package onsite, currently the build directory ships with a compiled version that matches the releaser's computer.

mbroadst: LGTM,  this must have been left in from early testing. Do you need to include the `binding.gyp` for that matter?

nbbeeken: If a prebuild isn't available then it falls back on attempting to compile the module and this file is needed to do so.

mbroadst: great point :facepalm:

bindings/node/test/release.test.js
```diff
@@ -0,0 +1,56 @@
...
+    fs.unlinkSync(packFile);
+  });
+
+  for (const requiredFile of REQUIRED_FILES) {
+    it(`should contain ${requiredFile}`, () => {
+      expect(tarFileList).to.includes(requiredFile);
+    });
+  }
+
+  it('should not have extraneous files', () => {
```
emadum, line 53: Looks like this test would make the above tests redundant; we could just call it something like 'should contain the correct files'.

emadum (approved):
Had one minor comment, but LGTM 👍

bindings/node/test/release.test.js
nbbeeken: Updated the test if yall wanna take a look

dariakp (approved):

PR #176, opened 2021-06-08
fix(NODE-3118): keyAltNames option not serialized

nbbeeken:
- Reintroduce options argument to `makeDataKeyContext` that accepts one key, `keyAltNames`
- Skip error message checking tests
Also,
- added clang-format as a dev dependency as its used in a format script

A number of the tests in this CI are skipped as they need to run against an actual mongodb. There is an accompanying PR I'll put up to the driver that will run these tests. (https://github.com/mongodb/node-mongodb-native/pull/2833)


nbbeeken:
@addaleax Can't request your review directly but please feel free to take a look and advise here, thanks! 😃

bindings/node/lib/cryptoCallbacks.js
```diff
@@ -87,9 +87,16 @@ function signRsaSha256Hook(key, input, output) {
   let result;
   try {
     const signer = crypto.createSign('sha256WithRSAEncryption');
-    const privateKey = Buffer.from(
-      `-----BEGIN PRIVATE KEY-----\n${key.toString('base64')}\n-----END PRIVATE KEY-----\n`
-    );
+    const privateKeyString = `-----BEGIN PRIVATE KEY-----\n${key.toString(
```
dariakp, line 90: this change shouldn't be here

bindings/node/test/clientEncryption.test.js
```diff
@@ -230,52 +234,54 @@ describe('ClientEncryption', function() {
...
     }
 
     describe('errors', function() {
       [42, 'hello', { keyAltNames: 'foobar' }, /foobar/].forEach(val => {
         it(`should fail if typeof keyAltNames = ${typeof val}`, function(done) {
-          try {
-            this.encryption.createDataKey('aws', makeOptions(val), () => {
+          const options = makeOptions(val);
+          expect(() =>
+            this.encryption.createDataKey('aws', options, () => {
```
dariakp, line 255: unnecessary check in callback; same comment on other test file

dariakp (changes requested):

bindings/node/src/mongocrypt.cc
```diff
@@ -674,6 +674,43 @@ NAN_METHOD(MongoCrypt::MakeDataKeyContext) {
...
+        if (Nan::Has(options, ALT_NAMES_KEY).FromMaybe(false)) {
+            v8::Local<v8::Value> keyAltNames = Nan::Get(options, ALT_NAMES_KEY).ToLocalChecked();
+
+            if (!(keyAltNames->IsNull() || keyAltNames->IsUndefined())) {
+                if (!keyAltNames->IsArray()) {
+                    Nan::ThrowTypeError("keyAltNames must be an array of strings");
+                    return;
+                }
+
+                v8::Local<v8::Array> keyAltNamesArray = v8::Local<v8::Array>::Cast(keyAltNames);
```
addaleax, line 689: Since you're using `Nan::To` everywhere else, for consistency?

```suggestion
                v8::Local<v8::Array> keyAltNamesArray = Nan::To<v8::Array>(keyAltNames);
```

addaleax (approved):
Looks good!

bindings/node/src/mongocrypt.cc
```diff
@@ -674,6 +674,43 @@ NAN_METHOD(MongoCrypt::MakeDataKeyContext) {
         return;
     }
 
+    if (info[1]->IsObject()) {
```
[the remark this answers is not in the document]
nbbeeken: Yes the only usage of this function is in clientEncryption.js in that constructor, so I suppose I can remove this. Just curious, is it an expensive check or just unnecessary?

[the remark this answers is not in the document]
addaleax: No, it’s a cheap check :) The conditional just made me wonder if I had missed something, that’s all.

bindings/node/src/mongocrypt.cc
nbbeeken: I'm getting a compiler error with that change: 
```
./src/mongocrypt.cc:688:53: error: no matching function for call to 'To'
            v8::Local<v8::Array> keyAltNamesArray = Nan::To<v8::Array>(keyAltNames);
                                                    ^~~~~~~~~~~~~~~~~~
../node_modules/nan/nan_converters.h:61:38: note: candidate template ignored: substitution failure [with T = v8::Array]: implicit instantiation of undefined template 'Nan::imp::ToFactory<v8::Array>'
typename imp::ToFactory<T>::return_t To(v8::Local<v8::Value> val) {
```
I also added `.ToLocalChecked()` but the error is still there, not sure what is different from the other line where we use this:
```cxx
v8::Local<v8::Object> options = Nan::To<v8::Object>(info[1]).ToLocalChecked();
```

addaleax: Ok – a little surprised that Nan only provides this for some types, but let’s just leave this as it is then :)

nbbeeken: Super, yea that is unexpected but TIL, thanks for your help!

bindings/node/lib/clientEncryption.js
```diff
@@ -202,12 +165,37 @@ module.exports = function(modules) {
...
+        keyAltNames = options.keyAltNames.map((keyAltName, i) => {
+          if (typeof keyAltName !== 'string') {
+            throw new TypeError(
+              `Option "keyAltNames" must be an array of string, but item at index ${i} was of type ${typeof keyAltName}`
+            );
+          }
+
+          return bson.serialize({ keyAltName });
+        });
+      } else if (options.keyAltNames && !Array.isArray(options.keyAltNames)) {
```
dariakp, line 191: Recommend moving this check up to 179

bindings/node/lib/clientEncryption.js
```diff
@@ -202,12 +165,37 @@ module.exports = function(modules) {
...
 
       const bson = this._bson;
-      options = sanitizeDataKeyOptions(bson, options);
-      const dataKeyBson = bson.serialize(Object.assign({ provider }, options.masterKey));
-      const context = this._mongoCrypt.makeDataKeyContext(dataKeyBson);
+
+      const dataKey = Object.assign({ provider }, options.masterKey);
+
+      let keyAltNames;
+      if (Array.isArray(options.keyAltNames) && options.keyAltNames.length > 0) {
```
dariakp, line 181: the check for the length makes it so that `keyAltNames` is `undefined` if an empty options array is passed in; is that the desired behavior?

dariakp (changes requested):

bindings/node/lib/clientEncryption.js
nbbeeken: Done!

bindings/node/lib/clientEncryption.js
nbbeeken: Yep, the c++ checks for undefined before using this option so I figure if there's nothing in the array there's nothing of relevance to pass in. It would be equivalent to pass down an empty array.

bindings/node/lib/clientEncryption.js
```diff
@@ -202,12 +165,39 @@ module.exports = function(modules) {
      * });
      */
     createDataKey(provider, options, callback) {
-      if (typeof options === 'function') (callback = options), (options = {});
+      if (typeof options === 'function') {
```
mbroadst, line 168: I think the majority of the changes in this file are whitespace/style changes, and it made it really confusing to review. It looks like you could make a one-line change on the original L210 to:
```
const context = this._mongoCrypt.makeDataKeyContext(dataKeyBson, options);
```
and with your C++ changes the `keyAltNames` would be applied. Is that correct?

bindings/node/src/mongocrypt.cc
```diff
@@ -674,6 +674,41 @@ NAN_METHOD(MongoCrypt::MakeDataKeyContext) {
         return;
     }
 
+    v8::Local<v8::Object> options = Nan::To<v8::Object>(info[1]).ToLocalChecked();
+    auto ALT_NAMES_KEY = Nan::New("keyAltNames").ToLocalChecked();
+    if (Nan::Has(options, ALT_NAMES_KEY).FromMaybe(false)) {
+        v8::Local<v8::Value> keyAltNames = Nan::Get(options, ALT_NAMES_KEY).ToLocalChecked();
+
+        if (!(keyAltNames->IsNull() || keyAltNames->IsUndefined())) {
```
mbroadst, line 682: nit: can we replace this if statement with `if (!keyAltNames->IsArray()) { // bail with error }` instead of checking for !null, !undefined, isArray?

mbroadst (changes requested):
Dang, this is a really subtle bug sorry to see it slipped through. In libmongocrypt 1.x we introduced [mongocrypt_ctx_setopt_key_encryption_key](https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in#L626-L668) which was meant to supersede all of the other `setopt` methods. It's not clear to me why `keyAltNames` was omitted from that list, and I think an argument could be made that it should support it.

bindings/node/lib/clientEncryption.js
nbbeeken: Indeed the changes could've been less than what's here. I standby the decision to make things easier to read/debug for the next engineer. Is there something specific that you think is less optimal here?

bindings/node/src/mongocrypt.cc
nbbeeken: Done! with slight change where if the input isn't an array it is ignored, the error handling down here is a duplication of what we do in the JS.

dariakp (approved):

bindings/node/lib/clientEncryption.js
mbroadst: I won't comment on whether these changes subjectively improve readability/debugability for future developers, but I can say for sure that changing it this way makes it harder to review and easier to introduce new bugs.

I anchored this comment to splitting the one-liner into a few lines, but I was more concerned about `sanitizeDataKeyOptions`. It _seems_ like you copied the contents into this function with some subtle changes (for-loop => map I think was one change). I read it a few times, but I couldn't tell you for sure whether there's a new bug there or not, or that the bug that I see in the old code is gone for sure.

If you're confident that it's a net improvement in code quality that's fine, but I'd ask that you consider the implications in future changes

nbbeeken: Thanks for the insight, I do agree that when addressing a bug fix it would be ideal to have just what is necessary to represent the fix in isolation from any refactor. Unfortunately this fix also required orchestrating new CI tasks/scripts to prove the code changes were, in fact, effective which required some debugging. Making the sanitization more functional was helpful as opposed to the key deletion and new array creating that then overwrote the old one logic.  
[We do have tests covering this usage here which tests for the proper errors as well](https://evergreen.mongodb.com/task/mongo_node_driver_ubuntu1804_custom_csfle_tests_run_custom_csfle_tests_51e976861ebc4a0e8b8501719656effeb7ed522e_21_06_22_14_05_14).

mbroadst (approved):

PR #179, opened 2021-07-02
Ensure Java bindings handle multiple calls MongoCrypt#close

rozza:
MONGOCRYPT-322


jyemin (approved):
This LGTM, but let's make a similar change in the driver code so that people who upgrade the driver but not mongodb-crypt also get a fix.

stIncMale (approved):

PR #186, opened 2021-08-19
MONGOCRYPT-330 fix and validate pkg-config scripts

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/61185a2de3c3316a179d32ad/tasks

Note that I tested by downloading the built packages from one of the tasks that builds packages and then validating that the pkg-config script which would be installed from the PPA no longer references the pkg-config module for libbson.


CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ option (ENABLE_PIC
    "Enables building of position independent code for static library components."
    ON
 )
+option (ENABLE_BUILD_FOR_PPA "Maintainer-only option for preparing PPA build" OFF)
+if (ENABLE_SHARED_BSON AND ENABLE_BUILD_FOR_PPA)
```
kevinAlbs, line 19: In refreshing myself on static/shared linking from prior conversations. From [this prior conversation](https://github.com/mongodb/libmongocrypt/pull/111/files#r497556530), a consumer linking to a shared libmongocrypt does not need to link to libbson.

I understand that to mean that `libmongocrypt.pc` does not need to specify `libbson-static-1.0` under `Requires`. But `libmongocrypt-static.pc` does need to specify `libbson-static-1.0` under `Requires`.

Is that correct? If so, why does this add `PPA` as a special case?

debian/rules
```diff
@@ -25,6 +25,7 @@ export DEB_CFLAGS_MAINT_APPEND  = -fPIC
 override_dh_auto_configure:
 	dh_auto_configure -B$(CURDIR)/cmake-build -- \
 	-DBUILD_VERSION=$(DEB_VERSION_UPSTREAM) \
+	-DENABLE_SHARED_BSON=ON \
```
kevinAlbs, line 28: Is linking to libbson shared necessary? I do not understand how this relates to the reported issue.

ShaneHarvey:
FYI we're looking into the Python macos failure in https://jira.mongodb.org/browse/PYTHON-2889.

.evergreen/pkgconfig-tests.sh
```diff
@@ -46,25 +46,33 @@ SRC_PATH="$(system_path ../)"
...
-# Build libmongocrypt, static linking against libbson
+# Build libmongocrypt, static linking against libbson and configured for the PPA
 cd $pkgconfig_tests_root/libmongocrypt-cmake-build
 PREFIX_PATH="$(system_path $pkgconfig_tests_root/install)"
 INSTALL_PATH="$(system_path $pkgconfig_tests_root/install/libmongocrypt)"
 SRC_PATH="$(system_path $libmongocrypt_root)"
-$CMAKE -DENABLE_SHARED_BSON=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS -DCMAKE_PREFIX_PATH="$PREFIX_PATH" -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" "$SRC_PATH"
+$CMAKE -DENABLE_SHARED_BSON=OFF -DENABLE_BUILD_FOR_PPA=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS -DCMAKE_PREFIX_PATH="$PREFIX_PATH" -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" "$SRC_PATH"
 $CMAKE --build . --target install --config RelWithDebInfo
+find ${PREFIX_PATH} -name libbson-static-1.0.a -execdir cp {} $(dirname $(find ${INSTALL_PATH} -name libmongocrypt-static.a )) \;
```
kevinAlbs, line 56: IIUC this copies libbson-static-1.0.a into the install directory of libmongocrypt. Why is that necessary?

.evergreen/pkgconfig-tests.sh
```diff
@@ -46,25 +46,33 @@ SRC_PATH="$(system_path ../)"
...
+
+# To validate the pkg-config scripts, we don't want the libbson script to be visible
```
kevinAlbs, line 58: Ah nice. IIUC if there was still a pkg-config `Requires` dependency on `libbson-static-1.0`, lines 62 and 63 would error since the PKG_CONFIG_PATH does not include the path to libbson-1.0.pc.

CMakeLists.txt
kevinAlbs: OK, thank you for the explanation. I had forgotten that `libbson-dev` was not part of the PPA. I still think we can remove the `Requires: libbson-static-1.0` from `libmongocrypt.pc` (but not `libmongocrypt-static.pc`. But I think that can be out of scope of this ticket.

debian/rules
kevinAlbs: Thank you for the explanation, I think it is OK to include as part of this commit.

kevinAlbs (approved):
LGTM! Thank you for your patience with my questions. I have a few more to check my understanding.

I drafted a document [libmongocrypt linking](https://docs.google.com/document/d/1YnRDmNoBZG_Ip4JmIw9bB4tItFt5mIBHLu4Ex73WSGY/edit#) with open comments to continue the discussion about possibly overlinking libbson. But I believe this PR solves the reported issue, and we can continue the discussion there.

.evergreen/pkgconfig-tests.sh
rcsanchez97: Because we now install the libbson and libmongocrypt builds into separate directories (to make sure the packager script does not pick up extraneous components), a program building with the `libmongocrypt.pc` module would need the `libbson-static-1.0.a` library to be located there or the build will fail:

```
[2021/09/02 22:23:10.588] + gcc -DBSON_STATIC -DMONGOCRYPT_STATIC_DEFINE -DKMS_MSG_STATIC -I/data/mci/951753fb2fbc9d90761587fce32befe3/libmongocrypt/pkgconfig_tests/install/include/libbson-1.0 -I/data/mci/951753fb2fbc9d90761587fce32befe3/libmongocrypt/pkgconfig_tests/install/libmongocrypt/include/mongocrypt -o example-state-machine test/example-state-machine.c /data/mci/951753fb2fbc9d90761587fce32befe3/libmongocrypt/pkgconfig_tests/install/libmongocrypt/lib64/libmongocrypt-static.a /data/mci/951753fb2fbc9d90761587fce32befe3/libmongocrypt/pkgconfig_tests/install/libmongocrypt/lib64/libkms_message-static.a /data/mci/951753fb2fbc9d90761587fce32befe3/libmongocrypt/pkgconfig_tests/install/libmongocrypt/lib64/libbson-static-1.0.a /usr/lib64/librt.so -pthread -lpthread -lssl -lcrypto
[2021/09/02 22:23:10.588] gcc: error: /data/mci/951753fb2fbc9d90761587fce32befe3/libmongocrypt/pkgconfig_tests/install/libmongocrypt/lib64/libbson-static-1.0.a: No such file or directory
```

Depending on what we determine in terms of whether including `libbson-static-1.0.a` in the libmongocrypt PPA distribution and then requiring consumers to link directly to `libbson-static-1.0.a` constitutes overlinking, this may become unnecessary at some future point.

.evergreen/pkgconfig-tests.sh
rcsanchez97: That's correct.  And since the only BSON component we want to ship is the `libbson-static-1.0.a` library, the actual library file needs to appear on the `Libs:` line in the `.pc` script since there is no `libbson-static-1.0.pc` script to provide the location of the library archive.

rcsanchez97:
@kevinAlbs Relative to your comment:

> OK, thank you for the explanation. I had forgotten that libbson-dev was not part of the PPA. I still think we can remove the Requires: libbson-static-1.0 from libmongocrypt.pc (but not libmongocrypt-static.pc. But I think that can be out of scope of this ticket.

(For some reason GitHub won't let me respond to it inline.)

Comparing the CMake scripts for reference, the `mongocrypt` target has no reference to the `bson_static` target, though the `mongocrypt_static` target does have such a reference.  That makes it seem likely that your suggested change could be made without any negative impact to any libmongocrypt consumer.  In order to confirm, we would need a test or example program that used only libmongocrypt components without any direct reference to libbson.  The current example, `example-state-machine.c`, directly references libbson and the build won't succeed without pointing the example program to libbson, which means that it would mask any failure that we would be trying to trigger by removing `libbson-static-1.0` from the `Requires`.

rcsanchez97:
Pushed as 6af46e4 and 2d3d1ca

PR #193, opened 2021-09-15
feat!(NODE-3474): convert to Node-API

addaleax:


dariakp (commented):
Nothing here stands out to me as being unreasonable, but I don't have much knowledge about this library or what potential impacts of this specific change could be - for that matter, I could use some context about what our goal here is (i.e., why we're doing this), without that I don't think I can provide any meaningful review @nbbeeken @addaleax

addaleax:
@dariakp Fwiw, as far as goals and advantages are concerned, the [scope doc](https://docs.google.com/document/d/1vPcodLFbQR3t9EWo8W7xhWLi0k8HrrA6cC6AOdXwbng/edit) has some notes on that.

The tl;dr is:
- Node-API is *the* way to do modern Node.js addons 
- Code written using Node-API is generally more readable/accessible to JS developers than code written using V8/Nan
- Code written using Node-API is easier to get right (almost all of the `.FromMaybe()`, `.ToLocalChecked()` handling in the current code is not really valid)
- Node-API means that addons generally follow better structural rules, such as avoiding storing JS constructors in global variables (which inherently clashes with thread safety)
- There is only one native binary produced, instead of one for each Node.js version (and this binary continues to work with future Node.js versions)
- The API for Node-API has fewer breaking changes across Node.js versions (none that I am aware of so far, actually)
- This *is* a breaking change as it affects Node.js version compatibility; in the scope doc, we set 12.9.0 as our goal (which I have manually tested as well)

Specifically, for us as the devtools team, not having to worry about breaking changes between Node.js versions is nice, as well as the fact that we can use the same precompiled binaries for Electron and Node.js (or at least we *should* be able to do that).

nbbeeken:
- [Patch testing against 4.x driver](https://spruce.mongodb.com/version/614203da57e85a4b9aef6291/tasks)
- [Patch testing against 3.x driver](https://spruce.mongodb.com/version/6142049361837d4ce4595a8d/tasks)

addaleax:
@nbbeeken I think those might have failed because I’m PRing this from a fork, not this repo here (due to lack of write access to this repo)?

bindings/node/package.json
```diff
@@ -67,7 +66,12 @@
     "mongodb": ">=3.4.0"
   },
   "engines": {
-    "node": ">=6"
+    "node": ">=12.9.0"
+  },
+  "binary": {
+    "napi_versions": [
+      4
```
nbbeeken, line 73: Should this be 6? Also "binary" doesn't appear to be an npm field, just curious what uses this

bindings/node/src/mongocrypt.cc
```diff
@@ -263,656 +233,551 @@ bool MongoCrypt::setupCryptoHooks(mongocrypt_t* mongoCrypt, CryptoHooks* cryptoH
...
-    Nan::SetPrototypeMethod(tpl, "addMongoOperationResponse", AddMongoOperationResponse);
-    Nan::SetPrototypeMethod(tpl, "finishMongoOperation", FinishMongoOperation);
-    Nan::SetPrototypeMethod(tpl, "nextKMSRequest", NextKMSRequest);
-    Nan::SetPrototypeMethod(tpl, "finishKMSRequests", FinishKMSRequests);
-    Nan::SetPrototypeMethod(tpl, "finalize", Finalize);
+// Store callbacks as nested properties on the MongoCrypt binding object
+// itself, and use these helpers to do so. Storing them as JS engine
+// References is a big memory leak footgun.
+Function MongoCrypt::GetCallback(const char* name) {
+    InstanceData* instance_data = Env().GetInstanceData<InstanceData>();
```
nbbeeken, line 595: Does fetching this have some side effect? its being reported as unused.

nbbeeken (commented):
Small Qs

bindings/node/src/mongocrypt.cc
addaleax: Err, yeah, thanks for catching. It’s a leftover from a previous version.

bindings/node/package.json
addaleax: prebuild told me to do this :) https://www.npmjs.com/package/prebuild#n-api-considerations

addaleax: Also, eh, I’m not sure. 4 is what we settled on – having the instance data feature available is *much* nicer though, and as the comments in the source indicate, it is available in all Node.js versions that have N-API version 4 (just not as stable yet – the feature has not been changed since it was experimental, so essentially we can consider it stable from the beginning)

nbbeeken: I think that's fine to rely on, but just so I understand in the header file we have:
```
#define NAPI_VERSION 6
#define NAPI_EXPERIMENTAL
```
should it be `NAPI_VERSION 4` and keep the define `NAPI_EXPERIMENTAL` instead? Essentially we're saying compiling against 4 with the experimental instance data feature would be the same as 6 but we still want to target 4. IIUC

And related to prebuild it says we should throw a define in the [node-gyp file](https://www.npmjs.com/package/prebuild#node-gyp), but since we're already successfully compiling this seems like a best practice thing?

dariakp:
> @dariakp Fwiw, as far as goals and advantages are concerned, the [scope doc](https://docs.google.com/document/d/1vPcodLFbQR3t9EWo8W7xhWLi0k8HrrA6cC6AOdXwbng/edit) has some notes on that.
> 
> The tl;dr is:
> 
> * Node-API is _the_ way to do modern Node.js addons
> * Code written using Node-API is generally more readable/accessible to JS developers than code written using V8/Nan
> * Code written using Node-API is easier to get right (almost all of the `.FromMaybe()`, `.ToLocalChecked()` handling in the current code is not really valid)
> * Node-API means that addons generally follow better structural rules, such as avoiding storing JS constructors in global variables (which inherently clashes with thread safety)
> * There is only one native binary produced, instead of one for each Node.js version (and this binary continues to work with future Node.js versions)
> * The API for Node-API has fewer breaking changes across Node.js versions (none that I am aware of so far, actually)
> * This _is_ a breaking change as it affects Node.js version compatibility; in the scope doc, we set 12.9.0 as our goal (which I have manually tested as well)
> 
> Specifically, for us as the devtools team, not having to worry about breaking changes between Node.js versions is nice, as well as the fact that we can use the same precompiled binaries for Electron and Node.js (or at least we _should_ be able to do that).

Thanks! The scope doc helps with the background :)

bindings/node/package.json
addaleax: > should it be `NAPI_VERSION 4` and keep the define `NAPI_EXPERIMENTAL` instead? Essentially we're saying compiling against 4 with the experimental instance data feature would be the same as 6 but we still want to target 4. IIUC

The problem is that in current Node.js versions, where the feature is no longer considered experimental, it’s only available with `NAPI_VERSION >= 6`.

> And related to prebuild it says we should throw a define in the [node-gyp file](https://www.npmjs.com/package/prebuild#node-gyp), but since we're already successfully compiling this seems like a best practice thing?

That’s for defining the `NAPI_VERSION` in the source code – it’s what these defines here are already doing :)

nbbeeken: Oh ok I see, thanks for clarifying. And yea I also saw that can be used to conditionally change the NAP_VERSION so, no need!

nbbeeken (approved):

dariakp (approved):

PR #195, opened 2021-09-20
MONGOCRYPT-334 clone mongo-c-driver from reference in all tests

kevinAlbs:
Resolves test failures on RHEL 6.

libmongocrypt pins to 1.17.0 of the C driver when building in prep_c_driver_source.sh.

But linker-tests.sh and pkgconfig-tests.sh work off of the tip of master for the C driver.

libmongocrypt tests does not compile on RHEL 6 / gcc 4.4 due to a dependency on atomic builtins introduced in CDRIVER-4124.

This updates tests to use C driver 1.17.0.


.evergreen/clone-mongo-c-driver.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -o xtrace
+set -o errexit
+
+MONOGO_C_DRIVER_REF=1.17.0
+
+# Force checkout with lf endings since .sh must have lf, not crlf on Windows
+git clone git@github.com:mongodb/mongo-c-driver.git --config core.eol=lf --config core.autocrlf=false
```
vector-of-bool, line 9: If the ref will always be a tag or a branch name, it will be better to check out the branch directly with `--branch=${ref} --depth=1`, unless we need the entire repo history for some reason. It depends on if the mongo-c-driver actually takes a long time to clone or not.

(Unless this is also using a very very very old Git version that doesn't know how to do `--branch` clone. That has bitten me in the past, and even better that it would just issue a warning and clone from remote `HEAD` as if nothing was wrong. That was quite a few hours of wasted time. The repo history in question was so absurdly large that it was faster to build a newer `git` from source on-the-fly as part of the build so that I could do the shallow-clone. This anecdote is almost entirely irrelevant, but I felt like sharing here.)

.evergreen/prep_c_driver_source.sh
```diff
@@ -3,12 +3,8 @@
 set -o xtrace
 set -o errexit
 
-MONOGO_C_DRIVER_REF=1.17.0
-
-# Force checkout with lf endings since .sh must have lf, not crlf on Windows
-git clone git@github.com:mongodb/mongo-c-driver.git --config core.eol=lf --config core.autocrlf=false
+./libmongocrypt/.evergreen/clone-mongo-c-driver.sh
```
vector-of-bool, line 6: Should this also use `$libmongocrypt_root`?

```suggestion
$libmongocrypt_root/.evergreen/clone-mongo-c-driver.sh
```

vector-of-bool (approved):
LGTM, just one question and one comment to consider.

.evergreen/prep_c_driver_source.sh
kevinAlbs: Done.

.evergreen/clone-mongo-c-driver.sh
kevinAlbs: Oh good point, done.

Oh that is good to know. The last [patch build](https://evergreen.mongodb.com/version/614a14a20ae60674e6363b01) succeeded. I checked Ubuntu 14.04 and RHEL 6 from spawn hosts. Both have new enough git to support clone with `--branch`.

.evergreen/compile.sh
```diff
@@ -16,7 +16,7 @@ set -o errexit
 save_pwd="$(pwd)"
 
 . ./libmongocrypt/.evergreen/setup-env.sh
-MONGO_C_DRIVER_VERSION=1.16.0-pre . ./libmongocrypt/.evergreen/prep_c_driver_source.sh
```
kevinAlbs, line 19: This was inaccurate. `prep_c_driver_source.sh` was cloning from 1.17.0 before.

.evergreen/linker-tests.sh
```diff
@@ -58,7 +58,7 @@ mkdir cmake-build
 cd cmake-build
 INSTALL_PATH="$(system_path $linker_tests_root/install/bson1)"
 SRC_PATH="$(system_path ../)"
-$CMAKE -DBUILD_VERSION=1.18.0-pre -DENABLE_MONGOC=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" "$SRC_PATH"
```
kevinAlbs, line 61: It is not necessary to specify `BUILD_VERSION` since this is included in the `VERSION_CURRENT` file.

kevinAlbs (commented):
Consolidated additional hard coded C driver versions. Removed clone-mongo-c-driver.sh.

vector-of-bool (approved):
LGTM

PR #197, opened 2021-10-03
MONGOCRYPT-336 move key wrap and unwrap functions

kevinAlbs:
See MONGOCRYPT-336 for more information. There are no functional changes. This is in preparation of the key wrap / unwrap functions being used by the upcoming KMIP KMS provider.

New task failures on the Debian 9.2, Linux Distro Packaging, and SLES 12 64-bit variants appear due to BUILD-14027.

### DIFF
--- a/src/mongocrypt-crypto.c
+++ b/src/mongocrypt-crypto.c
@@ -28,6 +28,8 @@
 #include "mongocrypt-private.h"
 #include "mongocrypt-status-private.h"
 
+#include <inttypes.h>
+
 /* Crypto primitives. These either call the native built in crypto primitives or
  * user supplied hooks. */
 static bool
--- a/src/mongocrypt-crypto.c
+++ b/src/mongocrypt-crypto.c
@@ -912,3 +912,83 @@ done:
    _mongocrypt_buffer_cleanup (&to_hmac);
    return ret;
 }
+
+bool
+_mongocrypt_wrap_key (_mongocrypt_crypto_t *crypto,
+                      _mongocrypt_buffer_t *kek,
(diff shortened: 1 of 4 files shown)


src/mongocrypt-crypto.c
```diff
@@ -911,4 +911,84 @@ _mongocrypt_calculate_deterministic_iv (
...
+                                   status)) {
+      return false;
+   }
+   dek->len = bytes_written;
+
+   if (dek->len != MONGOCRYPT_KEY_LEN) {
+      CLIENT_ERR ("decrypted key is incorrect length, expected: %" PRIu32
+                  ", got: %" PRIu32,
+                  MONGOCRYPT_KEY_LEN,
+                  dek->len);
```
eramongodb, line 990: Note: `PRIu32` and other format macro constants for fixed-width integral types are a C99 feature. The `<inttypes.h>` header should be included per "include what you use". Prior usage of `<inttypes.h>` entities in libmongocrypt are exclusively in test code ([test-conveniences.c](https://github.com/mongodb/libmongocrypt/blob/fcf716b100a164a0fc845081f6d3afc4d0ef6ccd/test/util/util.c#L624), [util.c](https://github.com/mongodb/libmongocrypt/blob/fcf716b100a164a0fc845081f6d3afc4d0ef6ccd/test/test-conveniences.c#L724)).

eramongodb (approved):

markbenvenuto (approved):

src/mongocrypt-crypto.c
kevinAlbs: Good spot. Added `#include <inttypes.h>`.

PR #198, opened 2021-10-03
MONGOCRYPT-337 separate and fix up test assertions and utilities

kevinAlbs:
# Summary of changes

## kms-message
- Move test assertions to a new file: `kms_test_assert.h`
- Move test utilities to a new file: `kms_test_util.h`.
- Make `target_include_directories` of `test_kms_azure_online` and `test_kms_gcp_online` consistent with `test_kms_request`
- Support custom TLS options in `test_kms_online_util.h`.
    - Rationale: this is needed for upcoming KMIP KMS tests. The KMIP test server needs a custom CA and client certificate to authenticate.

## libmongocrypt
- Move test assertions to a new file: `test-mongocrypt-assert.h`
- Move test utilities to a new file: `test-mongocrypt-util.h`.
- Rename `test-conveniences.h` to `test-mongocrypt-assert-match-bson.h`.
- Support custom TLS options in `util/csfle.c`.

## General improvements
- Include the file/function/line in assertion macros where possible.

## Rejected ideas
- Reuse assertions / utilities between libmongocrypt and kms-message.
  - Assertions like `ASSERT_CMPBYTES` are duplicated in `test-mongocrypt-assert.h` and `test_kms_assert.h`. As is the utility `data_to_hex`.
  - I chose to keep them separate to avoid complexity.

# Rationale
These changes are derived from a POC implementation of KMIP support: https://github.com/kevinAlbs/libmongocrypt/tree/kmip_tracer_bullet

The intent is to separate the test changes from any function changes for KMIP support.


kms-message/test/test_kmip_reader_writer.c
```diff
@@ -1,9 +1,8 @@
-#include "test_kms_request.h"
+#include "test_kms_assert.h"
```
eramongodb, line 1: Missing copyright notice?

test/test-mongocrypt-util.c
```diff
@@ -0,0 +1,58 @@
+#include "test-mongocrypt-util.h"
```
eramongodb, line 1: Missing copyright notice?

eramongodb (changes requested):
I have maintainability concerns regarding (lack of) symmetry between assertion macros as proposed and those that already exist in both libmongocrypt and the C driver. What is the rationale behind rejecting common reuse of assertion helpers across libmongocrypt and kms-message? The need for additional complexity and discrepancies in the assertion macros in `test_kms_assert.h` relative to those in `test-mongocrypt-assert.h` is not immediately apparent to me (e.g. calls to `kms_request_str_new_from_chars()`).

kevinAlbs:
> I have maintainability concerns regarding (lack of) symmetry between assertion macros as proposed and those that already exist in both libmongocrypt and the C driver. What is the rationale behind rejecting common reuse of assertion helpers across libmongocrypt and kms-message?

My main concern for doing it now is timing. There is no existing shared test code between kms-message and libmongocrypt. I did not know how to do it easily:

- Test assertions in kms-message already rely on API private in kms-message. E.g. `ASSERT_CONTAINS` uses `kms_request_str_t`.
- Test assertions in libmongocrypt already rely on API that kms-message cannot access. E.g. `bson_equal`. kms-message does not have a dependency on libbson.

> (e.g. calls to `kms_request_str_new_from_chars()`).

That is a fair point. `test-mongocrypt-assert.h` defines `ASSERT_STRCONTAINS` with `strstr` while `test_kms_assert.h` defines `ASSERT_CONTAINS` with `kms_request_str` to convert it to lowercase first.

I am happy to file a MONGOCRYPT ticket to share the test assertions and utilities between libmongocrypt and kms-message. I hoped moving assertions/test utilities to separate files improved the status quo. But if it does not, I would rather revert the movement and keep assertions / utilities in their current places.

eramongodb (approved):
> There is no existing shared test code between kms-message and libmongocrypt.

If that is the case, I am alright with preserving this separation of concerns.

> I hoped moving assertions/test utilities to separate files improved the status quo.

When kms-message and libmongocrypt are viewed in isolation from one another, I do believe these changes are an improvement. I am in favor of continuing with the changes as proposed.

kms-message/CMakeLists.txt
```diff
@@ -260,16 +261,14 @@ if(NOT mongoc-1.0_FOUND)
...
+   target_include_directories(test_kms_azure_online PRIVATE ${PROJECT_SOURCE_DIR})
    target_compile_definitions(test_kms_azure_online PRIVATE ${KMS_MESSAGE_DEFINITIONS})
    target_link_libraries(test_kms_azure_online mongo::mongoc_shared)
    target_link_libraries(test_kms_azure_online kms_message_static)
 
-   add_executable(test_kms_gcp_online test/test_kms_gcp_online.c test/test_kms.c test/test_kms_online_util.c)
-   target_include_directories(test_kms_gcp_online PRIVATE ${PROJECT_SOURCE_DIR}/src)
-   target_include_directories(test_kms_gcp_online PRIVATE ${PROJECT_SOURCE_DIR}/src/kms_message)
+   add_executable(test_kms_gcp_online test/test_kms_gcp_online.c test/test_kms_util.c test/test_kms_online_util.c)
+   target_include_directories(test_kms_gcp_online PRIVATE ${PROJECT_SOURCE_DIR})
```
vector-of-bool, line 271: I have (strong) opinions that the `<root>/src` directory should be the root of the include-search path for all libraries.

```suggestion
   target_include_directories(test_kms_gcp_online PRIVATE ${PROJECT_SOURCE_DIR}/src)
```

I do strongly agree with the removal of `src/kms_message` as a search dir, though.

vector-of-bool (approved):
Looks good. My only comment is a build-related concern.

kms-message/CMakeLists.txt
kevinAlbs: Done, on this and all kms-message test targets.

PR #201, opened 2021-10-06
chore(NODE-3632): limit CI publishing to one version of x64 linux

nbbeeken:


bindings/node/.evergreen/prebuild.sh
```diff
@@ -1,35 +1,52 @@
...
-elif [[ $VERSION_AT_HEAD != $VERSION_AT_HEAD_1 ]]; then
+elif [[ "$VERSION_AT_HEAD" != "$VERSION_AT_HEAD_1" ]]; then
   echo "Difference is package version ($VERSION_AT_HEAD_1 -> $VERSION_AT_HEAD)"
   echo "Beginning prebuild"
-  run_prebuild
+
+  if [[ "$OS" == "LINUX" ]]; then
+    # We're on linux should be able to use uname -m
+    ARCH=$(uname -m)
+    if [[ "$ARCH" == "x86_64" && $DISTRO_ID == "rhel70-small" ]]; then
```
addaleax, line 41: Fwiw, I mentioned `$DISTRO_ID` because it already implies the specfific architecture – different architectures will have different distro ids, x86_64 just happens to be the implied default

addaleax: Yeah, makes sense – the conditional here is redundant because its second part implies the first part.

Also, I do think you want to have a final `else` that is just `npm run prebuild`, like the `Will prebuild without submit` part below?

nbbeeken: Yea that makes sense, I think I've cleaned it up to be clearer now 🙂

addaleax (approved):

PR #202, opened 2021-10-06
MONGOCRYPT-332 add KMIP requests

kevinAlbs:
**Summary**
- Add KMIP Register, Get, and Activate requests.
- Register uses a 96 byte SecretData. This acts as the key encryption key (KEK).
- The requests are based on the [KMIP spec version 1.4](https://docs.oasis-open.org/kmip/spec/v1.4/os/kmip-spec-v1.4-os.html).

**Open questions**
1. The `Register` request uses `SecretDataType` of `Seed` from [Section 9.1.3.2.9](http://docs.oasis-open.org/kmip/spec/v1.4/os/kmip-spec-v1.4-os.html#_Toc490660929). This was arbitrary. I considered an extension value, but using `0x80000001` resulted in an error from a PyKMIP test server. AFAIK the choice of `SecretDataType` will not impact our use of it. We only need to retrieve the 96 byte data.

**Etc**
XML comments in the test data was generated with [kmip_dump](https://github.com/kevinAlbs/c-bootstrap/tree/master/tools/kmip_dump), a small debug utility. I do not intend to put `kmip_dump` in review to keep the changes for KMIP support small.


kms-message/src/kms_message_private.h
```diff
@@ -46,6 +46,20 @@ struct _kms_request_t {
    bool auto_content_length;
    _kms_crypto_t crypto;
    kms_request_provider_t provider;
+
+   /* TODO (MONGOCRYPT-???): make a union for each KMS provider type.
```
kevinAlbs, line 50: I think a union is appropriate for this. I did not want to modify existing fields to keep the diff small. If there is consensus, I will create a MONGOCRYPT ticket and update the comment.

kms-message/src/kms_kmip_request.c
```diff
@@ -0,0 +1,233 @@
...
+     <BatchCount tag="0x42000d" type="Integer" value="1"/>
+    </RequestHeader>
+    <BatchItem tag="0x42000f" type="Structure">
+     <Operation tag="0x42005c" type="Enumeration" value="3"/>
+     <RequestPayload tag="0x420079" type="Structure">
+      <ObjectType tag="0x420057" type="Enumeration" value="7"/>
+      <TemplateAttribute tag="0x420091" type="Structure">
+      </TemplateAttribute>
+      <SecretData tag="0x420085" type="Structure">
+       <SecretDataType tag="0x420086" type="Enumeration" value="2"/>
```
kevinAlbs, line 58: I was unsure whether to use `Password` or `Seed` for the `SecretDataType`. http://docs.oasis-open.org/kmip/spec/v1.4/os/kmip-spec-v1.4-os.html#_Toc490660929 I chose `Seed` arbitrarily.

Attempting to use an extension value 0x800001 resulted in an error on a PyKMIP test server. cc @markbenvenuto in case you foresee a problem with using `Seed`.

kms-message/src/kms_message/kms_kmip_request.h
```diff
@@ -0,0 +1,56 @@
...
+#include "kms_message_defines.h"
+#include "kms_request.h"
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* kms_request_register_secretdata_new creates a KMIP Register request with
```
eramongodb, line 29:
```suggestion
/* kms_kmip_request_register_secretdata_new creates a KMIP Register request with
```

kms-message/src/kms_message/kms_kmip_request.h
```diff
@@ -0,0 +1,56 @@
...
+kms_kmip_request_register_secretdata_new (void *reserved,
+                                          uint8_t *data,
+                                          uint32_t len);
+
+/* kms_request_activate_new creates a KMIP Activate request with the
+ * provided unique identifer.
+ * - unique_identifier must be a NULL terminated string.
+ * - Callers must check for an error by calling kms_request_get_error. */
+KMS_MSG_EXPORT (kms_request_t *)
+kms_request_activate_new (void *reserved, char *unique_identifier);
```
eramongodb, line 43:
```suggestion
/* kms_kmip_request_activate_new creates a KMIP Activate request with the
 * provided unique identifer.
 * - unique_identifier must be a NULL terminated string.
 * - Callers must check for an error by calling kms_request_get_error. */
KMS_MSG_EXPORT (kms_request_t *)
kms_kmip_request_activate_new (void *reserved, const char *unique_identifier);
```
Consistency with other KMIP request functions.

Parameter `unique_identifier` should be `const char *` as function does not modify argument.

kms-message/src/kms_kmip_request.c
```diff
@@ -0,0 +1,233 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kms_kmip_reader_writer_private.h"
+#include "kms_message/kms_kmip_request.h"
+#include "kms_message_private.h"
+
```
eramongodb, line 20:
```suggestion
#include "kms_message/kms_kmip_request.h"

#include "kms_message_private.h"
#include "kms_kmip_reader_writer_private.h"

```
Header file should be first include of corresponding source file.

kms-message/src/kms_kmip_request.c
```diff
@@ -0,0 +1,233 @@
...
+static void
+copy_writer_buffer (kms_request_t *req, kmip_writer_t *writer) {
+   const uint8_t *buf;
+   size_t buflen;
+
+   buf = kmip_writer_get_buffer (writer, &buflen);
+   req->kmip.data = malloc (buflen);
+   memcpy (req->kmip.data, buf, buflen);
+   req->kmip.len = (uint32_t) buflen;
```
eramongodb, line 32: Note: I would like narrowing casts to be preceeded by a check or assertion to avoid unexpected implementation-defined runtime behavior, but this may be better suited for a project-wide audit of narrowing, signed, and implicit conversions, likely as part of a warning reduction effort.

kms-message/src/kms_message/kms_kmip_request.h
```diff
@@ -0,0 +1,56 @@
...
+
+/* kms_kmip_request_get_new creates a KMIP Get request with the provided unique
+ * identifer.
+ * - unique_identifier must be a NULL terminated string.
+ * - Callers must check for an error by calling kms_request_get_error. */
+KMS_MSG_EXPORT (kms_request_t *)
+kms_kmip_request_get_new (void *reserved, char *unique_identifier);
```
eramongodb, line 50:
```suggestion
kms_kmip_request_get_new (void *reserved, const char *unique_identifier);
```
Parameter unique_identifier should be const char * as function does not modify argument.

kms-message/test/test_kms_kmip_request.c
```diff
@@ -0,0 +1,216 @@
...
+kms_kmip_request_get_test (void)
+{
+   kms_request_t *req;
+
+   uint8_t *actual_bytes;
+   uint32_t actual_len;
+   uint8_t expected_bytes[] = {GET_REQUEST};
+   uint32_t expected_len = sizeof (expected_bytes);
+
+   req = kms_kmip_request_get_new (NULL, GET_UNIQUE_IDENTIFIER);
```
eramongodb, line 153:
```suggestion
   static const char *const GET_UNIQUE_IDENTIFIER =
      "7FJYvnV6XkaUCWuY96bCSc6AuhvkPpqI";

   req = kms_kmip_request_get_new (NULL, GET_UNIQUE_IDENTIFIER);
```
I recommend using core language over preprocessor when able.

kms-message/test/test_kms_kmip_request.c
```diff
@@ -0,0 +1,216 @@
...
+kms_kmip_request_activate_test (void)
+{
+   kms_request_t *req;
+
+   uint8_t *actual_bytes;
+   uint32_t actual_len;
+   uint8_t expected_bytes[] = {ACTIVATE_REQUEST};
+   uint32_t expected_len = sizeof (expected_bytes);
+
+   req = kms_request_activate_new (NULL, ACTIVATE_UNIQUE_IDENTIFIER);
```
eramongodb, line 208:
```suggestion
   static const char *const ACTIVATE_UNIQUE_IDENTIFIER =
      "7FJYvnV6XkaUCWuY96bCSc6AuhvkPpqI";

   req = kms_request_activate_new (NULL, ACTIVATE_UNIQUE_IDENTIFIER);
```
I recommend using core language over preprocessor when able.

kms-message/src/kms_kmip_request.c
```diff
@@ -0,0 +1,233 @@
...
+   kmip_writer_t *writer;
+   kms_request_t *req;
+
+   req = calloc (1, sizeof (kms_request_t));
+   req->provider = KMS_REQUEST_PROVIDER_KMIP;
+
+   if (len != 96) {
+      KMS_ERROR (req, "expected SecretData length of 96, got %" PRIu32, len);
+      return req;
+   }
```
eramongodb, line 80:
```suggestion
   if (len != KMS_KMIP_REQUEST_SECRETDATA_LENGTH) {
      KMS_ERROR (req,
                 "expected SecretData length of %" PRIu32 ", got %" PRIu32,
                 (uint32_t) KMS_KMIP_REQUEST_SECRETDATA_LENGTH,
                 len);
      return req;
   }
```
Given this value is constant, I suggest defining this constant and using it to replace/implement `SECRET_DATA_INVALID_LENGTH` in unit tests.

eramongodb (commented):
The volume of calls requiring addition of `check_and_prohibit_kmip()` suggests that, rather than integrating support for KMIP to the existing interface for `kms_request_t`, it may be worth considering a separate KMIP-exclusive interface (`kms_kmip_*()` functions) operating on a `kms_kmip_request_t` (adjust name as appropriate). The new `kmip` data member in `kms_request_t` may be ill-placed.
```c
struct _kms_kmip_request_t {
   kms_request_t *request;
   uint8_t *data;
   uint32_t len;
};
```
Please reject this suggestion if it is not applicable/feasible given KMIP support requirements or integration requirements with the existing KMS interface.

kms-message/src/kms_kmip_request.c
```diff
@@ -0,0 +1,233 @@
...
+                                          uint32_t len)
+{
+   /*
+   Create a KMIP Register request with a 96 byte SecretData of this form:
+
+   <RequestMessage tag="0x420078" type="Structure">
+    <RequestHeader tag="0x420077" type="Structure">
+     <ProtocolVersion tag="0x420069" type="Structure">
+      <ProtocolVersionMajor tag="0x42006a" type="Integer" value="1"/>
+      <ProtocolVersionMinor tag="0x42006b" type="Integer" value="4"/>
```
markbenvenuto, line 47: We should target the earliest version we can. In my POC, I targeted 1.4 since 2.0 has breaking changes. But some KMIP servers only support earlier versions. I think 1.0 will work.

kms-message/src/kms_kmip_request.c
markbenvenuto: I think `Password` might be slightly better. `Seed` is for seeding random number generates.

Extenions are something that is per-KMIP server and not something a client can arbitrarily make up

kms-message/src/kms_message_private.h
markbenvenuto: Agreed, I think a union is a good idea.

eramongodb: +1

eramongodb (approved):
Thank you for providing context on the design. In the interest of being consistent with existing selective interfaces for KMS providers as well as impact on existing consumers, I think the design as currently proposed is acceptable. I am also in favor of the proposed followup work to reorganize data members of `_kms_request_t` into union data members corresponding to each supported KMS provider type.

kms-message/src/kms_kmip_request.c
kevinAlbs: Done.

kms-message/src/kms_message_private.h
kevinAlbs: Filed MONGOCRYPT-342

kms-message/test/test_kms_kmip_request.c
kevinAlbs: Done, that seems sensible to me. I assume the reason is for additional type safety.

PR #206, opened 2021-10-13
MONGOCRYPT-333 support "kmip" provider in libmongocrypt

kevinAlbs:
The last PR for KMIP support in libmongocrypt.

# Summary of Changes

- Support "kmip" KMS provider for each libmongocrypt context (createDatakey, encrypt, and decrypt):
- Add `mongocrypt_kms_ctx_get_kms_provider`.
    - Rationale: It is necessary for drivers to identify the KMS provider associated with a `mongocrypt_kms_ctx_t`. This allows drivers to set custom TLS options on the TLS connection for KMIP.
- Support "kmip" options in `csfle` test utility.

# Testing
- I manually used the `csfle` test utility to test against a PyKMIP server and Hashicorp Vault server. This is not run in Evergreen, but I think it would be valuable. I filed MONGOCRYPT-350.

### DIFF
--- a/src/mongocrypt-kms-ctx.c
+++ b/src/mongocrypt-kms-ctx.c
@@ -795,6 +795,7 @@ _ctx_done_kmip_register (mongocrypt_kms_ctx_t *kms_ctx)
 
    if (!_mongocrypt_buffer_steal_from_string (&kms_ctx->result, uid)) {
       CLIENT_ERR ("Error storing KMS UniqueIdentifer result");
+      bson_free (uid);
       goto done;
    }
    ret = true;
@@ -836,6 +837,7 @@ _ctx_done_kmip_get (mongocrypt_kms_ctx_t *kms_ctx)
    if (!_mongocrypt_buffer_steal_from_data_and_size (
           &kms_ctx->result, secretdata, secretdata_len)) {
       CLIENT_ERR ("Error storing KMS SecretData result");
+      bson_free (secretdata);
       goto done;
    }
 
--- a/src/mongocrypt-buffer-private.h
+++ b/src/mongocrypt-buffer-private.h
@@ -182,7 +182,7 @@ _mongocrypt_buffer_copy_from_data_and_size (_mongocrypt_buffer_t *buf,
 /* _mongocrypt_buffer_steal_from_data_and_size initializes @buf from @data and
  * @len and takes ownership of @data.
  * - Returns false on error.
- * - @buf takes ownership of @data on error.
+ * - @buf does not take ownership of @str on error.
  * - Caller must call _mongocrypt_buffer_cleanup. */
 bool
 _mongocrypt_buffer_steal_from_data_and_size (_mongocrypt_buffer_t *buf,
@@ -195,7 +195,7 @@ _mongocrypt_buffer_steal_from_data_and_size (_mongocrypt_buffer_t *buf,
  * @buf retains a pointer to @str.
  * @str must be NULL terminated.
  * - Returns false on error.
- * - @buf takes ownership of @str on error.
+ * - @buf does not take ownership of @str on error.
  * - Caller must call _mongocrypt_buffer_cleanup. */
 bool
 _mongocrypt_buffer_steal_from_string (_mongocrypt_buffer_t *buf,
--- a/src/mongocrypt-buffer.c
+++ b/src/mongocrypt-buffer.c
@@ -513,11 +513,11 @@ _mongocrypt_buffer_steal_from_data_and_size (_mongocrypt_buffer_t *buf,
                                              size_t len)
 {
    _mongocrypt_buffer_init (buf);
-   buf->data = data;
-   buf->owned = true;
    if (!size_to_uint32 (len, &buf->len)) {
       return false;
    }
+   buf->data = data;
+   buf->owned = true;
    return true;
 }
 
@@ -525,10 +525,10 @@ bool
 _mongocrypt_buffer_steal_from_string (_mongocrypt_buffer_t *buf, char *str)
 {
    _mongocrypt_buffer_init (buf);
-   buf->data = (uint8_t *) str;
-   buf->owned = true;
    if (!size_to_uint32 (strlen (str), &buf->len)) {
       return false;
    }
+   buf->data = (uint8_t *) str;
+   buf->owned = true;
    return true;
 }


test/util/util.c
```diff
@@ -497,7 +497,7 @@ connect_stream_with_tls (mongoc_ssl_opt_t *ssl_opt,
                          int32_t connecttimeoutms,
                          bson_error_t *error)
 {
-   mongoc_stream_t *stream;
+   mongoc_stream_t *stream = NULL;
```
kevinAlbs, line 500: Resolved a clang-tidy warning locally. If `goto done` had run before `stream` was initialized, `mongoc_stream_destroy (stream);` would be given an uninitialized `stream`.

src/mongocrypt-ctx-datakey.c
```diff
@@ -45,6 +47,135 @@ _next_kms_ctx (mongocrypt_ctx_t *ctx)
...
+      CLIENT_ERR ("KMS provider is not KMIP");
+      _mongocrypt_ctx_fail (ctx);
+      goto done;
+   }
+
+   user_supplied_keyid = ctx->opts.kek.provider.kmip.key_id;
+
+   if (ctx->opts.kek.provider.kmip.endpoint) {
+      endpoint = ctx->opts.kek.provider.kmip.endpoint;
+   } else if (ctx->crypt->opts.kms_provider_kmip.endpoint) {
```
chardan, line 72: Ok, two different sources to get the endpoint config.

src/mongocrypt-ctx-datakey.c
```diff
@@ -45,6 +47,135 @@ _next_kms_ctx (mongocrypt_ctx_t *ctx)
...
+   char *user_supplied_keyid;
+   _mongocrypt_endpoint_t *endpoint;
+   mongocrypt_status_t *status = ctx->status;
+   _mongocrypt_buffer_t secretdata = {0};
+
+   dkctx = (_mongocrypt_ctx_datakey_t *) ctx;
+
+   if (ctx->opts.kek.kms_provider != MONGOCRYPT_KMS_PROVIDER_KMIP) {
+      CLIENT_ERR ("KMS provider is not KMIP");
+      _mongocrypt_ctx_fail (ctx);
```
chardan, line 64: Much as variable "ret" is set to false to indicate (wisely, I think) failure by default, should this also be indicated? Or perhaps always done in the goto target, if it's meant to set up failure any time the whole function fails. Rather than trying to set it each time; alternatively, maybe a fail: label that always correctly sets up failure is a good idea, or incoporating this behavior into label done: is a good idea. It looks to me like /you/ got it right in this function, hopefully the next person will be as attentive.

src/mongocrypt-kms-ctx.c
```diff
@@ -110,7 +111,13 @@ _init_common (mongocrypt_kms_ctx_t *kms,
               _mongocrypt_log_t *log,
               _kms_request_type_t kms_type)
 {
-   kms->parser = kms_response_parser_new ();
+   if (kms_type == MONGOCRYPT_KMS_KMIP_REGISTER ||
```
chardan, line 114: Seeing that this gets repeated a couple times, it might be worth considering making it an is_kms() predicate. OTOH, it it's also on the edge of just being noise to do so-- up to you.

src/mongocrypt-util-private.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef MONGOCRYPT_UTIL_PRIVATE_H
+#define MONGOCRYPT_UTIL_PRIVATE_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/* A utility for safely casting from size_t to uint32_t.
+ * Returns false if @in exceeds the maximum value of a uint32_t. */
+bool size_to_uint32 (size_t in, uint32_t *out);
```
eramongodb, line 26:
```suggestion
bool
size_to_uint32 (size_t in, uint32_t *out);
```
ClangFormat.

eramongodb (changes requested):

test/data/key-document-kmip.json
```diff
@@ -0,0 +1,32 @@
...
+        }
+    },
+    "status": {
+        "$numberInt": "0"
+    },
+    "masterKey": {
+        "provider": "kmip",
+        "keyId": "ywxrSj5TLjswd1G4oGFJ6hwWgtTsQip0"
+    },
+    "keyAltNames": [ "keyDocumentKMIP" ]
```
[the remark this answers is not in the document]
kevinAlbs: Oops, I meant to commit the suggestion.

src/mongocrypt-kms-ctx.c
```diff
@@ -1378,4 +1479,146 @@ _mongocrypt_kms_ctx_init_gcp_decrypt (mongocrypt_kms_ctx_t *kms,
...
+
+   ret = true;
+fail:
+   return ret;
+}
+
+const char *
+mongocrypt_kms_ctx_get_kms_provider (mongocrypt_kms_ctx_t *kms, uint32_t *len)
+{
+#define SET_AND_RET(x)                              \
```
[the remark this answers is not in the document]
chardan: +1 (no need to adopt my perhaps-odd usage of the comma operator if it gives you the creeps (I like it because it "says" that the two operations MUST go together, but YYMV)). Thinking about it more, I'm also pretty confident that the optimizer will probably also handle this well if it's a helper function (it would be good to measure, but if this doesn't need every tic, I think we're ok either way).

src/mongocrypt-kms-ctx.c
```diff
@@ -757,6 +770,82 @@ _ctx_done_gcp (mongocrypt_kms_ctx_t *kms, const char *json_field)
...
+      CLIENT_ERR (
+         "Error getting UniqueIdentifer from KMIP Register response: %s",
+         kms_response_get_error (res));
+      goto done;
+   }
+
+   if (!_mongocrypt_buffer_steal_from_string (&kms_ctx->result, uid)) {
+      CLIENT_ERR ("Error storing KMS UniqueIdentifer result");
+      goto done;
+   }
```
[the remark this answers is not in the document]
eramongodb: Confirming, `ksm_ctx->result` holding a partially-formed value on error (`data` and `owned` updated, but not `len`?) was surprising behavior. I am satisfied with the proposed updates.

src/mongocrypt-ctx-private.h
```diff
@@ -137,6 +137,10 @@ typedef struct {
    _mongocrypt_buffer_t key_doc;
    _mongocrypt_buffer_t plaintext_key_material;
    _mongocrypt_buffer_t encrypted_key_material;
+
+   char *kmip_unique_identifier;
```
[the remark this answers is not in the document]
eramongodb: I find it much preferable to cast on free rather than permitting users unnecessary permissions to modify the target string. `free` (and equivalents) often not taking pointer-to-const is indeed unfortunate for reasons outlined in the linked SO post.

eramongodb (approved):

PR #209, opened 2021-11-03
MONGOCRYPT-355 remove SLES 12 s390x variant

kevinAlbs:
**Background & Motivation**

SLES 12 s390x is no longer a required platform to build.

Removing is required for the 1.3.0 release since the `publish-packages` task depends on the `build-and-test-and-upload` task on the `suse12-s390x` variant to succeed. The task is failing due to the issues noted in BUILD-14027.


rcsanchez97 (approved):
LGTM

PR #213, opened 2021-11-05
MONGOCRYPT-359 simplify HMAC creation, fix compatibility with OpenSSL…

remicollet:
… 3.0

https://jira.mongodb.org/browse/MONGOCRYPT-359

### DIFF
--- a/src/crypto/libcrypto.c
+++ b/src/crypto/libcrypto.c
@@ -190,13 +190,23 @@ _native_crypto_hmac_sha_512 (const _mongocrypt_buffer_t *key,
                              mongocrypt_status_t *status)
 {
    const EVP_MD *algo;
+
+   algo = EVP_sha512 ();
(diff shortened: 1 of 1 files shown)


remicollet:
Notice, simplified code enables only for OpenSSL 1.1.0, as simpler to test for me on Fedora, BTW, seems to be usable starting with 1.0.2 (on RHEL-7), no idea about LibreSSL.

chardan (approved):
@remicollet First, thank you for this patch! It looks good to me, and I'm approving it-- much appreciated! As a note, I think it's a small enough delta that the macro is more clear than breaking this into more functions, but if we need to add another version sometime that wouldn't be a bad idea. Thanks again for your contribution!

kevinAlbs (approved):
Thank you for the fix @remicollet!

PR #214, opened 2021-11-05
PYTHON-3006 Support testing with PyMongo 4 and >=3.11

ShaneHarvey:
Test suite now passes with pymongo 3.11, 3.12, and 4.0.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -57,7 +57,7 @@ setup(
     keywords=["mongo", "mongodb", "pymongocrypt", "pymongo", "mongocrypt",
               "bson"],
     test_suite="test",
-    tests_require=["pymongo"],
+    tests_require=["pymongo>=3.11"],
     license="Apache License, Version 2.0",
     python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*",
     classifiers=[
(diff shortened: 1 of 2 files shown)


bindings/python/setup.py
```diff
@@ -57,7 +57,7 @@ def get_tag(self):
     keywords=["mongo", "mongodb", "pymongocrypt", "pymongo", "mongocrypt",
               "bson"],
     test_suite="test",
-    tests_require=["pymongo"],
+    tests_require=["pymongo>=3.11"],
```
ShaneHarvey, line 60: UNSPECIFIED was introduced in 3.11 (see PYTHON-2152).

bindings/python/test/test_mongocrypt.py
```diff
@@ -265,8 +266,8 @@ def test_encrypt(self):
             self.assertEqual(ctx.state, lib.MONGOCRYPT_CTX_READY)
 
             encrypted = ctx.finish()
-            self.assertEqual(
-                BSON(encrypted).decode(), json_data('encrypted-command.json'))
+            self.assertEqual(bson.decode(encrypted, OPTS),
```
ShaneHarvey, line 269: bson.decode and bson.encode are preferred over using the BSON class (see PYTHON-1785).

behackett (approved):

juliusgeo (approved):

PR #221, opened 2021-11-23
feat(NODE-3777): add node bindings for kms provider

durran:
Adds Node bindings for KMIP provider support. Bindings now wrap `mongocrypt_kms_ctx_get_kms_provider` and add the typescript definition for the new KMIP provider with a single `endpoint` url.


jyemin (approved):
LGTM after adding a test for the new method that returns the name of the current KMS provider.

PR #222, opened 2021-11-24
feat(NODE-3781): add tls options for kms providers

durran:
Adds typescript definitions for allowed TLS options for the KMS provider.

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -147,10 +147,6 @@ export interface ClientEncryptionTLSOptions {
    */
   tls?: boolean;
   /**
-   * Specifies the location of a local TLS Certificate
-   */
-  tlsCertificateFile?: string;
-  /**
    * Specifies the location of a local .pem file that contains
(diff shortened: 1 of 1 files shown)


bindings/node/index.d.ts
```diff
@@ -131,6 +131,45 @@ export interface KMSProviders {
...
+ */
+export interface ClientEncryptionTLSOptions {
+  /**
+   * Enables or disables TLS/SSL for the connection.
+   */
+  tls?: boolean;
+  /**
+   * Specifies the location of a local TLS Certificate
+   */
+  tlsCertificateFile?: string;
```
addaleax, line 152: In [NODE-3591](https://jira.mongodb.org/browse/NODE-3591) we talked about deprecating this – my understanding is that this is only being added for a new feature (KMIP support), so maybe it’s a good idea to omit this?

durran: Thanks for the suggestion - I've removed.

nbbeeken (approved):
LGTM

PR #224, opened 2021-12-03
Move CMAKE_C_FLAGS from CMakeLists.txt into Evergreen build scripts

eramongodb:
Hard-coding compiler-specific flags in the CMakeLists.txt file, especially by overwriting the `CMAKE_C_FLAGS` variable (whose purpose should be for users to communicate _their_ compiler flag preferences, potentially at a global scale across more than one CMake project), makes it difficult for users to use and integrate a given CMake project (e.g. hard-coding `-Werror` prevents users from enabling additional warnings that they may find useful for development). This PR proposes moving the currently hard-coded `CMAKE_C_FLAGS` flags into corresponding Evergreen build scripts instead.


eramongodb:
Note: as currently proposed by this PR, _all_ currently hard-coded C flags are being moved into build scripts:

* Windows: `/W3 /wd4996 /D_CRT_SECURE_NO_WARNINGS /WX`
* Linux: `-Wall -Werror -Wno-missing-braces`

However, the only major user-obstructive flags are `-Werror` and `/WX` ([treat warnings as errors](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#treat-warnings-as-errors)). If we want to keep the other warning flags enabled regardless of user preferences (e.g. `/D_CRT_SECURE_NO_WARNINGS`), I suggest that _at least_ the `-Werror` flag be moved as proposed.

.evergreen/build_all.sh
```diff
@@ -36,7 +36,17 @@ command -v $CMAKE || (echo "CMake could not be found...aborting!"; exit 1)
 cd libmongocrypt
 mkdir cmake-build
 cd cmake-build
-$CMAKE -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ../
+
+if [ "$OS" = "Windows_NT" ]; then
+    # W4996 - POSIX name for this item is deprecated
+    # TODO: add support for clang-cl which is detected as MSVC
```
markbenvenuto, line 42: clang-cl will work as-is The only caveat is that the warnings enable/disable is pretty hacky in clang-cl. A lot of the warnings switches in cl.exe are no-ops in clang-cl.

markbenvenuto (approved):

.evergreen/build_all.sh
eramongodb: I am just preserving a comment that was [already present](https://github.com/mongodb/libmongocrypt/blob/ddd2f105fe34ce7682fbec76c4a443f26f4ed076/CMakeLists.txt#L105). AFAIK these changes do nothing to address it. If it is no longer applicable or relevant, I can remove it as part of these changes.

kevinAlbs (approved):

rcsanchez97 (approved):
LGTM

durran (approved):

PR #227, opened 2021-12-20
chore(NODE-3633): update Socks5 support for driver PR feedback

addaleax:
- Update the Socks5 options naming to use the connection string
  parameter names as keys, rather than shortened variants,
  for easier integration into the Node.js driver
- Fix and test username/password authentication (noticed while
  addressing the above, this is absolutely my bad)
- [related driver feedback](https://github.com/mongodb/node-mongodb-native/pull/3041#discussion_r772337188)


nbbeeken (approved):
LGTM thanks for the quick refactor!

durran (approved):

PR #235, opened 2022-01-17
feat(NODE-3777): set tls options per kms provider

durran:
- Changes the TLS provider options to function per KMS provider. This now ensures each KMS provider can customise their TLS settings.
- Adds validation on the TLS options not to support any insecure options.

Note Prose and spec tests are located in the main Node repo. Related PR https://github.com/mongodb/node-mongodb-native/pull/3070

Node CSFLE run against this specific branch: https://spruce.mongodb.com/version/61e753873066153369a025de/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -203,7 +203,13 @@ export interface ClientEncryptionOptions {
   /**
    * TLS options for kms providers to use.
    */
-  tlsOptions?: ClientEncryptionTLSOptions;
+  tlsOptions?: {
+    aws?: ClientEncryptionTLSOptions;
+    local?: ClientEncryptionTLSOptions;
+    azure?: ClientEncryptionTLSOptions;
+    gcp?: ClientEncryptionTLSOptions;
+    kmip?: ClientEncryptionTLSOptions;
+  }
 }
 
 /**
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -151,11 +151,7 @@ export interface KMSProviders {
  *  - tlsDisableOCSPEndpointCheck
  *  - tlsDisableCertificateRevocationCheck
  */
-export interface ClientEncryptionTLSOptions {
-  /**
-   * Enables or disables TLS/SSL for the connection.
-   */
-  tls?: boolean;
+export interface ClientEncryptionTlsOptions {
   /**
    * Specifies the location of a local .pem file that contains
    * either the client's TLS/SSL certificate and key or only the
@@ -203,13 +199,7 @@ export interface ClientEncryptionOptions {
   /**
    * TLS options for kms providers to use.
    */
-  tlsOptions?: {
-    aws?: ClientEncryptionTLSOptions;
-    local?: ClientEncryptionTLSOptions;
-    azure?: ClientEncryptionTLSOptions;
-    gcp?: ClientEncryptionTLSOptions;
-    kmip?: ClientEncryptionTLSOptions;
-  }
+  tlsOptions?: { [kms in keyof KMSProviders]?: ClientEncryptionTLSOptions };
 }
 
 /**
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -38,6 +38,7 @@ module.exports = function(modules) {
      * @param {MongoClient} client The client used for encryption
      * @param {object} options Additional settings
      * @param {string} options.keyVaultNamespace The namespace of the key vault, used to store encryption keys
+     * @param {object} options.tlsOptions An object that maps KMS provider names to TLS options.
      * @param {MongoClient} [options.keyVaultClient] A `MongoClient` used to fetch keys from a key vault. Defaults to `client`
      * @param {KMSProviders} [options.kmsProviders] options for specific KMS providers to use
      *
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -66,6 +66,7 @@ module.exports = function(modules) {
       this._client = client;
       this._bson = options.bson || client.topology.bson;
       this._proxyOptions = options.proxyOptions;
+      this._tlsOptions = options.tlsOptions;
 
       if (options.keyVaultNamespace == null) {
         throw new TypeError('Missing required option `keyVaultNamespace`');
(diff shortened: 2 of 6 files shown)


bindings/node/lib/stateMachine.js
```diff
@@ -305,6 +307,28 @@ module.exports = function(modules) {
...
+      const tlsOptions = this.options.tlsOptions;
+      if (tlsOptions) {
+        const providerTlsOptions = tlsOptions[kmsProvider];
+        if (providerTlsOptions) {
+          if (providerTlsOptions.tlsCertificateKeyFile) {
+            options.cert = await readFile(providerTlsOptions.tlsCertificateKeyFile, { encoding: 'ascii' });
+          }
+          if (providerTlsOptions.tlsCAFile) {
+            options.ca = await readFile(providerTlsOptions.tlsCAFile, { encoding: 'ascii' });
+          }
```
addaleax, line 324: totally optional, but since these are fully independent:

```suggestion
          await Promise.all([
            providerTlsOptions.tlsCertificateKeyFile && (async() => {
             options.cert = await readFile(providerTlsOptions.tlsCertificateKeyFile, { encoding: 'ascii' });
            })(),
            providerTlsOptions.tlsCAFile && (async() => {
              options.ca = await readFile(providerTlsOptions.tlsCAFile, { encoding: 'ascii' });
            })()
          ]);
```

also, quoting the Node.js Buffer docs on the `ascii` encoding:

> 'ascii': For 7-bit ASCII data only. When encoding a string into a Buffer, this is equivalent to using 'latin1'. When decoding a Buffer into a string, using this encoding will additionally unset the highest bit of each byte before decoding as 'latin1'. Generally, there should be no reason to use this encoding, as 'utf8' (or, if the data is known to always be ASCII-only, 'latin1') will be a better choice when encoding or decoding ASCII-only text. It is only provided for legacy compatibility.

I'd remove the encoding option entirely and just read it as a Buffer if that's possible, the TLS API accepts it that way anyway and it saves a string decoding/encoding roundtrip.

(i.e. the suggestion above would become:


```suggestion
          await Promise.all([
            providerTlsOptions.tlsCertificateKeyFile && (async() => {
             options.cert = await readFile(providerTlsOptions.tlsCertificateKeyFile);
            })(),
            providerTlsOptions.tlsCAFile && (async() => {
              options.ca = await readFile(providerTlsOptions.tlsCAFile);
            })()
          ]);
```

)

durran: I actually changed since our conversation to use `readFileSync` for the file loading but have also subsequently removed the ascii encoding and returned buffers. Do you still want the file reads async?

bindings/node/index.d.ts
```diff
@@ -203,7 +203,13 @@ export interface ClientEncryptionOptions {
   /**
    * TLS options for kms providers to use.
    */
-  tlsOptions?: ClientEncryptionTLSOptions;
+  tlsOptions?: {
```
durran, line 206: TLS options to KMS providers are configured on a per KMS provider basis for future flexibility.

bindings/node/lib/stateMachine.js
```diff
@@ -38,6 +40,14 @@ module.exports = function(modules) {
     [MONGOCRYPT_CTX_DONE, 'MONGOCRYPT_CTX_DONE']
   ]);
 
+  const INSECURE_TLS_OPTIONS = [
```
durran, line 43: Spec provides this as the list of insecure TLS options.

bindings/node/lib/stateMachine.js
```diff
@@ -305,6 +325,44 @@ module.exports = function(modules) {
...
+     * @param {string} kmsProvider The KMS provider name.
+     * @param {ClientEncryptionTLSOptions} tlsOptions The client TLS options for the provider.
+     *
+     * @returns {Error} If any option is invalid.
+     */
+    validateTlsOptions(kmsProvider, tlsOptions) {
+      const tlsOptionNames = Object.keys(tlsOptions);
+      for (const option of INSECURE_TLS_OPTIONS) {
+        if (tlsOptionNames.includes(option)) {
+          return new MongoCryptError(`Insecure TLS options prohibited for ${kmsProvider}: ${option}`);
```
durran, line 341: Spec requires the error message contain the text "Insecure TLS options prohibited".

bindings/node/lib/stateMachine.js
```diff
@@ -305,6 +325,44 @@ module.exports = function(modules) {
...
+    }
+
+    /**
+     * @ignore
+     * Sets only the valid secure TLS options.
+     *
+     * @param {ClientEncryptionTLSOptions} tlsOptions The client TLS options for the provider.
+     * @param {Object} options The existing connection options.
+     */
+    setTlsOptions(tlsOptions, options) {
```
durran, line 353: Only these 3 secure TLS options are allowed to be mapped to the TLS connection.

bindings/node/lib/stateMachine.js
addaleax: > Do you still want the file reads async?

I mean, unless there is a reason for them *not* to be asynchronous? Synchronous I/O is definitely an antipattern in production code

bindings/node/index.d.ts
```diff
@@ -203,7 +203,13 @@ export interface ClientEncryptionOptions {
...
+    aws?: ClientEncryptionTLSOptions;
+    local?: ClientEncryptionTLSOptions;
+    azure?: ClientEncryptionTLSOptions;
+    gcp?: ClientEncryptionTLSOptions;
+    kmip?: ClientEncryptionTLSOptions;
+  }
```
addaleax, line 212: If the expectation is that this list is always supposed to match the `KMSProviders` keys, and that we likely won’t need special options for individual ones, this could also be:

```suggestion
  tlsOptions?: { [kms in keyof KMSProviders]?: ClientEncryptionTLSOptions };
```

(or maybe

```suggestion
  tlsOptions?: { [kms in Exclude<keyof KMSProviders, 'local'>]?: ClientEncryptionTLSOptions };
```

I’m not sure how much sense it makes to specify TLS options for `local` … :slightly_smiling_face: )

addaleax (approved):

bindings/node/index.d.ts
durran: The spec doesn't explicitly exclude local so I took the first suggestion.

bindings/node/lib/clientEncryption.js
```diff
@@ -66,6 +66,7 @@ module.exports = function(modules) {
       this._client = client;
```
baileympearson, line 66: I can't comment above because of the way Github structures diffs but should we add the option, `tlsOptions`, to the doc string above?  Looks like the other options are included in the doc string.

baileympearson (approved):
I left one minor (non-blcoking) comment.  Otherwise the changes look good to me

bindings/node/lib/clientEncryption.js
durran: Added.

PR #237, opened 2022-01-25
Revert #224 and move only warnings-as-error flags into scripts

eramongodb:
This PR reverts most of #224 for the following reasons:

- PR was merged without including related updates to the kms-message project, which has its own `CMAKE_C_FLAGS` configuration.
- Given old application of warnings-as-error enforces that related warnings are not present, keeping the warning flags enabled is preferable so new instances are immediately detectable (during local development, not delayed until Evergreen CI build report).
- The primary blocker to integration as described in #224 is the warnings-as-error flag (`-Werror`, `/WX`). However, the merged PR was too greedy/drastic in moving _all_ flags, including those that are specifically and deliberately specified such as `/D_CRT_SECURE_NO_WARNINGS`, into Evergreen build scripts.

This PR instead opts for [the more conservative approach](https://github.com/mongodb/libmongocrypt/pull/224#issuecomment-985712676) as previously described/suggested, with the addition of corresponding updates to the kms-message project.


kevinAlbs (approved):

rcsanchez97 (approved):
LGTM

PR #234, opened 2022-01-12
MONGOCRYPT-340 add macOS m1 variant

kevinAlbs:
# Summary
- Add variant named `macos_m1` running on distro `macos-1100-arm64` to run the task `build-and-test-and-upload`.
- Use `/opt/homebrew/bin/cmake` instead of `/Applications/Cmake.app/Contents/bin/cmake`. The former has `arm64` architecture and produces `arm64` builds by default. The latter has `x86_64` architecture and builds with `x86_64` architecture by default. See BUILD-14565.

# Caveats

Bindings tasks, like `build-and-test-node`, `build-and-test-java`, `build-and-test-csharp`, and `test-python` will require further changes. [This Evergreen patch build](https://spruce.mongodb.com/version/61df38b73e8e8655c0380c22) is configured with all tasks and should provide error logs.

The linked patch build includes a failing `build-and-test-node` on Windows 2016. I think it is unrelated to these changes.

PR #238, opened 2022-01-27
MONGOCRYPT-372 make macos M1 builds universal builds

kevinAlbs:
# Summary
- Update `macos_m1` builds to build with both x86_64 and arm64 architecture with `-DCMAKE_OSX_ARCHITECTURES='arm64;x86_64'`
- Rename the `macos_m1` variant to `macos_universal`.

# Notes
The first attempt at using `-DCMAKE_OSX_ARCHITECTURES='arm64;x86_64'` failed on the [macos-1014 distro](https://evergreen.mongodb.com/lobster/evergreen/task/libmongocrypt_macos_build_and_test_universal_mac_patch_03720bce6905a0bc4c3e8fff7bfa5c92f3eb6c14_61f1c49257e85a6a75727184_22_01_26_22_00_50/0/task#bookmarks=0%2C167&l=). That motivated using the `macos-1100-arm64` distro.

### DIFF
--- a/bindings/java/mongocrypt/build.gradle.kts
+++ b/bindings/java/mongocrypt/build.gradle.kts
@@ -135,7 +135,7 @@ val jnaMappingList: List<LibMongoCryptS3Data> = listOf(
         LibMongoCryptS3Data("rhel-67-s390x", "linux-s390x"),
         LibMongoCryptS3Data("ubuntu1604-arm64", "linux-aarch64"),
         LibMongoCryptS3Data("windows-test", "win32-x86-64"),
-        LibMongoCryptS3Data("macos", "darwin")
+        LibMongoCryptS3Data("macos_x86_64", "darwin")
 )
 
 jnaMappingList.forEach {
--- a/.evergreen/build_all.sh
+++ b/.evergreen/build_all.sh
@@ -61,6 +61,18 @@ $CMAKE --build . --target test-mongocrypt --config RelWithDebInfo
 $CMAKE --build ./kms-message --target test_kms_request --config RelWithDebInfo
 cd $evergreen_root
 
+# MONGOCRYPT-372, ensure macOS universal builds contain both x86_64 and arm64 architectures.
+if [ "$MACOS_UNIVERSAL" = "ON" ]; then
+    echo "Checking if libmongocrypt.dylib contains both x86_64 and arm64 architectures..."
+    ARCHS=$(lipo -archs $MONGOCRYPT_INSTALL_PREFIX/lib/libmongocrypt.dylib)
+    if [[ "$ARCHS" == *"x86_64"* && "$ARCHS" == *"arm64"* ]]; then
+        echo "Checking if libmongocrypt.dylib contains both x86_64 and arm64 architectures... OK"
+    else
+        echo "Checking if libmongocrypt.dylib contains both x86_64 and arm64 architectures... ERROR. Got: $ARCHS"
+        exit
+    fi
+fi
+
 if [ "$PPA_BUILD_ONLY" ]; then
     echo "Only building/installing for PPA";
     exit 0;
--- a/.evergreen/build_install_bson.sh
+++ b/.evergreen/build_install_bson.sh
@@ -27,6 +27,10 @@ else
     fi
 fi
 
+if [ "$MACOS_UNIVERSAL" = "ON" ]; then
+    ADDITIONAL_CMAKE_FLAGS="$ADDITIONAL_CMAKE_FLAGS -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64'"
+fi
+
 $CMAKE --version
 
 # Remove remnants of any earlier build
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -472,7 +472,7 @@ tasks:
       name: build-and-test-and-upload
     - variant: ubuntu2004-arm64
       name: build-and-test-and-upload
-    - variant: macos_universal
+    - variant: macos
       name: build-and-test-and-upload
   commands:
     - command: shell.exec
@@ -521,7 +521,7 @@ tasks:
     - func: "download tarball"
       vars: { variant_name: "ubuntu2004-arm64" }
     - func: "download tarball"
-      vars: { variant_name: "macos_universal" }
+      vars: { variant_name: "macos" }
     - command: archive.targz_pack
       params:
         target: libmongocrypt-all.tar.gz
@@ -1065,7 +1065,7 @@ buildvariants:
   run_on: ubuntu2004-small
   tasks:
   - name: debian-package-build
-- name: macos_universal
+- name: macos
   display_name: macOS m1 (Apple LLVM)
   run_on: macos-1100-arm64
   expansions:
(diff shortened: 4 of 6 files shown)


JamesKovacs (commented):
The changes look straightforward. I'm wondering why we are producing `macos` and `macos_universal` binaries. Why do we still need `macos` if `macos_universal` contains both x64 and arm64 code?

rcsanchez97 (approved):
LGTM!

rcsanchez97:
@JamesKovacs I saw in a discussion (it could have been in Slack, but I'm a bit fuzzy on that), that we should prefer to produce single-arch binaries for the specific target arch.  However, there are likely cases where users would want to produce universal binaries.  At least, that's my understanding.

blink1073:
My understanding is that for users not on M1, they get the benefit of a smaller download if you provide `macos` binaries.

JamesKovacs:
The size of a universal binary is going to depend considerably on the split between executable code versus other resources like embedded data/files. I suspect that `libmongocrypt` is going to be mostly executable code. Thus the size of the universal binary is going to be approximately `sizeof(x64) + sizeof(arm64)`. `sizeof(x64)` is about 221KB when I compiled it. I'd expect `sizeof(arm64)` to be similar. So the question is whether ~200KB savings is worth having separate `x64` and `universal` builds - including all the support headaches when someone unwittingly downloads the `x64` binary onto their M1 laptop. I would argue that providing a single universal binary that works on all MacOS systems is preferable.

kevinAlbs:
> My understanding is that for users not on M1, they get the benefit of a smaller download if you provide `macos` binaries.

From comparing the [attachment on macos-1014](https://spruce.mongodb.com/task/libmongocrypt_macos_build_and_test_and_upload_patch_03720bce6905a0bc4c3e8fff7bfa5c92f3eb6c14_61f1f9d67742ae451ab3539f_22_01_27_01_48_07/files?execution=0) and the universal attachment on [macos-1100-arm64](https://spruce.mongodb.com/task/libmongocrypt_macos_universal_build_and_test_and_upload_patch_03720bce6905a0bc4c3e8fff7bfa5c92f3eb6c14_61f1f9d67742ae451ab3539f_22_01_27_01_48_07/files?execution=0) the sizes of libmongocrypt.dylib are 387K and 983K. IMO not a big enough size difference to concern a user.

> The changes look straightforward. I'm wondering why we are producing `macos` and `macos_universal` binaries. Why do we still need `macos` if `macos_universal` contains both x64 and arm64 code?

The motivation for keeping both build variants with the `macos-1100-arm64` and `macos-1014` distros is to continue building bindings tasks on macOS.

The motivation for adding the universal build to the `macos-1100-arm64` was [this failure](https://evergreen.mongodb.com/lobster/evergreen/task/libmongocrypt_macos_build_and_test_universal_mac_patch_03720bce6905a0bc4c3e8fff7bfa5c92f3eb6c14_61f1c49257e85a6a75727184_22_01_26_22_00_50/0/task#bookmarks=0%2C167&l=) observed on the `macos-1014` distro.

The motivation for not adding all bindings tasks to the `macos-1100-arm64` distro were failures from this [patch build](https://spruce.mongodb.com/version/61df38b73e8e8655c0380c22/tasks) noted in https://github.com/mongodb/libmongocrypt/pull/234.

I think it is sensible to have the universal build be the default. I replaced the default with the universal build.

Before MONGOCRYPT-340, libmongocrypt-all.tar.gz contained these directories:
- macos (x86_64 libraries)

After MONGOCRYPT-340 libmongocrypt-all.tar.gz contained these directories:
- macos_m1 (arm64 libraries)
- macos (x86_64 libraries)

With the proposed changes libmongocrypt-all.tar.gz will contain these directories:
- macos (arm64 and x86_64 universal libraries)
- macos_x86_64 (x86_64 libraries)

blink1073:
"Generally speaking, because Pip 20.3 is required for the universal2 wheel, most packages should provide both x86_64 and universal2 wheels for now. Once Pip 20.3+ is common on macOS, then it should be possible to ship only the universal2 wheel."

https://cibuildwheel.readthedocs.io/en/stable/faq/

GAUTAMRAJU15:
Hey there, I don't see issues page, but i saw this PR about mac m1.
I use Mac m1, and node 14.17.0 LTS. I successfully installed _mongodb-client-encryption_ npm package. Now when i try to run my index.js where i require('mongodb-cliebt-encryption"), it gives me following error in the terminal.

`Error: dlopen(/Users/ray/Desktop/crypt/cryptpad/database/node_modules/mongodb-client-encryption/build/Release/mongocrypt.node, 0x0001): tried: '/Users/ray/Desktop/crypt/cryptpad/database/node_modules/mongodb-client-encryption/build/Release/mongocrypt.node' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/lib/mongocrypt.node' (no such file)`

Anyone knows what might be the error ? I followed the README of this repo, but still same issue on mac M1

kevinAlbs:
@GAUTAMRAJU15 I have replied to MONGOCRYPT-375 to note watching NODE-3631 for further updates.

GAUTAMRAJU15:
@kevinAlbs Won't this PR solve it ?

kevinAlbs:
> @kevinAlbs Won't this PR solve it ?

I believe there are additional changes required to update the mongodb-client-encryption package. This PR updates the build for libmongocrypt, but does not make the associated changes to language bindings.

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -135,7 +135,7 @@ val jnaMappingList: List<LibMongoCryptS3Data> = listOf(
         LibMongoCryptS3Data("rhel-67-s390x", "linux-s390x"),
         LibMongoCryptS3Data("ubuntu1604-arm64", "linux-aarch64"),
         LibMongoCryptS3Data("windows-test", "win32-x86-64"),
-        LibMongoCryptS3Data("macos", "darwin")
```
rozza, line 138: To clarify the Java library will only package the `macos_x86_64` bindings for Mac users.

kevinAlbs: Yes.

The `publish-snapshot` task depends on the `build-and-test-java` task running on the `macos-1014` distro. The `macos-1014` distro only publishes `x86_64` builds.

The description of https://github.com/mongodb/libmongocrypt/pull/234 includes an attempted patch build with running `build-and-test-java` on the `macos-1100-arm64`. The `build-and-test-java` task failed with what looked like [Java environment issue](https://spruce.mongodb.com/task/libmongocrypt_macos_m1_build_and_test_java_patch_f4fcabfca2873a73c841556eed0e4a7370a2dfba_61df38b73e8e8655c0380c22_22_01_12_20_23_19/logs?execution=0).

I decided to keep the current behavior. Have `publish-snapshot` depend on the `build-and-test-and-java` run on the `macos-1014`

rozza: :+1:

rozza (approved):

JamesKovacs (approved):
LGTM

PR #240, opened 2022-01-31
Fix node js security issue

Tofandel:
Npm audit:
```txt
simple-get  <4.0.1
Severity: high
Exposure of Sensitive Information in simple-get - https://github.com/advisories/GHSA-wpg7-2c88-r8xv
fix available via `npm audit fix --force`
Will install mongodb-client-encryption@0.2.0, which is a breaking change
node_modules/mongodb-client-encryption/node_modules/simple-get
  prebuild-install  <=6.1.4
  Depends on vulnerable versions of simple-get
  node_modules/mongodb-client-encryption/node_modules/prebuild-install
    mongodb-client-encryption  >=0.3.0
    Depends on vulnerable versions of prebuild-install
    node_modules/mongodb-client-encryption
```
https://github.com/advisories/GHSA-wpg7-2c88-r8xv

### DIFF
--- a/bindings/node/package.json
+++ b/bindings/node/package.json
@@ -40,7 +40,7 @@
     "bindings": "^1.5.0",
     "bl": "^2.2.1",
     "node-addon-api": "^4.1.0",
-    "prebuild-install": "6.1.2",
+    "prebuild-install": "^7.0.1",
     "socks": "^2.6.1"
   },
   "devDependencies": {


nbbeeken:
@Tofandel Thanks for bringing this to our attention, we have a number of other updates we want to perform that we'll tackle in #241 and the fix here will go out in a 2.0 release. Unfortunately it isn't possible to patch this for 1.x versions since the fixed version of `prebuild-install` has dropped support for node versions we still support in that release. I'll close this, but please feel free to reach out with any further questions.

PR #243, opened 2022-02-07
Utilities for strings and paths

vector-of-bool:
This contains a beginning of a bundle of string and path handling utilities. While not yet used directly by the rest of the library, it will be used as part of DLL handling later on. This is also intended to be useful generically and eventually to be shared in other C codebases.

The API is all `static inline` so as to produce no visible symbols, and guards against inclusion by any files that do not explicitly link to it via CMake. Because no external symbols are produced, identifiers are only qualified enough to prevent direct collision within a single translation unit, enough to identify the components as part of the private API, but not necessarily globally unique.

Also includes an init-once API that will be later relevant.

@rcsanchez97 added as a reviewer to audit the unintended effects this may have on the build process. Hopefully it will not have any effect on installation, though.


src/mlib/path.h
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * @brief Obtain the parent path of the given path.
+ *
+ * @param path A path string
+ * @return mstr_view A substring of the given path string that views only the
+ * parent directory of the given path.
+ */
+static inline mstr_view
+mpath_parent (mstr_view path)
```
addaleax, line 52: iiuc, this means that `mpath_parent()` returns an empty string if the input is `/tmp` or (on win32) `C:\`. Is that expected? At least the path handling libraries of Python and Node.js differ in returning `/` and `C:\`, respectively, and I think that would be the most reasonable behavior here as well.

Does this code need to be able to handle Windows UNC paths here (e.g. `\\?\Ç:\Users\me`)? Since the PR description says that this is a general-purpose library and it will handle user-defined input, I would assume that the answer is ‘yes’ as well.

(Also, speaking from Python and Node.js experience, both platforms also provide Windows-specific and POSIX-specific versions of their path handling functions regardless, and will only switch the default depending on the host OS. That has helped a lot with testing, since it means that developers can work on the path handling code of one platform without necessarily being on that platform.)

src/mlib/str.h
```diff
@@ -0,0 +1,767 @@
...
+ * zero-initialized. The caller can then modify the code units in the array via
+ * the @ref mstr_mut::data member. Once finished modifying, can be converted to
+ * an immutable mstr by copying the @ref mtsr_mut::mstr union member.
+ *
+ * @param len The length of the new string.
+ * @return mstr_mut A new mstr_mut
+ *
+ * @note The @ref mstr_mut::mstr member MUST eventually be given to
+ * @ref mstr_free().
+ */
```
addaleax, line 143: Who is responsible for handling allocation failures?

kevinAlbs: Failure to allocate is not well handled in libmongocrypt (MONGOCRYPT-208).

src/mlib/str.h
```diff
@@ -0,0 +1,767 @@
...
+static inline int
+mstr_rfind (mstr_view given, mstr_view needle)
+{
+   if (needle.len > given.len) {
+      return -1;
+   }
+   const char *scan = given.data + given.len - needle.len;
+   const char *const needle_end = needle.data + needle.len;
+   for (; scan != given.data; --scan) {
```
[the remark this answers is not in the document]
vector-of-bool: Good catch

kevinAlbs (approved):
LGTM

PR #247, opened 2022-02-11
Dynamic library loading

vector-of-bool:
This is a prereq for MONGOCRYPT-373. This introduces private components that can be used to load a dynamic library at runtime. This changes also include tweaks to the mlib path APIs added in #243 

Each commit can be reviewed individually for easier reading.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,7 +336,7 @@ target_include_directories (test-mongocrypt PRIVATE ${BSON_INCLUDES} "${CMAKE_CU
 target_compile_definitions (test-mongocrypt PRIVATE ${BSON_DEFINITIONS})
 
 # Set a definition so that testcases can know where test-mongocrypt.exe was written to
-target_compile_definitions(test-mongocrypt PRIVATE "TEST_MONGOCRYPT_OUTPUT_PATH=\"$<TARGET_FILE:test-mongocrypt>\"")
+target_compile_definitions (test-mongocrypt PRIVATE "TEST_MONGOCRYPT_OUTPUT_PATH=\"$<TARGET_FILE:test-mongocrypt>\"")
 
 add_test (
    NAME mongocrypt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,7 @@ add_subdirectory (bindings/cs)
 include (GenerateExportHeader)
 include (GNUInstallDirs)
 
-enable_testing ()
+include (CTest)
 
 set (MONGOCRYPT_PUBLIC_HEADERS
    src/mongocrypt-compat.h
@@ -287,6 +287,16 @@ set_target_properties (mongocrypt_static PROPERTIES
    OUTPUT_NAME "mongocrypt-static"
 )
 
+if (BUILD_TESTING)
+   # Use C++ in the testing DLL to ensure we can load a library with the C++ runtime
+   enable_language (CXX)
+   add_library (test-dll MODULE test/test-dll.cpp)
+   set_target_properties (test-dll PROPERTIES
+      SUFFIX ".dll"
+      PREFIX ""
+      )
+endif ()
+
 set (TEST_MONGOCRYPT_SOURCES
    test/test-mongocrypt-assert-match-bson.c
    test/test-mongocrypt-buffer.c
@@ -299,6 +309,7 @@ set (TEST_MONGOCRYPT_SOURCES
    test/test-mongocrypt-ctx-encrypt.c
    test/test-mongocrypt-ctx-setopt.c
    test/test-mongocrypt-datakey.c
+   test/test-mongocrypt-dll.c
    test/test-mongocrypt-endpoint.c
    test/test-mongocrypt-kek.c
    test/test-mongocrypt-key.c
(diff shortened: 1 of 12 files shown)


vector-of-bool:
I adopted an `mcr_` prefix for these private components, because typing `_mongocrypt_` was hurting my hands 🙂. Open to bikeshedding.

src/mlib/path.h
```diff
@@ -76,31 +161,219 @@ mpath_parent (mstr_view path)
...
+      }
+   } else {
+      // No root name
+      if (rdir.len) {
+         if (f == MPATH_POSIX) {
+            // No root name, but a root directory on a POSIX path indicates an
+            // absolute path
+            return mstr_copy (path);
+         }
+         mstr abs_base = mpath_absolute (base, f);
```
kevinAlbs, line 353: Note, this can result in a runtime recursive loop if `f == MPATH_WIN32` and is run on a POSIX machine.

I am not sure if that is a realistic concern. My understanding is that MPATH_NATIVE will only be used, but that selecting at runtime eases development.

```c
mpath_absolute_from (mstrv_lit("\\foo\\bar"), mstrv_lit ("blah"), MPATH_WIN32));
```

src/mlib/path.h
```diff
@@ -7,66 +7,151 @@
...
- * @brief The preferred path separator for the current platform
- */
-static const char MPATH_PREFERRED_PATH_SEPARATOR =
+#ifndef _WIN32
+#include <unistd.h>
+#else
+#include <windows.h>
+#endif
+
+typedef enum mpath_format {
```
kevinAlbs, line 16: Curious, was the motivation for making the path format a runtime choice to ease development? I.e. you no longer need to use a Windows machine to test this?

src/mongocrypt-dll-private.h
```diff
@@ -0,0 +1,78 @@
...
+#endif
+
+#define MCR_DLL_NULL ((mcr_dll){._native_handle = NULL, .error_string = NULL})
+
+/**
+ * @brief A dynamically-loaded library i.e. returned by LoadLibrary() or
+ * dlopen()
+ */
+typedef struct mcr_dll {
+   // (All supported platforms using a void* as the library handle type)
```
kevinAlbs, line 23:
```suggestion
   // (All supported platforms use a void* as the library handle type)
```

src/os_win/os_dll.c
```diff
@@ -0,0 +1,60 @@
...
+   // Convert all slashes to the native Windows separator
+   mstr filepath =
+      mpath_to_format (MPATH_WIN32, mstrv_view_cstr (filepath_), MPATH_WIN32);
+   // Check if the path is just a filename.
+   bool is_just_filename =
+      mstr_eq (mpath_filename (filepath.view, MPATH_WIN32), filepath.view);
+   if (!is_just_filename) {
+      // If the path is only a filename, we'll allow LoadLibrary() to do a
+      // proper full DLL search. If the path is NOT just a filename, resolve the
+      // given path to a single unambiguous absolute path t suppress
```
kevinAlbs, line 26:
```suggestion
      // given path to a single unambiguous absolute path to suppress
```

kevinAlbs (approved):
Nicely done. And thank you for splitting up the changes into logical commits.

The handling of DLL loading on Windows (use default search path if filename, otherwise unambiguously use an absolute path) LGTM as well.

CMakeLists.txt
```diff
@@ -316,11 +331,18 @@ add_executable (test-mongocrypt ${TEST_MONGOCRYPT_SOURCES})
...
 target_link_libraries (test-mongocrypt PRIVATE mongocrypt_static)
 target_include_directories (test-mongocrypt PRIVATE ./src "${CMAKE_CURRENT_SOURCE_DIR}/kms-message/src")
-target_link_libraries (test-mongocrypt PRIVATE ${BSON_TARGET})
+target_link_libraries (test-mongocrypt PRIVATE ${BSON_TARGET} mongo::mlib)
 target_include_directories (test-mongocrypt PRIVATE ${BSON_INCLUDES} "${CMAKE_CURRENT_LIST_DIR}/test")
 target_compile_definitions (test-mongocrypt PRIVATE ${BSON_DEFINITIONS})
 
-add_test (mongocrypt test-mongocrypt WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+# Set a definition so that testcases can know where test-mongocrypt.exe was written to
+target_compile_definitions(test-mongocrypt PRIVATE "TEST_MONGOCRYPT_OUTPUT_PATH=\"$<TARGET_FILE:test-mongocrypt>\"")
```
rcsanchez97, line 339: Additional space needed before the opening parenthesis.
```suggestion
target_compile_definitions (test-mongocrypt PRIVATE "TEST_MONGOCRYPT_OUTPUT_PATH=\"$<TARGET_FILE:test-mongocrypt>\"")
```

rcsanchez97 (approved):
LGTM

src/mlib/path.h
```diff
@@ -7,66 +7,151 @@
...
+   MPATH_NATIVE =
 #ifdef _WIN32
-   '\\'
+      MPATH_WIN32,
 #else
-   '/'
+      MPATH_POSIX,
 #endif
-   ;
+   MPATH_NATIVE_ = 'n'
```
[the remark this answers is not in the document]
vector-of-bool: No, that's a mistake

src/mlib/path.h
```diff
@@ -7,66 +7,151 @@
...
+mpath_current_path ()
+{
+#if _WIN32
+   while (1) {
+      DWORD len = GetCurrentDirectoryW (0, NULL);
+      wchar_t *wstr = calloc (sizeof (wchar_t), len);
+      DWORD got_len = GetCurrentDirectoryW (len, wstr);
+      if (got_len > len) {
+         free (wstr);
+         continue;
```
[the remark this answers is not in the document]
vector-of-bool: Yes, there is a race between getting the size and using it.

src/mlib/path.h
vector-of-bool: I added a special case to work around this. Rather than crashing it will fall-back to the native format if `base` is non-absolute.

PR #249, opened 2022-02-12
Add instructions for Ubuntu 20.04

NunoSempere:
Current package is missing, but development branch exists. This commit points out the issue, perhaps until it can be fixed.


rcsanchez97:
@kevinAlbs I think that this is not the right thing to do.  We started publishing Ubuntu 20.04 packages with libmongocrypt 1.2.2.  So, I think that the right approach is to update `README.md` to change occurrences of `1.0` to `1.3`.

When I substituted `1.3` in the instructions on an Ubuntu 20.04 host, I get this:

```
ubuntu@ip-10-122-15-6:~$ apt-cache policy libmongocrypt-dev
libmongocrypt-dev:
  Installed: (none)
  Candidate: 1.3.1-0
  Version table:
     1.3.1-0 500
        500 https://libmongocrypt.s3.amazonaws.com/apt/ubuntu focal/libmongocrypt/1.3/universe amd64 Packages
     1.3.0-0 500
        500 https://libmongocrypt.s3.amazonaws.com/apt/ubuntu focal/libmongocrypt/1.3/universe amd64 Packages
```

Perhaps the release instructions could also be updated to ensure that `README.md` is changed on future point releases.

br4in3x:
I agree with @rcsanchez97. Version 1.0 causes following errors when following the [official tutorial on Client-Side Field Level Encryption](https://docs.mongodb.com/drivers/security/client-side-field-level-encryption-guide/#client-side-field-level-encryption-guide) in Go:
```
go.mongodb.org/mongo-driver/x/mongo/driver/mongocrypt
/home/max/go/pkg/mod/go.mongodb.org/mongo-driver@v1.8.4/x/mongo/driver/mongocrypt/mongocrypt.go:134:11: could not determine kind of name for C.mongocrypt_ctx_setopt_key_encryption_key
/home/max/go/pkg/mod/go.mongodb.org/mongo-driver@v1.8.4/x/mongo/driver/mongocrypt/mongocrypt.go:212:11: could not determine kind of name for C.mongocrypt_setopt_kms_providers
```
I found an answer here: https://github.com/mongodb/libmongocrypt/pull/249#issuecomment-1060116610, which also mentions updating from 1.0 to 1.3.

kevinAlbs:
Good catch @rcsanchez97. I agree the README should be updated to use 1.3, and the release instructions should be updated. Can you make that change?

rcsanchez97 (approved):

PR #250, opened 2022-02-15
Add keyMaterial to setopt ctx option validity tests

eramongodb:
A small followup to #246. Updates to `_mongocrypt_ctx_opts_spec_t` [did not include](https://github.com/mongodb/libmongocrypt/pull/246/commits/97ba5faed3e1213ec79a2a1f06fdb38a9b6c334e) the addition of test cases to corresponding test cases asserting permission or prohibition of the `keyMaterial` field in datakey options in different contexts.


kevinAlbs (approved):
LGTM. Good catch.

PR #256, opened 2022-03-03
Compile Warnings-as-errors Cleanup

vector-of-bool:
PHPC found: [MONGOCRYPT-390](https://jira.mongodb.org/browse/MONGOCRYPT-390). The failing conversion to `PINIT_ONCE_FN` only occurs on 32-bit MSVC builds, which we do not seem to have running in CI? It only requires adding a missing calling convention.

This code as a new warnings CMake module, replacing manual manipulation of C flags as part of the build.


CMakeLists.txt
```diff
@@ -22,6 +22,18 @@ endif ()
...
 set (CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake)
 
+
+# Enable a higher warning level and warnings-as-errors
+include (MongoC-Warnings)
+option (ENABLE_MORE_WARNINGS_AS_ERRORS "Enable extra warnings-as-errors in the build." ON)
+if (ENABLE_MORE_WARNINGS_AS_ERRORS)
+   mongoc_add_platform_compile_options (
+      msvc:/W3 msvc:/WX msvc:/wd4996
+      gnu-like:-Werror
```
kevinAlbs, line 32: Moving `-Werror` into CMakeLists.txt guarded by an option seems good. It is less platform specific flags required to pass into cmake.

@eramongodb are you OK with the proposed change? I ask since I initially thought this contradicts https://github.com/mongodb/libmongocrypt/pull/224. Since this behind an option, I am in favor of this change.

.evergreen/config.yml
```diff
@@ -147,7 +147,8 @@ functions:
     - command: "shell.exec"
       params:
         script: |-
-          ${compile_env|} CONFIGURE_ONLY=ON ./libmongocrypt/.evergreen/build_all.sh
+          env ${compile_env|} CONFIGURE_ONLY=ON CC=clang CXX=clang++ \
+              bash libmongocrypt/.evergreen/build_all.sh
```
kevinAlbs, line 151: Why this change?

kevinAlbs (approved):
LGTM

> The failing conversion to PINIT_ONCE_FN only occurs on 32-bit MSVC builds, which we do not seem to have running in CI?

Filed MONGOCRYPT-391 to add 32 bit Windows builds.

cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,74 @@
...
+     gnu-like:$<$<AND:${is_c_lang},${__not_old_gnu}>:-Werror=int-conversion> msvc:/we4047
+     # Discarding qualifiers
+     gnu:$<$<AND:${is_c_lang},${__not_old_gnu}>:-Werror=discarded-qualifiers>
+     clang:$<${is_c_lang}:-Werror=ignored-qualifiers>
+     msvc:/we4090
+     # Definite use of uninitialized value
+     gnu-like:-Werror=uninitialized msvc:/we4700
+
+     # Aside: Disable CRT insecurity warnings
+     msvc:/D_CRT_SECURE_NO_WARNINGS
```
eramongodb, line 73: I recommend this flag be moved into the root `CMakeLists.txt`. It's function/purpose does not fit the rest of the flags in this file that are defining unconditional errors (not disabling them).

.evergreen/build_all.sh
```diff
@@ -64,7 +55,7 @@ $CMAKE --build . --target install --config RelWithDebInfo
 # library to support consumers that static link to libmongocrypt
 find ${BSON_INSTALL_PREFIX} \( -name libbson-static-1.0.a -o -name bson-1.0.lib -o -name bson-static-1.0.lib \) -execdir cp {} $(dirname $(find ${MONGOCRYPT_INSTALL_PREFIX} -name libmongocrypt-static.a -o -name mongocrypt-static.lib)) \;
 $CMAKE --build . --target test-mongocrypt --config RelWithDebInfo
-$CMAKE --build ./kms-message --target test_kms_request --config RelWithDebInfo
+$CMAKE --build . --target test_kms_request --config RelWithDebInfo
```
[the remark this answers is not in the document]
vector-of-bool: Yes. The way it was written previously only worked incidentally because of the shape of some CMake generators. It breaks on e.g. Ninja generators.

.evergreen/config.yml
vector-of-bool: `clang-tidy` will read the compile options from `compile_commands.json` generated by CMake. By default CMake will select GCC as the compiler, and the new warnings include `-Wdiscarded-qualifiers`, which is a GCC-only option. Telling CMake to use Clang here will ensure that it generates a Clang-compatible `compile_commands.json`.

CMakeLists.txt
```diff
@@ -22,6 +22,18 @@ endif ()
...
+      gnu-like:-Wno-missing-braces
```
jmikola, line 33: Does this pertain to [MONGOCRYPT-389](https://jira.mongodb.org/browse/MONGOCRYPT-389)? If so, should that issue (currently still in "Needs Triage" status) also be referenced in this PR and advanced?

Also, I'll point out that [MONGOCRYPT-390](https://jira.mongodb.org/browse/MONGOCRYPT-390) is currently only mentioned in the PR description. You may want to add it to either the PR title or a commit message unless you indent to take care of that at merge time.

CMakeLists.txt
kevinAlbs: Thank you for the perspective. Having cmake scripts make few assumptions about the platform seems like a good goal.

.evergreen/config.yml
kevinAlbs: Thank you for the explanation. TIL `-Wdiscarded-qualifiers` is GCC-only.

CMakeLists.txt
vector-of-bool: I switched the default to `OFF`. I was debating which default would be preferrable. I wish there was an easy way to detect whether we're being compiled for someone hacking on the project rather than a user building the project, especially because it would be nice to enable extra warnings in CI and have developers see those warnings by default on their local project rather than waiting to see them appear after a push.

cmake/MongoC-Warnings.cmake
vector-of-bool: I (mostly) copied this file directly from the one I added to mongo-c-driver. If there were other warnings that we wanted to disable, I would add them here. It just happens that this is the only significant one I can think of at the time.

src/mongocrypt-dll-private.h
```diff
@@ -13,7 +13,8 @@
 #define MCR_DLL_SUFFIX ".so"
 #endif
 
-#define MCR_DLL_NULL ((mcr_dll){._native_handle = NULL, .error_string = NULL})
+#define MCR_DLL_NULL \
+   ((mcr_dll){._native_handle = NULL, .error_string = MSTR_NULL})
```
jmikola, line 17: This appears to just be a whitespace change. Am I correct to assume that this won't actually address [MONGOCRYPT-389](https://jira.mongodb.org/browse/MONGOCRYPT-389) and we'll continue to see a warning when compiling this file unless we explicitly it in our compiler flags? Not a problem, as it was a minor issue to begin with, but I just want to be clear on what might be required should we need to address this on the PHPC side.

vector-of-bool: The initializer for `error_string` was changed from `NULL` to `MSTR_NULL`. That was the cause of the warning, which wasn't obvious from the original diagnostic.

jmikola: Oh, I totally missed that while staring at the diff. Thank you!

.evergreen/build_all.sh
```diff
@@ -37,22 +37,13 @@ cd libmongocrypt
...
-fi
-
 for suffix in "dll" "dylib" "so"; do
     if test -f "mongo_csfle_v1.$suffix"; then
         ADDITIONAL_CMAKE_FLAGS="$ADDITIONAL_CMAKE_FLAGS -DMONGOCRYPT_TESTING_CSFLE_FILE=$PWD/mongo_csfle_v1.$suffix"
     fi
 done
 
-$CMAKE -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_CFLAGS} ${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ../
+$CMAKE -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ../
```
eramongodb, line 46:
```suggestion
$CMAKE -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DENABLE_MORE_WARNINGS_AS_ERRORS=ON DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}" -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ../
```

.evergreen/build_all.sh
```diff
@@ -88,7 +79,7 @@ fi
 cd libmongocrypt
 mkdir cmake-build-nocrypto
 cd cmake-build-nocrypto
-$CMAKE -DDISABLE_NATIVE_CRYPTO=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_CFLAGS} ${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}/nocrypto" ../
+$CMAKE -DDISABLE_NATIVE_CRYPTO=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}/nocrypto" ../
```
eramongodb, line 82:
```suggestion
$CMAKE -DDISABLE_NATIVE_CRYPTO=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DENABLE_MORE_WARNINGS_AS_ERRORS=ON -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}/nocrypto" ../
```

.evergreen/build_all.sh
```diff
@@ -99,6 +90,6 @@ cd $evergreen_root
 cd libmongocrypt
 mkdir cmake-build-sharedbson
 cd cmake-build-sharedbson
-$CMAKE -DENABLE_SHARED_BSON=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_CFLAGS} ${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}/sharedbson" ../
+$CMAKE -DENABLE_SHARED_BSON=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}/sharedbson" ../
```
eramongodb, line 93:
```suggestion
$CMAKE -DENABLE_SHARED_BSON=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS "${LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS}" -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_EXTRA_CFLAGS}" -DENABLE_MORE_WARNINGS_AS_ERRORS=ON -DCMAKE_PREFIX_PATH="${BSON_INSTALL_PREFIX}" "-DCMAKE_INSTALL_PREFIX=${MONGOCRYPT_INSTALL_PREFIX}/sharedbson" ../
```

bindings/node/etc/build-static.sh
```diff
@@ -41,21 +41,12 @@ popd #./deps/tmp
...
-else
-    # GNU, Clang, AppleClang
-    LIBMONGOCRYPT_CFLAGS="-fPIC -Werror"
-fi
-
 CMAKE_FLAGS="-DDISABLE_NATIVE_CRYPTO=1 -DCMAKE_INSTALL_LIBDIR=lib"
 if [ "$OS" == "Windows_NT" ]; then
   WINDOWS_CMAKE_FLAGS="-Thost=x64 -A x64 -DCMAKE_C_FLAGS_RELWITHDEBINFO=\"/MT\""
-  $CMAKE $CMAKE_FLAGS $WINDOWS_CMAKE_FLAGS -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_CFLAGS}" -DCMAKE_PREFIX_PATH="`cygpath -w $DEPS_PREFIX`" -DCMAKE_INSTALL_PREFIX="`cygpath -w $DEPS_PREFIX`" "`cygpath -w $LIBMONGOCRYPT_DIR`"
+  $CMAKE $CMAKE_FLAGS $WINDOWS_CMAKE_FLAGS -DCMAKE_PREFIX_PATH="`cygpath -w $DEPS_PREFIX`" -DCMAKE_INSTALL_PREFIX="`cygpath -w $DEPS_PREFIX`" "`cygpath -w $LIBMONGOCRYPT_DIR`"
```
eramongodb, line 47:
```suggestion
  $CMAKE $CMAKE_FLAGS $WINDOWS_CMAKE_FLAGS -DENABLE_MORE_WARNINGS_AS_ERRORS=ON -DCMAKE_PREFIX_PATH="`cygpath -w $DEPS_PREFIX`" -DCMAKE_INSTALL_PREFIX="`cygpath -w $DEPS_PREFIX`" "`cygpath -w $LIBMONGOCRYPT_DIR`"
```

bindings/node/etc/build-static.sh
```diff
@@ -41,21 +41,12 @@ popd #./deps/tmp
...
 else
-  $CMAKE $CMAKE_FLAGS -DCMAKE_C_FLAGS="${LIBMONGOCRYPT_CFLAGS}" -DCMAKE_PREFIX_PATH=$DEPS_PREFIX -DCMAKE_INSTALL_PREFIX=$DEPS_PREFIX -DCMAKE_OSX_DEPLOYMENT_TARGET="10.12" $LIBMONGOCRYPT_DIR
+  $CMAKE $CMAKE_FLAGS -DCMAKE_PREFIX_PATH=$DEPS_PREFIX -DCMAKE_INSTALL_PREFIX=$DEPS_PREFIX -DCMAKE_OSX_DEPLOYMENT_TARGET="10.12" $LIBMONGOCRYPT_DIR
```
eramongodb, line 49:
```suggestion
  $CMAKE $CMAKE_FLAGS -DCMAKE_PREFIX_PATH=$DEPS_PREFIX -DENABLE_MORE_WARNINGS_AS_ERRORS=ON -DCMAKE_INSTALL_PREFIX=$DEPS_PREFIX -DCMAKE_OSX_DEPLOYMENT_TARGET="10.12" $LIBMONGOCRYPT_DIR
```

eramongodb (changes requested):

eramongodb (approved):
I've marked areas in modified scripts where the new `ENABLE_MORE_WARNINGS_AS_ERRORS` config variable should be enabled. Otherwise, LGTM.

PR #257, opened 2022-03-04
MONGOCRYPT-382 require `aws: {}` to enter NEED_KMS_CREDENTIALS

kevinAlbs:
# Background & Motivation
https://github.com/mongodb/libmongocrypt/pull/252 introduced a new context state, `MONGOCRYPT_CTX_NEED_KMS_CREDENTIALS`. In the `NEED_KMS_CREDENTIALS` state, `mongocrypt_ctx_provide_kms_providers()` may be called to provide credentials on-demand. This PR is a follow-up to address two caveats:

1. Users are expected to configure `aws: {}` in the KMS providers in the driver to opt into on-demand credentials. libmongocrypt errors on `aws: {}` set in the KMS providers. Drivers need to remove, and remember, if an empty `aws: {}` is configured on the KMS providers.
2. Drivers request aws credentials without needing them in data key contexts. For example: suppose `aws: {}` and `azure` are configured as KMS providers. Calling `createDataKey("azure")` from the driver will result in the libmongocrypt entering the `NEED_KMS_CREDENTIALS` state, and the driver will refresh `aws` credentials.

# Summary
- Require `aws: {}` to be set in `mongocrypt_setopt_kms_providers` to enter the `NEED_KMS_CREDENTIALS` state.
- Only enter the `NEED_KMS_CREDENTIALS` state for a data key context if creating an "aws" data key.
- Require a non-empty document for `mongocrypt_setopt_kms_providers` and `mongocrypt_ctx_provide_kms_providers`.

# Notes
I did not add a function like `mongocrypt_ctx_get_requested_kms_provider`. It was proposed in https://github.com/kevinAlbs/mongo-go-driver/pull/1#issuecomment-1058214761. I did not think it provided value to add now. I filed MONGOCRYPT-394 to track supporting on-demand credentials for other KMS providers. If / when we do it, libmongocrypt and drivers will both need additive changes.

The [Go POC has been updated](https://github.com/kevinAlbs/mongo-go-driver/pull/1). It no longer needs to check and remove `aws: {}` from the KMS providers.

### DIFF
--- a/bindings/node/test/autoEncrypter.test.js
+++ b/bindings/node/test/autoEncrypter.test.js
@@ -198,6 +198,9 @@ describe('AutoEncrypter', function () {
       const mc = new AutoEncrypter(client, {
         keyVaultNamespace: 'admin.datakeys',
         logger: () => {},
+        kmsProviders: {
+          aws: {}
+        },
         async onKmsProviderRefresh() {
           return { aws: { accessKeyId: 'example', secretAccessKey: 'example' } };
         }
--- a/bindings/node/test/autoEncrypter.test.js
+++ b/bindings/node/test/autoEncrypter.test.js
@@ -149,6 +149,9 @@ describe('AutoEncrypter', function () {
       const mc = new AutoEncrypter(client, {
         keyVaultNamespace: 'admin.datakeys',
         logger: () => {},
+        kmsProviders: {
+          aws: {}
+        },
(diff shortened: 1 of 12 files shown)


jyemin (approved):
LGTM on the idea, but not on the implementation, which I haven't looked at.

vector-of-bool (approved):
LGTM

PR #261, opened 2022-03-09
MONGOCRYPT-392 add crypto tokens for FLE 2

kevinAlbs:
# Summary
Add crypto tokens for FLE 2.

# Background & Motivation
Each token is represented as a separate type. It is more verbose. The rationale is to reduce risk of using the wrong token.

Tokens are defined in https://github.com/markbenvenuto/mongo-enterprise-modules/blob/fle2/fle_protocol.md#reference-keys-and-tokens. The "TwiceDerived" tokens are only used server-side and are not generated in libmongocrypt. ClientUserDataEncryptionLevel1Token was removed in a recent revision of the design (see [slack conversation](https://mongodb.slack.com/archives/C0316KB5VM4/p1646765242204479?thread_ts=1646762387.874499&cid=C0316KB5VM4)) and is not generated in libmongocrypt.

I used the server implementation as a reference: https://github.com/mongodb/mongo/blob/master/src%2Fmongo%2Fcrypto%2Ffle_crypto.h#L189


CMakeLists.txt
```diff
@@ -184,7 +185,16 @@ target_include_directories (
    PUBLIC
       "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>"
    )
-target_compile_definitions (mongocrypt PRIVATE ${BSON_DEFINITIONS})
+include (TestBigEndian)
+TEST_BIG_ENDIAN (MONGOCRYPT_BIG_ENDIAN)
+if (MONGOCRYPT_BIG_ENDIAN)
+   message ("Detected byte order: big endian")
+   set (MONGOCRYPT_DEFINITIONS "${MONGOCRYPT_DEFINITIONS}" MONGOCRYPT_BIG_ENDIAN)
```
vector-of-bool, line 192: Prefer `list(APPEND)`:

```cmake
list(APPEND MONOGCRYPT_DEFINITIONS MONGOCRYPT_BIG_ENDIAN)
```

(See also: `list(INSERT)`)

src/mc-tokens-private.h
```diff
@@ -0,0 +1,203 @@
...
+ * ECCDerivedFromDataToken = HMAC(ECCToken, v)
+ *
+ * EDCDerivedFromDataTokenAndCounter = HMAC(EDCDerivedFromDataToken, u)
+ * ESCDerivedFromDataTokenAndCounter = HMAC(ESCDerivedFromDataToken, u)
+ * ECCDerivedFromDataTokenAndCounter = HMAC(ECCDerivedFromDataToken, u)
+ * ======================== End: FLE 2 Token Reference ========================
+ */
+
+
+typedef struct _mc_CollectionsLevel1Token_t mc_CollectionsLevel1Token_t;
```
vector-of-bool, line 50: A lot of this looks duplicated. I played around and created two preprocessor macros that emit equivalent code, as far as I can tell. I've attached a patch with the difference.
[macros.txt](https://github.com/mongodb/libmongocrypt/files/8276616/macros.txt)

vector-of-bool (commented):
I see no major issues or concerns, except I notice a lot of duplication in the `mc-tokens` files. I left a comment and suggestion on that, but you can ignore if you think my solution won't be viable in the long-term.

CMakeLists.txt
kevinAlbs: Good thinking. Done.

kevinAlbs:
> I see no major issues or concerns, except I notice a lot of duplication in the `mc-tokens` files. I left a comment and suggestion on that, but you can ignore if you think my solution won't be viable in the long-term.

Thank you for the patch! I like the macros, and have applied this patch.

vector-of-bool (approved):

src/mc-tokens.c
```diff
@@ -0,0 +1,312 @@
...
+      _mongocrypt_buffer_cleanup (&self->data);                      \
+      bson_free (self);                                              \
+   }                                                                 \
+   /* Constructor. Parameter list given as variadic args. */         \
+   T *CONCAT (Prefix, _new) (_mongocrypt_crypto_t * crypto,          \
+                             __VA_ARGS__,                            \
+                             mongocrypt_status_t * status)
+
+DEF_TOKEN_TYPE (mc_CollectionsLevel1Token, const _mongocrypt_buffer_t *RootKey)
+{
```
sgolemon, line 52: I know I snickered at the macro stacking in slack a little, but I think we can probably go a little further to mirror what we have in the lovely comment above.  Take this as a shared idea, not a requirement for refactor.  It's got pros-and-cons to it.

```
#define IMPL_TOKEN_NEW_1 (Name, Key, Arg, Clean) \
{ \
  CONCAT (Name, _t)* t = bson_malloc (sizeof (CONCAT (Name, T))); \
  _mongocrypt_buffer_init (&t->data); \
  _mongocrypt_buffer_resize (&t->data, MONGOCRYPT_HMAC_SHA256_LEN); \
\
   if (!_mongocrypt_hmac_sha_256 ( \
          crypto, Key, Arg, &t->data, status)) { \
      CONCAT(Name, _destroy) (t); \
      Clean;
      return NULL; \
   } \
   Clean;
   return t; \
}

// Use this form for hashing with a mongocrypt buffer with dynamic contents.
#define IMPL_TOKEN_NEW_1 (Name, Key, Arg) \
    IMPL_TOKEN_NEW_1 (name, Key, Arg)
    
// Use this form for lower level key derivation using fixed constants.
#define IMPL_TOKEN_NEW_CONST(Name, Key, Arg) \
{ \
    _mongocrypt_buffer_t to_hash; \
    _mongocrypt_buffer_copy_from_uint64 (&to_hash, MONGOCRYPT_UINT64_TO_LE (Arg) ); 
    IMPL_TOKEN_NEW_1 (Name, Key, to_hash, _mongocrypt_buffer_cleanup (&to_hash)) \
}
```

Then each type implementation can be:

```
DEF_TOKEN_TYPE (mc_CollectionsLevel1Token, const _mongocrypt_buffer_t *RootKey)
IMPL_TOKEN_NEW_CONST (mc_CollectionsLevel1Token, RootKey, 1)

DEF_TOKEN_TYPE (mc_ServerDataEncryptionLevel1Token, const _mongocrypt_buffer_t *RootKey)
IMPL_TOKEN_NEW_CONST (mc_ServerDataEncryptionLevel1Token, RootKey, 3)

DEF_TOKEN_TYPE (mc_EDCToken, const mc_CollectionsLevel1Token_t *CollectionsLevel1Token)
IMPL_TOKEN_NEW_CONST (mc_EDCToken, mc_CollectionsLevel1Token_get (CollectionsLevel1Token), 1)

// etc...

DEF_TOKEN_TYPE (mc_EDCDerivedFromDataToken, const mc_EDCToken_t *EDCToken, const _mongocrypt_buffer_t *v)
IMPL_TOKEN_NEW (mc_EDCDerivedFromDataToken, mc_EDCToken_get(EDCToken), v)

// etc...
```

Plus or minus a bit to improve readability and bring the usages closer to looking like their documented formats.

src/mongocrypt-buffer-private.h
```diff
@@ -201,4 +201,13 @@ bool
...
                                       char *str) MONGOCRYPT_WARN_UNUSED_RESULT;
 
+/* _mongocrypt_buffer_copy_from_uint64 initializes @buf from the byte
+ * representation of @value.
+ * Caller must call _mongocrypt_buffer_cleanup.
+ * Endianness of @value is not modified. Use mongocrypt-endian-private.h
+ * to convert endianness beforehand.
+ */
+void
+_mongocrypt_buffer_copy_from_uint64 (_mongocrypt_buffer_t *buf, uint64_t value);
```
sgolemon, line 211: Not sure what the local convention is, perhaps we always just assume LE since that's what's in the BSON spec, but if we're inventing a new part of the API, it'd be great to be explicit about our endianness since it is not machine dependent.

src/mongocrypt-buffer.c
```diff
@@ -532,3 +532,11 @@ _mongocrypt_buffer_steal_from_string (_mongocrypt_buffer_t *buf, char *str)
    buf->owned = true;
    return true;
 }
+
+void
+_mongocrypt_buffer_copy_from_uint64 (_mongocrypt_buffer_t *buf, uint64_t value)
+{
+   _mongocrypt_buffer_init (buf);
+   _mongocrypt_buffer_resize (buf, 8);
+   memcpy (buf->data, &value, buf->len);
```
sgolemon, line 541: ~Okay, I take back my comment above. This appears to be machine dependent.  Is that your intent? I thought this was for representing a BSON int64 which needs to be fixed as LE. This will succeed on LE architectured, but fail on BE.~

Edit: I take back my take back, it looks like the macros below deal with endianness for us.  Still, at the minimum I'd like a comment reminding the reader that this particular function is NOT endian safe by itself and MUST be called with the endian conversion macros to be reliable across platforms.

Also, probably shouldn't hardcode the size at '8', but use `sizeof (value)`.

sgolemon (approved):
LGTM
All comments are opinions and not required.

src/mongocrypt-buffer-private.h
kevinAlbs: Good point. I have updated this to suffix with `_le` and put the endian conversion in the body of the function.

src/mongocrypt-buffer.c
kevinAlbs: Updated to use `sizeof (value)`.

src/mc-tokens.c
kevinAlbs: Thank you for the suggested code. I like the idea of getting this closer to the documented formats. Updated.

PR #263, opened 2022-03-09
Integrate csfle dynamic library as part of mongocrypt's encryption routines

vector-of-bool:
When `mongocrypt_ctx` transitions into `NEED_MARKINGS` state, it will additionally call `_try_run_csfle_marking()`, which will try to obtain the marked command from the csfle library. If no library is loaded, this function is a no-op. Otherwise, it generates the marked command and transitions to the next state in the process. This replaces the need to use mongocryptd to perform query analysis.

~~TODO: The csfle library mandates a single global `mongo_csfle_v1_lib` object for the entire application. This can be particularly hairy if we are loading/unloading the library, and especially if more than one instance of `mongocrypt_t` is created and they are destroyed out-of-order. This could require a global reference counting, and therefore some global synchronization around the object.~~


kevinAlbs (commented):
There were two questions answered over slack that change the scope of this PR:

Q: Is the limit of one active mongo_csfle_v1_lib per process feasible to remove?
    Based on the [global initialization/destruction](https://github.com/10gen/mongo-enterprise-modules/blob/b1464938b7f3bc81f47625ef02561044175926b5/src/fle/lib/mongo_csfle.cpp#L252-L279) my guess is “no”.
A: No, it is not feasible to support multiple mongo_csfle_v1_lib. The code is derived from the server which does not support multiple copies of itself in on process. There are a lot of shared global variables

Q: Is loading two copies of the csfle library in the same process not supported?
    My guess is “no”. I think I goofed in the scope, and will need to revise.
A: No

src/mongocrypt.c
```diff
@@ -574,30 +530,403 @@ mongocrypt_init (mongocrypt_t *crypt)
...
+                  old_state.vtable.status_get_code (status));
+      }
+      old_state.vtable.status_destroy (status);
+
+#ifndef __linux__
+      mcr_dll_close (old_state.dll);
+#endif
+      /// NOTE: On Linux, skip closing the CSFLE library itself, since a bug in
+      /// the way ld-linux and GCC interact causes static destructors to not run
+      /// during dlclose(). Still, free the error string:
```
addaleax, line 665: As mentioned on SERVER-63710, this isn’t actually Linux-specific, and not a bug in GCC/ld-linux either, so it seems like we should a) either skip the unloading everywhere or nowhere, and b) adjust the comment:

```suggestion
      /// NOTE: Skip closing the CSFLE library itself, since the CSFLE library 
      /// performs static initialization without corresponding static cleanup,
      /// so loading it repeatedly would result in memory leaks.
```

src/mongocrypt.c
```diff
@@ -574,30 +530,403 @@ mongocrypt_init (mongocrypt_t *crypt)
...
+      mcr_dll_path_result found_path = mcr_dll_path (found.lib);
+      assert (found_path.path.data && "Failed to get the DLL path of the "
+                                      "library that we just barely found");
+      if (!mstr_eq (found_path.path.view, existing_path)) {
+         // Our find-result should only ever find the existing same library.
+         // Error:
+         CLIENT_ERR (
+            "An existing CSFLE library is loaded by the application "
+            "at [%s], but the current call to mongocrypt_init() attempted to "
+            "load a second CSFLE library from [%s]. This is not allowed.",
```
addaleax, line 610: It might be nice to have a comment in the source about the actual problem that this is solving.

Would the two library instances conflict with each other in some way? (The global state shouldn’t really conflict, esp. since we’re using `RTLD_LOCAL` on POSIX and Windows has that behavior anyway). At a glance, it seems like this would only be leaking memory, but that seems fine-ish.

Also, keep in mind that there may still be multiple copies of libmongocrypt loaded at the same time, or the same copy loaded and unloaded repeatedly, and consequently it’s hard to actually guarantee singleton-ness of the csfle shared library from here.

src/os_posix/os_dll.c
```diff
@@ -41,3 +59,62 @@ mcr_dll_sym (mcr_dll dll, const char *sym)
...
+
+#include <mach-o/dyld.h>
+#include <mach-o/nlist.h>
+
+mcr_dll_path_result
+mcr_dll_path (mcr_dll dll)
+{
+   // Clear the three low bits of the module handle:
+   uintptr_t needle = ((uintptr_t) dll._native_handle & ~UINT64_C (0x3));
+   /// NOTE: Not thread safe. Is there a thread-safe way to do this?
```
addaleax, line 73: I think at least one reasonable protection here would be to handle a `dyld_name == NULL` result below.

src/mongocrypt.c
vector-of-bool: Whether two loaded csfle libraries will conflict (even if they share distinct memory mappings) is up to the server team. The library they have given us assumes that it will be the only loaded instance for the entire process, and this code is just to ensure those guarantees as best as we can, even if not perfect.

src/os_posix/os_dll.c
vector-of-bool: Maybe, but if `idx` can be invalidated between the loop beginning and `_dyld_get_image_name()`, then the returned pointer can also be invalidated between the call of `_dyld_get_image_name()` and `dlopen()` (Since `dyld` owns that pointer).

addaleax: Yeah, good point. I think what you have come up with in [1b84fed](https://github.com/mongodb/libmongocrypt/pull/263/commits/1b84fed369ac63373cb5d286e042c478e8bfd0db) seems like a pretty good solution.

src/mongocrypt.c
addaleax: Thank you for the clarification – I only did a :+1: reaction and left this comment unresolved because I still think it might be nice to have a comment in the source about why we’re doing this, but ultimately, this concern is addressed :)

(It’s a bit sad that the server team can’t give us stronger guarantees here, but I also get why they don’t.)

addaleax (approved):
Nice work! I still think it would be nice to address the comment about static destructors not being a Linux-specific issue, but either way, this looks good from my side :+1:

src/mongocrypt.c
```diff
@@ -574,30 +530,403 @@ mongocrypt_init (mongocrypt_t *crypt)
...
+      /* set default hooks. */
+      crypt->crypto = bson_malloc0 (sizeof (*crypt->crypto));
+      BSON_ASSERT (crypt->crypto);
+#endif
+   }
+
+   if (!_wants_csfle (crypt)) {
+      // User does not want csfle. Just succeed.
+      return true;
+   }
```
kevinAlbs, line 927: This is vague in the scope document. The behavior I have in mind:

1. Try loading csfle from user provided paths
2. Otherwise if user did not provide paths, try loading csfle from system paths
3. Otherwise, use mongocryptd

Is it OK to modify `_try_find_csfle` search with `$SYSTEM` if neither `mongocrypt_setopt_append_csfle_search_path` or `mongocrypt_setopt_set_csfle_lib_path_override` were called?

kevinAlbs (commented):
This looks solid. The use of meaningful methods in mongocrypt.c is helpful for understanding. I like the `MONGOCRYPT_WITH_MUTEX` macro.

I only have one substantive comment. I request attempting to load from the system default path when no overridden path or search paths are set.

CMakeLists.txt
```diff
@@ -301,6 +300,19 @@ if (BUILD_TESTING)
       PREFIX ""
       )
 
+   # Create two stubbed csfle libraries
+   add_library (stubbed-csfle SHARED test/csfle-stub.cpp)
+   add_library (stubbed-csfle-2 SHARED test/csfle-stub.cpp)
+
+   set_target_properties(stubbed-csfle stubbed-csfle-2 PROPERTIES
+      INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}/src"
+      LINK_LIBRARIES "mongo::bson_static"
```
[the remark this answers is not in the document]
kevinAlbs: > Does that sound OK? If so, I can submit a PR to only build the `mongocrypt` target on 32 bit Windows.

Here is a [branch with the proposed changes](https://github.com/vector-of-bool/libmongocrypt/compare/use-csfle-lib...kevinAlbs:32bit-windows?expand=1).

On second thought, I disagree with my proposal. If PHPC builds libmongocrypt from source and runs on 32 bit Windows, that seems like reason to run tests of libmongocrypt.

Since this warning is only in the stubbed-csfle library, can the warning be ignored with pragmas?

src/mongocrypt.c
vector-of-bool: If that were the default behavior, how might a driver disable using the csfle library? My intuition is that "If you give me zero places to look, I will not look."

I feel like this option should be deferred to drivers, which can add `$SYSTEM` as the sole search path if they want that "default" behavior. e.g.:

```py
def init_mongocrypt(lib: MongoCryptLibrary, opts: InitOptions):
    if opts.enable_csfle:
        add_csfle_paths(lib)
    lib.init()

def add_csfle_paths(lib: MongoCryptLibrary, opts: InitOptions):
    for p in opts.paths_search:
        lib.add_csfle_search_path(p)
    p.add_csfle_search_path('$SYSTEM')
    for p in opts.paths_fallback:
        lib.add_csfle_search_path(p)
```

kevinAlbs: Good point. Having drivers apply the default by adding `$SYSTEM` as the sole search path SGTM.

kevinAlbs (approved):
Changes LGTM. Please resolve the Windows 32 bit build failure before merging.

vector-of-bool:
@kevinAlbs I had to do a merge with the FLE2 changes from `master`. This required a bit of reworking of the markings command generation, since both changes tweaked it. Could you do another check on the `_mongo_op_markings` function in `mongocrypt-ctx-encrypt.c` to make sure it all looks okay to you?

src/mongocrypt-ctx-encrypt.c
```diff
@@ -244,42 +203,84 @@ _fle2_mongo_op_markings (mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out)
...
 {
-   _mongocrypt_ctx_encrypt_t *ectx;
+   _mongocrypt_ctx_encrypt_t *ectx = (_mongocrypt_ctx_encrypt_t *) ctx;
 
-   ectx = (_mongocrypt_ctx_encrypt_t *) ctx;
    if (_mongocrypt_buffer_empty (&ectx->mongocryptd_cmd)) {
+      // We need to generate the command document
       bson_t cmd_bson;
       if (!_create_markings_cmd_bson (ctx, &cmd_bson)) {
+         // Failed
```
kevinAlbs, line 276:
```suggestion
         // Failed
         bson_destroy (&cmd_bson)
```

src/mongocrypt-ctx-encrypt.c
```diff
@@ -244,42 +203,84 @@ _fle2_mongo_op_markings (mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out)
...
 static bool
 _mongo_op_markings (mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out)
 {
-   _mongocrypt_ctx_encrypt_t *ectx;
+   _mongocrypt_ctx_encrypt_t *ectx = (_mongocrypt_ctx_encrypt_t *) ctx;
 
-   ectx = (_mongocrypt_ctx_encrypt_t *) ctx;
    if (_mongocrypt_buffer_empty (&ectx->mongocryptd_cmd)) {
+      // We need to generate the command document
       bson_t cmd_bson;
```
kevinAlbs, line 274:
```suggestion
      bson_t cmd_bson = BSON_INITIALIZER;
```

If L245 returns,out may be left uninitialized.

If L255 returns false, out may be initialized to a non-empty BSON document. The caller needs to call bson_destroy to prevent leaking. bson_t heap allocates when the size of the BSON document exceeds 120 bytes.

Initializing `cmd_bson` is consistent with the other call to `_create_markings_cmd_bson`.

kevinAlbs (approved):
Thank you for checking. LGTM with a leak fix.

PR #264, opened 2022-03-11
MONGOCRYPT-411 Support releases from branches

kevinAlbs:
# Summary
This makes several changes to support releases of libmongocrypt from a non-master release branch on a separate Evergreen project.

- Update the URLs in upload-all to set `${project}` to "libmongocrypt" and `${branch_name}` to "master". Use `${libmongocrypt_s3_suffix_copy}` of "latest-${branch}" for non-master branch. This avoids breaking workflows for bindings tasks.
- Update the URL in windows-upload to set `${project}` to "libmongocrypt" to avoid breaking workflows for users.
- Modify the libmongocrypt release process to create a branch before tagging. Update the libmongocrypt-release Evergreen project with the new branch.
- Do not run the publish-snapshot task on non-master branches.

# Background & Motivation
Please see [MONGOCRYPT-411 Proposal](https://docs.google.com/document/d/1QTWrK_w6UciKgDR-pQ2BkAT6Ignzs8uN03Mu_Wxn36E/edit#) for background and motivation. Please leave comments in that document.

After this is merged and a libmongocrypt-release project is created in [EVG-16517](https://jira.mongodb.org/browse/EVG-16517), the next proposed step is to release 1.3.2-rc0 test the process before doing a stable 1.3.2 release.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -643,7 +643,7 @@ tasks:
       params:
         aws_key: '${aws_key}'
         aws_secret: '${aws_secret}'
-        remote_file: '${project}/windows-test/${branch_name}/${libmongocrypt_s3_suffix}/libmongocrypt.tar.gz'
+        remote_file: 'libmongocrypt/windows-test/master/${libmongocrypt_s3_suffix}/libmongocrypt.tar.gz'
         bucket: mciuploads
         extract_to: libmongocrypt_download
     - command: shell.exec
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -190,6 +190,10 @@ functions:
           if [ "${is_patch}" = "true" ]; then
             echo "Patch build detected, skipping"
           fi
+          if [ "${branch}" != "master" ]; then
+            echo "publish snapshot is only run on master branch."
+            echo "Detected branch '${branch}', skipping."
+          fi
     - command: shell.exec
       params:
         silent: true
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -190,6 +190,10 @@ functions:
           if [ "${is_patch}" = "true" ]; then
             echo "Patch build detected, skipping"
           fi
+          if [ "${branch}" != "master" ]; then
+            echo "publish snapshot is only run on master branch."
+            echo "Detected branch '${branch}', skipping."
+          fi
```
rcsanchez97, line 196: You will also need to add a check in the next section:

```
           if [ "${is_patch}" = "true" ]; then
             exit 0
           fi
+          if [ "${branch}" != "master" ]; then
+            exit 0
+          fi
```

That said, it's not obvious to me why there is a check with a message in one `command` section and then the same check with the `exit` in the next `command` section.  These two sections could probably be merged, but that's up to you.

rcsanchez97 (changes requested):
Just one requested change.

.evergreen/config.yml
kevinAlbs: Good catch. Updated. The sections cannot be merged since the second `shell.exec` has `silent: true`. I assume the reason is to prevent logging credentials.

rcsanchez97: Ah, that makes sense why it is in two separate sections.

rcsanchez97 (approved):
LGTM

PR #265, opened 2022-03-15
feat(NODE-4085): add Node.js csfle shared library option support

addaleax:
Node.js driver typings: https://github.com/mongodb/node-mongodb-native/pull/3179

### DIFF
--- a/bindings/node/src/mongocrypt.cc
+++ b/bindings/node/src/mongocrypt.cc
@@ -93,7 +93,8 @@ Function MongoCrypt::Init(Napi::Env env) {
                     InstanceMethod("makeDecryptionContext", &MongoCrypt::MakeDecryptionContext),
                     InstanceMethod("makeExplicitDecryptionContext", &MongoCrypt::MakeExplicitDecryptionContext),
                     InstanceMethod("makeDataKeyContext", &MongoCrypt::MakeDataKeyContext),
-                    InstanceAccessor("status", &MongoCrypt::Status, nullptr)
+                    InstanceAccessor("status", &MongoCrypt::Status, nullptr),
+                    InstanceAccessor("csfleVersionInfo", &MongoCrypt::CSFLEVersionInfo, nullptr)
                   });
 }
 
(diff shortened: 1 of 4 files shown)


bindings/node/src/mongocrypt.cc
```diff
@@ -418,15 +419,46 @@ MongoCrypt::MongoCrypt(const CallbackInfo& info)
...
+        Napi::Value search_paths_v = options["csfleSearchPaths"];
+        if (!search_paths_v.IsArray()) {
+            throw TypeError::New(Env(), "Option `csfleSearchPaths` must be an array");
+        }
+        Array search_paths = search_paths_v.As<Array>();
+        for (uint32_t i = 0; i < search_paths.Length(); i++) {
+            mongocrypt_setopt_append_csfle_search_path(
+                _mongo_crypt.get(),
+                search_paths.Get(i).ToString().Utf8Value().c_str());
+        }
```
addaleax, line 432: @kevinAlbs Can you confirm that it is expected that `mongocrypt_setopt_append_csfle_search_path()` is exposed in this way (or a similar one)? I want to make sure because it’s not listed in the scope, but is part of the mongocrypt.h public API.

kevinAlbs: Yes, there is no plan to change the declaration or behavior of [mongocrypt_setopt_append_csfle_search_path](https://github.com/mongodb/libmongocrypt/blob/f1582c5a6fc09bbdda3f84382d4eed55f78af93e/src/mongocrypt.h.in#L440-L443).

bindings/node/test/autoEncrypter.test.js
```diff
@@ -538,4 +608,52 @@ describe('AutoEncrypter', function () {
...
+        kmsProviders: {
+          aws: { accessKeyId: 'example', secretAccessKey: 'example' },
+          local: { key: Buffer.alloc(96) }
+        },
+        extraOptions: {
+          csfleSearchPaths: [path.dirname(sharedLibraryStub)]
+        }
+      });
+
+      expect(this.mc._mongocryptdManager).to.equal(undefined);
```
dariakp, line 650: Just a small nit: could we use the chai `property` assertion in all the tests here? it makes error logs prettier if this ever fails and differentiates between props whose value is undefined vs those that aren't set at all, which makes the intent clearer:

```js
// if we expect that the prop exists on the object
expect(obj.prop).to.equal('something'); 
// => expected undefined to equal 'something'
expect(obj).to.have.property('prop', 'something'); 
// => if the prop is there:
// 	=> expected obj to have property 'prop' of 'something' but got undefined
// => if the prop was never set:
//	=> expected obj to have property 'prop'

// if we expect that the prop does not exist on the object
expect(obj.prop).to.equal(undefined);
// => expected 'something' to equal undefined
expect(obj).to.not.have.property('prop');
// => expected obj to not have property 'prop'
```

dariakp (changes requested):
Just a small nit on testing, but I'd like someone with more C++ experience to look at that part of the code

baileympearson (commented):
This looks good to me, once Daria's comment is approved.  Since it's not in CI, do we want to run a patch from the node driver that uses this PR to confirm we're not breaking the driver?  

Just as a sanity check 😄

bindings/node/test/autoEncrypter.test.js
addaleax: Sure, done! :+1:

addaleax:
>  Since it's not in CI, do we want to run a patch from the node driver that uses this PR to confirm we're not breaking the driver?

Feel free to do that – since this is purely additive, I would be very surprised if something shows up, though :)

dariakp (approved):
LGTM! Thanks!

baileympearson:
Patch on the node driver passed: https://spruce.mongodb.com/version/62421e6cd1fe07433f84d73e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

baileympearson (approved):

baileympearson (approved):

PR #266, opened 2022-03-15
feat(NODE-3978): allow identifying decrypted fields for internal usage/testing

addaleax:


bindings/node/lib/autoEncrypter.js
```diff
@@ -276,3 +284,47 @@ module.exports = function (modules) {
...
+ * we do not need to worry about circular references.
+ */
+function decorateDecryptionResult(decrypted, original, bson, isTopLevelDecorateCall = true) {
+  if (isTopLevelDecorateCall) {
+    // The original value can have been either a JS object or a BSON buffer
+    if (Buffer.isBuffer(original)) {
+      original = bson.deserialize(original);
+    }
+    if (Buffer.isBuffer(decrypted)) {
+      throw new Error('Expected result of decryption to be deserialize BSON object');
```
nbbeeken, line 302: Despite this error being pretty fatal to the process we're trying to do here, as would be bson.deserializing errors, I think we should make sure we funnel any throws here through the callback to `decrypt`.

bindings/node/lib/autoEncrypter.js
```diff
@@ -276,3 +284,47 @@ module.exports = function (modules) {
 
   return { AutoEncrypter };
 };
+
+const decryptedKeys = Symbol.for('@@mdb.decryptedKeys');
```
nbbeeken, line 288: This is probably a non-issue, I haven't used the global registry for symbols much. Is it important that this be registered as early as possible? Maybe it should be moved to be above the `module.exports = function (modules) {` module func that wraps everything in here.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -143,6 +143,27 @@ describe('AutoEncrypter', function () {
...
+        kmsProviders: {
+          aws: { accessKeyId: 'example', secretAccessKey: 'example' },
+          local: { key: Buffer.alloc(96) }
+        }
+      });
+      mc[Symbol.for('@@mdb.decorateDecryptionResult')] = true;
+      mc.decrypt(input, (err, decrypted) => {
+        if (err) return done(err);
+        expect(decrypted).to.eql({ filter: { find: 'test', ssn: '457-55-5462' } });
+        expect(decrypted[Symbol.for('@@mdb.decryptedKeys')]).to.eql(undefined);
```
nbbeeken, line 161: The current implementation will define the array only if it is needed, so maybe this assertion could be explicit about that:
```js
expect(decrypted).to.not.have.property(Symbol.for('@@mdb.decryptedKeys'));
```

But it raises the question would it be more convenient to define it always so `.length` is always safe to check? but then again maybe that's unnecessary overhead, thoughts?

bindings/node/lib/autoEncrypter.js
```diff
@@ -259,7 +259,15 @@ module.exports = function (modules) {
         proxyOptions: this._proxyOptions,
         tlsOptions: this._tlsOptions
       });
-      stateMachine.execute(this, context, callback);
+
+      const decorateResult = this[Symbol.for('@@mdb.decorateDecryptionResult')];
```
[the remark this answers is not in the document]
addaleax: Yes, this is exactly what I did to try this out (and what I imagine we will do in practice as well) :slightly_smiling_face:

bindings/node/lib/autoEncrypter.js
addaleax: Yeah, makes sense – done! :+1:

bindings/node/lib/autoEncrypter.js
addaleax: I think it doesn’t really matter where it’s defined, I just moved it into `decorateDecryptionResult` itself if that works for you

bindings/node/test/autoEncrypter.test.js
addaleax: Sure, works for me.

> But it raises the question would it be more convenient to define it always so `.length` is always safe to check? but then again maybe that's unnecessary overhead, thoughts?

I’m not so much worried about performance here, but I’m a little hesitant to make the presence of the property for all objects something that consuming code should rely on, tbh.

bindings/node/lib/autoEncrypter.js
```diff
@@ -276,3 +284,47 @@ module.exports = function (modules) {
...
+
+  if (!decrypted || typeof decrypted !== 'object') return;
+  for (const k of Object.keys(decrypted)) {
+    const originalValue = original[k];
+
+    // An object was decrypted by libmongocrypt if and only if it was
+    // a BSON Binary object with subtype 6.
+    if (
+      originalValue &&
+      originalValue.constructor.name === 'Binary' &&
```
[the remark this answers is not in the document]
nbbeeken: I'm not seeing any changes on my end, forget to push maybe?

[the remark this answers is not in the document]
addaleax: yes 😇

bindings/node/test/autoEncrypter.test.js
nbbeeken: Yea that makes sense to me too. Semi-related, should the array be frozen after being filled? I think it would be important that no keys get removed, although it is a feature used in the more or less locked ecosystem of shell/compass, seems safer to guarantee finality at this layer. overkill?

bindings/node/lib/autoEncrypter.js
nbbeeken: Yea I supposed as long as the string matches it doesn't matter where the first Symbol.for('xx') call is made. Boxing it inside the helper function LGTM

bindings/node/test/autoEncrypter.test.js
addaleax: It seems a bit like overkill, yeah. I don’t see why we would *want* to modify it, but we might at some point do that in Compass, and conversely, any code which would be prevented from updating the array would almost certainly be able to get around that by just updating the entire object tree .

bindings/node/test/autoEncrypter.test.js
```diff
@@ -143,6 +143,27 @@ describe('AutoEncrypter', function () {
       });
     });
 
+    it('should decrypt mock data and mark decrypted items if enabled for testing', function (done) {
+      const input = readExtendedJsonToBuffer(`${__dirname}/data/encrypted-document.json`);
```
dariakp, line 147: Should/can we add a test for the js object case so that we cover both top level code paths (object and buffer)? 

It could be also good to cover the new error case (though I don't know if it's too hard to do to be worthwhile).

Does the test document cover both parts of the `if (!decrypted || typeof decrypted !== 'object') return` condition and also the case of the Binary with sub_type other than 6?

dariakp (commented):
Just a couple of nits and a few questions around testing

bindings/node/test/autoEncrypter.test.js
nbbeeken: Very true, sounds like we've got what we need here. Ty!

nbbeeken (approved):

bindings/node/test/autoEncrypter.test.js
addaleax: > Should/can we add a test for the js object case so that we cover both top level code paths (object and buffer)?

Sure, I think that’s fine. Done! :+1: 

> It could be also good to cover the new error case (though I don't know if it's too hard to do to be worthwhile).

Yeah, it would be hard – it’s more of an assertion than a real error case, and triggering it would require mocking out other code to make it appear buggy.

> Does the test document cover both parts of the `if (!decrypted || typeof decrypted !== 'object') return` condition 

Not yet, I think, but it’s easy to add.

> and also the case of the Binary with sub_type other than 6?

Also easy to add, done :+1:

baileympearson (approved):

dariakp (approved):
Thanks!

bindings/node/test/autoEncrypter.test.js
```diff
@@ -143,6 +143,34 @@ describe('AutoEncrypter', function () {
...
+
+        // The same, but with an object containing different data types as the input
+        mc.decrypt({ a: [null, 1, { c: new BSON.Binary('foo', 1) }] }, (err, decrypted) => {
+          if (err) return done(err);
+          expect(decrypted).to.eql({ a: [null, 1, { c: new BSON.Binary('foo', 1) }] });
+          expect(decrypted).to.not.have.property(Symbol.for('@@mdb.decryptedKeys'));
+          done();
+        });
+      });
+    });
```
nbbeeken, line 172: Sorry back with one more ask, can we add another test with an assertion that a false/nullish value for this flag will *not* add the special key & array, I think a small regression protection against accidentally enabling this for users is worth having.

nbbeeken: Perfect, light speed! ty!

nbbeeken (approved):
LGTM I'll merge as soon as we have the node CIs ✅

PR #273, opened 2022-03-19
(Debian packaging) adjust gbp config for new branch name

rcsanchez97:


kevinAlbs (approved):
Thank you for the fix!

PR #277, opened 2022-03-24
MONGOCRYPT-415 add FLE 2 AEAD primitives

kevinAlbs:
# Summary
- Add `_mongocrypt_fle2_do_encryption` and `_mongocrypt_fle2_do_decryption` for primitives described in [AEAD with CTR](https://docs.google.com/document/d/1eCU7R8Kjr-mdyz6eKvhNIDVmhyYQcAaLtTfHeK7a_vE/).
  - I referred to [aead_encryption_fle2_test_vectors.sh](https://github.com/mongodb/mongo/blob/7671b3cca3c5fb1024b8affcafe92c9dfed37280/src/mongo/crypto/scripts/aead_encryption_fle2_test_vectors.sh). 
- Allow empty string as hex string for `_mongocrypt_buffer_copy_from_hex`.
   - This was motivated by tests specifying an [empty hex string](https://github.com/kevinAlbs/libmongocrypt/blob/132091ed10e48ff8aef1a1a1fec5f836d09b3617/test/test-mongocrypt-crypto.c#L523).
- Check for 0 length buffer in `_mongocrypt_buffer_cmp` to silence clang-tidy warnings.
- Add Python scripts to etc to generate FLE 2 AEAD test data.


src/mongocrypt-crypto.c
```diff
@@ -1016,3 +1030,212 @@ _mongocrypt_hmac_sha_256 (_mongocrypt_crypto_t *crypto,
...
+   }
+
+   memset (ciphertext->data, 0, ciphertext->len);
+   *bytes_written = 0;
+
+   /* Declare variable names matching [AEAD with
+    * CTR](https://docs.google.com/document/d/1eCU7R8Kjr-mdyz6eKvhNIDVmhyYQcAaLtTfHeK7a_vE/).
+    */
+   /* M is the input plaintext. */
+   _mongocrypt_buffer_t M = {.data = plaintext->data, .len = plaintext->len};
```
sgolemon, line 1085: Should probably preserve constness on M/Ke/IV/Km/AD since they come from const sources. Decrypt block as well, though inverted selection, of course.

sgolemon: Also, how common is this pattern? Would it be worth making a view API?

```
bool _mongocrypt_buffer_view (_mongocrypt_buffer_t *out, _mongocrypt_buffer_t *in, size_t ofs, size_t len);
bool _mongocrypt_buffer_const_view (const _mongocrypt_buffer_t *out, const _mongocrypt_buffer_t *in, size_t ofs, size_t len);
```

An API would help guard against struct changes and more importantly build bounds checking into the calls to protect against mistakes.  Though I'll be honest, writing that out makes me miss templates and auto. :)

sgolemon (changes requested):

etc/fle2_aead_selftest.py
```diff
@@ -0,0 +1,40 @@
+# payloads_test.py is a self-test of payloads.py.
```
eramongodb, line 1: Should this comment be updated? There is no `payloads_test.py` or `payloads.py` file.

src/mongocrypt-crypto.c
kevinAlbs: Added a `_mongocrypt_buffer_view` to do bounds checking.

I think the constness of the arguments is preserved. `const _mongocrypt_buffer_t` ensures the `data` pointer and `len` are not modified. But it does not prevent modification to the data pointed to by the `data`.

As an aside: I think it would be safer to have the `const _mongocrypt_buffer_t` arguments guarantee that data pointed to by `data` is not modified. One idea is to consider an API similar to the [mstr types](https://github.com/mongodb/libmongocrypt/blob/master/src/mlib/str.h#L14-L86). We could have a separate `_mongocrypt_buffer_view` type containing a `const uint8_t* data` pointer. But I prefer keeping that out of the scope of this PR.

src/mongocrypt-buffer-private.h
```diff
@@ -209,4 +209,12 @@ _mongocrypt_buffer_steal_from_string (_mongocrypt_buffer_t *buf,
 void
 _mongocrypt_buffer_copy_from_uint64_le (_mongocrypt_buffer_t *buf, uint64_t value);
 
+/* _mongocrypt_buffer_view initializes @out as a writable view of the data from
+ * @in. Returns false on error. */
+bool
+_mongocrypt_buffer_view (_mongocrypt_buffer_t *out,
+                         const _mongocrypt_buffer_t *in,
+                         uint32_t offset,
+                         uint32_t len) MONGOCRYPT_WARN_UNUSED_RESULT;
```
eramongodb, line 218:
```suggestion
/* _mongocrypt_buffer_from_view initializes @out as a non-owning buffer to a
 * range of data from @in specified by @offset and @len. Returns false on error.
 */
bool
_mongocrypt_buffer_from_subrange (_mongocrypt_buffer_t *out,
                                  const _mongocrypt_buffer_t *in,
                                  uint32_t offset,
                                  uint32_t len) MONGOCRYPT_WARN_UNUSED_RESULT;
```
I like this addition. However, `_mongocrypt_buffer_t` does not support const correctness, only ownership correctness (there is no read-only "view" mode for `_mongocrypt_buffer_t` objects). I suggest these slight changes to name and description to avoid confusion with the interface of a possible const-correct alternative to `_mongocrypt_buffer_t`.

eramongodb (changes requested):

src/mongocrypt-crypto.c
sgolemon: Right. Too much C++ in my head.

src/mongocrypt-buffer-private.h
kevinAlbs: Good idea. Will do.

src/mongocrypt-crypto-private.h
```diff
@@ -44,9 +44,19 @@ typedef struct {
 uint32_t
 _mongocrypt_calculate_ciphertext_len (uint32_t plaintext_len);
 
+/* _mongocrypt_fle2aead_calculate_ciphertext_len returns the required length of
+ * the ciphertext for _mongocrypt_fle2aead_do_encryption. */
+uint32_t
+_mongocrypt_fle2aead_calculate_ciphertext_len (uint32_t plaintext_len);
```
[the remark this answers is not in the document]
kevinAlbs: I prefer to keep this as-is, but I am open to further discussion.

I consider `_mongocrypt_fle2aead_do_encryption` and `_mongocrypt_fle2aead_do_decryption` to be specific to FLE 2 since they follow the construction from [AEAD with CTR](https://docs.google.com/document/d/1eCU7R8Kjr-mdyz6eKvhNIDVmhyYQcAaLtTfHeK7a_vE/).

Using `fle2aead` is consistent with these functions in the MongoDB server: https://github.com/mongodb/mongo/blob/cf22bb8204da47c2d3696aabb2574077a66dcd66/src/mongo/crypto/aead_encryption.h#L99

AFAIK there is no opportunity for reuse currently.

IMO having `fle2` as part of the symbol makes it easier to track what applies to FLE 1 and FLE 2.

eramongodb (approved):

PR #279, opened 2022-03-28
Download universal MacOS shared library

jyemin:
This ensures that the M1 processor is supported

MONGOCRYPT-418

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -429,14 +429,18 @@ tasks:
       name: prep-c-driver-source
     - variant: rhel-62-64-bit
       name: build-and-test-java
+    - variant: rhel-62-64-bit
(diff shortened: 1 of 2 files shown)


kevinAlbs (commented):
Does the [publish-snapshot task](https://github.com/mongodb/libmongocrypt/blob/1687a19667d546fe1b6be22f13fcae4463af3f42/.evergreen/config.yml#L436-L439) need to be updated?

It currently depends on the `macos_x86_64` variant's `build-and-test-java` task.

Perhaps it should depend on the `macos` variant's `build-and-test-and-upload` task.

jyemin:
Update publish-snapshot dependencies to add more `build-and-test-and-upload` tasks

PR #281, opened 2022-03-30
MONGOCRYPT-413 always use HEAD reference for generating Debian tarball

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/6244c7023066157a6c01e417/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

We apparently had this issue in the C driver about a year ago and somehow I forgot about making the fix here as well.


kevinAlbs (approved):
LGTM!

PR #284, opened 2022-04-05
MONGOCRYPT-403 Decrypt `FLE2IndexedEqualityEncryptedValue` in automatic decryption.

kevinAlbs:
# Summary
- Decrypt `FLE2IndexedEqualityEncryptedValue` in automatic decryption.
- Add `_mongocrypt_key_broker_restart` to transition the key broker from `KB_DONE` to `KB_REQUESTING`
- Fix `_assert_match_bson` to abort on assertion error. Previously it [only printed](https://github.com/mongodb/libmongocrypt/blob/b4beca89c4071cc0c67c4ef6c1766d6ee03b6acb/test/test-mongocrypt-assert-match-bson.c#L1068-L1075).
- Add `mc-fle-blob-subtype-private.h` with an enum for the different fle_blob_subtype values.
- Add `--key_material` option to `csfle` CLI tool.

# Background & Motivation
Decrypting `FLE2IndexedEqualityEncryptedValue` requires two passes to fetch keys. One to get the outer `S_Key`. `S_Key` is used to decrypt a nested ciphertext containing the `K_KeyId`. `K_Key` is used to decrypt the `ClientEncryptedValue`. That motivated the addition of `_mongocrypt_key_broker_restart`.

Adding `--key_material` to the `csfle` tool made it easier to generate test data from a specific Data Encryption Key.

# Notes

This PR does not enable decrypting in explicit encryption.


vector-of-bool (approved):
LGTM! Nothing stands out here.

PR #286, opened 2022-04-06
MONGOCRYPT-397 Process Placeholder into InsertUpdatePayload

sgolemon:
# Summary
- Adds type `mc_FLE2EncryptionPlaceholder_t` with init/parse/cleanup methods for marshalling from a BSON payload containing subsubtype 3 (FLE2EncryptionPlaceholder).
- Adds type `mc_FLE2InsertUpdatePayload_t` with init/parse/serialize/cleanup methods for producing BSON payloads containing subsubtype 4 (FLE2InsertUpdatePayload)
- Performs token generation and value transformation for insert/update queries.
  - Capture subsubtype 3 (FLE2InsertUpdatePlaceholder) during collect_key_from_marking
  - Process markings into FLE2InsertUpdatePayload during _mongocrypt_marking_to_ciphertext

# Background & Motivation
This is part of the InsertFindCriticalPath necessary for transforming insert/update operations into encrypted payloads.


sgolemon:
This needs a little more testing yet.  Is it possible to *just* override the crypto_random hook and not the others? Would allow for generating deterministic output for testing). If not, I could apply a reverse of the IV which *would* get to a deterministic payload (because that's how CTR works), but that feels like a lot of effort.  Certainly, once we have decryption ready, we can work out a full loop.

Mostly putting this up to get feedback on style.  I ended up making a number of judgement calls which may/may-not align with what you want for this project long term, so let please be brutally honest. :D

test/fle2/insert/cmd.json
```diff
@@ -0,0 +1,12 @@
+{
```
kevinAlbs, line 1: Please move the test data files into the `test/data` directory?

kevinAlbs, line 1: Can you move the test data files into the directory `test/data/fle2`?

src/mongocrypt-marking.c
```diff
@@ -150,40 +222,371 @@ _mongocrypt_marking_cleanup (_mongocrypt_marking_t *marking)
...
+
+   BSON_ASSERT (
+      _mongocrypt_buffer_from_subrange (&iv, out, 0, MONGOCRYPT_IV_LEN));
+   if (!_mongocrypt_random (crypto, &iv, MONGOCRYPT_IV_LEN, status)) {
+      return false;
+   }
+
+   BSON_ASSERT (_mongocrypt_buffer_from_subrange (
+      &ciphertext, out, MONGOCRYPT_IV_LEN, out->len - MONGOCRYPT_IV_LEN));
+   if (!_native_crypto_aes_256_ctr_encrypt (
```
kevinAlbs, line 370: I suggest using `_mongocrypt_fle2_do_encryption`. `_mongocrypt_fle2_do_encryption` will eventually use the hook for AES-256-CTR.

kevinAlbs, line 370: I suggest using `_mongocrypt_fle2_do_encryption`. That will eventually use the crypto hooks.

kevinAlbs (commented):
Nice work! Style looks good overall. I included an optional comment to consider separating `_mongocrypt_marking_t` into separate structs.

I like the addition of overriding the `_mongocrypt_tester_t` test file paths.

src/mongocrypt-marking.c
sgolemon: Yep. This was done before `_mongocrypt_fle2_do_encryption` was available. :D

test/fle2/insert/cmd.json
sgolemon: Done

sgolemon (commented):
Rebased (and thus force pushed) to get your latest utility functions available, then applied your suggestions.  Big edits, so might as well start review from scratch at this point. ;)

kevinAlbs:
Here is an experimental test with a fixed IV to compare against an FLE2InsertUpdatePayload captured from server unit tests:
https://github.com/mongodb/libmongocrypt/commit/6924633e797dc090fc8ef31ec33d853d3da5aee2#diff-70d771e7c35927baab41e7a4472474753d07fbb2cccf73640626663ad3a3b771R2103

It is verbose, but it may serve as a way to validate.

I captured the IV by dumping the FLE2InsertUpdatePayload captured from server unit tests here: https://github.com/mongodb/libmongocrypt/commit/c64da14fde1f0f71edbae0c70261a196a7c102df#diff-60a4dbebd8f2ee9ebf592f419a5f28bd9b590b9e8716250ec35ad045868127c0R1

sgolemon (commented):
Those hooks are exactly what I was looking for in a test (using real data and real encryption outputs).  I pulled the passthroughs and the error functions out to a separate unit to make them more reusable, and I put the JSON into the test files instead of hardcoded, but otherwise pretty much took your test as-is,

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -1953,6 +1955,177 @@ _test_encrypt_with_bypassqueryanalysis (_mongocrypt_tester_t *tester)
    mongocrypt_destroy (crypt);
 }
 
+
+static void
+_test_FLE2InsertUpdatePayload_parse (_mongocrypt_tester_t *tester)
```
kevinAlbs, line 1960:
```suggestion
_test_FLE2EncryptionPlaceholder_parse (_mongocrypt_tester_t *tester)
```

src/mongocrypt-marking.c
```diff
@@ -148,42 +175,330 @@ _mongocrypt_marking_init (_mongocrypt_marking_t *marking)
...
+   // p := EncryptCTR(ECOCToken, ESCDerivedFromDataTokenAndCounter ||
+   // ECCDerivedFromDataTokenAndCounter)
+   {
+      _mongocrypt_buffer_t tokens[] = {payload.escDerivedToken,
+                                       payload.eccDerivedToken};
+      _mongocrypt_buffer_t p;
+      _mongocrypt_buffer_concat (&p, tokens, 2);
+      mc_ECOCToken_t *ecocToken =
+         mc_ECOCToken_new (crypto, collectionsLevel1Token, status);
+      if (!ecocToken) {
```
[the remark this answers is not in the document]
kevinAlbs: Still needs to be addressed.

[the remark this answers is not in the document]
kevinAlbs: Oops, nevermind. ecocToken does not need to be freed. Resolving.

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -1953,6 +1955,177 @@ _test_encrypt_with_bypassqueryanalysis (_mongocrypt_tester_t *tester)
...
+
+   return true;
+}
+
+
+static void
+_test_encrypt_fle2_insert_payload (_mongocrypt_tester_t *tester)
+{
+   mongocrypt_t *crypt;
+   int random_call_count;
```
kevinAlbs, line 2042:
```suggestion
   int random_call_count = 0;
```

kevinAlbs (approved):
Fantastic work! LGTM once tests are passing. I think the "expected random to be called exactly twice, got called a third time" error may be resolved by initializing `random_call_count`.

sgolemon:
@kevinAlbs Can you help me understand the remaining errors?    Some variants say there's a SIGABORT, but provide no further information about what where or why.  Other variants are failing in KMS code that I haven't touched.

PR #295, opened 2022-04-12
MONGOCRYPT-406 add new context to support ClientEncryption.Compact

kevinAlbs:
# Summary
- Add compact context for drivers to implement the ClientEncryption::Compact helper. It is initialized with `mongocrypt_ctx_compact_init`.
- Add `mc_EncryptedFieldConfig_t` to store a parsed EncryptedFieldConfig.

# Background & Motivation
The ClientEncryption::Compact helper is defined in WRITING-8932. It runs the server command [compactStructuredEncryptionData](https://github.com/markbenvenuto/mongo-enterprise-modules/blob/e9954e024f7bfd214c779e7f937776fe898072d4/fle_protocol.md#reference-compactstructuredencryptiondata).

I tested generating a sample compact command with a key and encrypted field config printed from [server tests](https://gist.github.com/kevinAlbs/4ad9b7ad511a02899bbab99b521c9de5) to ensure the output `ECOCToken`s matched. I did not add those to unit tests. The keys in the server test are ephemeral. So there is no output to reference.


src/mc-efc.c
```diff
@@ -0,0 +1,121 @@
...
+              bson_t *field,
+              mongocrypt_status_t *status)
+{
+   bson_iter_t field_iter;
+   if (!bson_iter_init_find (&field_iter, field, "keyId")) {
+      CLIENT_ERR ("unable to find 'keyId' in 'field' document");
+      return false;
+   }
+   if (!BSON_ITER_HOLDS_BINARY (&field_iter)) {
+      CLIENT_ERR ("expected 'fields[].keyId' to be type binary, got: %d",
```
eramongodb, line 34: Recommend adjusting error messages to either include index of `fields` array (e.g. `expected 'fields[0].keyid' to be type binary`) or avoid using subscript notation without an index (e.g. `expected 'keyId' to be type binary`) as `[]` alone looks like it is missing an index (format error). Suggestion applies to all other instances of `'fields[]'` in error messages in this file.

eramongodb (approved):
Minor feedback, but otherwise LGTM.

src/mc-efc.c
kevinAlbs: Good point. Removed the subscript notation.

PR #296, opened 2022-04-15
MONGOCRYPT-398 Process FLE2EncryptionPlaceholder into FLE2FindEqualityPayload

sgolemon:
# Summary
- Adds type `mc_FLE2FindEqualityPayload_t` with init/parse/serialize/cleanup methods for producing BSON payloads containing subsubtype 5 (FLE2FindEqualityPayload)
- Pull generation of '`d`', '`s`', and '`c`' fields out to a helper method: `_mongocrypt_fle2_placeholder_common`
  - Use for processing both insertupdate and findequality payloads during ciphertext generation
- Refactor `_test_encrypt_fle2_insert_payload` into `_test_encrypt_fle2_encryption_placeholder` to allow it to perform the same steps, with different data, for both insert and find test queries. 

# Background & Motivation
This is part of the InsertFindCriticalPath necessary for transforming find operations into encrypted payloads.


test/data/fle2-find/key-document.json
```diff
@@ -0,0 +1,33 @@
+{
+    "status": {
+        "$numberInt": "1"
+    },
+    "_id": {
+        "$binary": {
+            "base64": "YWFhYWFhYWFhYWFhYWFhYQ==",
```
kevinAlbs, line 7: Please removing key-document.json. The test uses the keys from `test/data/keys` directory.

test/data/fle2-find/encrypted-payload.json
```diff
@@ -0,0 +1,15 @@
+{
+   "documents": [
+      {
+         "_id": 1,
+         "ssn": {
+            "$binary": {
+               "base64": "BYkAAAAFZAAgAAAAAHb62aV7+mqmaGcotPLdG3KP7S8diFwWMLM/5rYtqLrEBXMAIAAAAAAVJ6OWHRv3OtCozHpt3ZzfBhaxZirLv3B+G8PuaaO4EgVjACAAAAAAsZXWOWA+UiCBbrJNB6bHflB/cn7pWSvwWN2jw4FPeIUSY20AAAAAAAAAAAAA",
```
kevinAlbs, line 7: I expect this to equal the base64 in `docWithTransformedPlaceholders` from the first kFind case in the [captured payloads](https://gist.github.com/kevinAlbs/cba611fe0d120b3f67c6bee3195d4ce6):

```
roundTripTest ... begin
... operation = Operation::kFind
... algorithm = FLE2AlgorithmInt::kEquality
... inputDoc = {"plainText":"sample","encrypted":"value123"}
... docWithPlaceholder (payload 3) = {"plainText":"sample","encrypted":{"$binary":{"base64":"A2EAAAAQdAACAAAAEGEAAgAAAAVraQAQAAAABBI0VngSNJh2EjQSNFZ4kBIFa3UAEAAAAASrze+rEjSYdhI0EjRWeJASAnYACQAAAHZhbHVlMTIzABJjbQAAAAAAAAAAAAA=","subType":"6"}}}
... docWithTransformedPlaceholders (payload 4 or 5 or 6) = {"plainText":"sample","encrypted":{"$binary":{"base64":"BYkAAAAFZAAgAAAAAPtVteJQAlgb2YMa/+7YWH00sbQPyt7L6Rb8OwBdMmL2BXMAIAAAAAAd44hgVKnEnTFlwNVC14oyc9OZOTspeymusqkRQj57nAVjACAAAAAA19X9v9NlWidu/wR5/C/7WUV54DfL5CkNmT5WYrhxdDcSY20AAAAAAAAAAAAA","subType":"6"}}}
... finalDoc (payload 7 or 6) = {"plainText":"sample","encrypted":{"$binary":{"base64":"BYkAAAAFZAAgAAAAAPtVteJQAlgb2YMa/+7YWH00sbQPyt7L6Rb8OwBdMmL2BXMAIAAAAAAd44hgVKnEnTFlwNVC14oyc9OZOTspeymusqkRQj57nAVjACAAAAAA19X9v9NlWidu/wR5/C/7WUV54DfL5CkNmT5WYrhxdDcSY20AAAAAAAAAAAAA","subType":"6"}},"__safeContent__":[]}
roundTripTest ... end
```

Why does this differ?

sgolemon: 1. I was using a different find command (ssn: "123-45-6789") and didn't realize we have expected outputs for everything in here.  I've updated the test to use one of these (the one using (value: 123456, specifically, since there's not a clean JSON way of expressing a regex pattern).

2. Looks like there was a legit bug because I misread edcDerivedDataToken as being the same as edcDerivedDataTokenAndCounter.  Fixed in latest commit, matches the expected output now.

test/data/fle2-find/key-document.json
sgolemon: Gone, and removed the copy from 397 as well.

kevinAlbs (approved):
Great, work. LGTM!

PR #299, opened 2022-04-18
update release URL in build.gradle.kts

kevinAlbs:
update release process to note updating this URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Do the following when releasing:
    - Create a branch named `rx.y`.
    - Update the [libmongocrypt-release](https://evergreen.mongodb.com/projects##libmongocrypt-release) Evergreen project to set `Branch Name` to `rx.y`.
    - Update the Linux distribution package installation instructions in the below sections to refer to the new version x.y.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -128,6 +128,7 @@ Do the following when releasing:
    - Create a branch named `rx.y`.
    - Update the [libmongocrypt-release](https://evergreen.mongodb.com/projects##libmongocrypt-release) Evergreen project to set `Branch Name` to `rx.y`.
    - Update the Linux distribution package installation instructions in the below sections to refer to the new version x.y.
+   - In the Java binding build.gradle.kts, update the `LibMongoCryptS3Data` function. Replace the minor version `rx.y` in the URL with the new branch.
```
jyemin, line 131: This isn't going to work.  The build script has a check to ensure that the git clone is not dirty, so any manual change will cause that check to fail.  We need to add the ability to parameterize this value at build time via system property,  similar to how we do with the gitRevision.

rozza: The tagging (git commit and git tag) is occurring after this step.

rozza (approved):
LGTM

rozza:
I'll look to simplify the build and use libmongocrypt-all.tar.gz (MONGOCRYPT-324) in the next week or so.

kevinAlbs:
Thank you for the reviews.

PR #310, opened 2022-04-20
test: loosen test error assertion for fails to autospawn

nbbeeken:
#### What is changing?

Loosening the test error message checking to instead assert we raise a subclass of our own error classes. The test is failing due to DNS configuration of the system. We don't want to add the DNS case to the filter that generates the "helpful" message the test refers to since it wouldn't be helpful to be told to ensure mongocryptd is in the path when the real issue is an incorrect mongocryptdURI.

It is blocking the adoption of new nodejs versions in the CI for the driver.

- [x] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>`


addaleax (approved):
@nbbeeken Does this need to be in draft? :)

nbbeeken:
It does not, thanks! 😅

dariakp (approved):

PR #311, opened 2022-04-21
fix appending of encryptionInformation for FLE 2 commands

kevinAlbs:
# Summary

- Use exact BSON match for test files in `_test_encrypt_fle2_encryption_placeholder`.
- Always append `encryptionInformation` after encryption with FLE 2.
- Remove `encryptionInformation` from mongocryptd reply.

# Background & Motivation

`encryptionInformation` is expected to be appended to outgoing encrypted commands for FLE 2. `encryptionInformation` was incorrectly only appended when the input command did not contain values to encrypt.

Fixing that behavior necessitated removing the `encryptionInformation` from the mongocryptd reply. Otherwise two `encryptionInformation` fields could be present. mongocryptd echos back the `encryptionInformation` it is sent. Here is a sample reply:
```js
{
  hasEncryptionPlaceholders: false,
  schemaRequiresEncryption: true,
  result: {
    insert: 'db.test',
    encryptionInformation: {
      type: 1,
      schema: {
        'db.test': {
          escCollection: 'db.test.esc',
          eccCollection: 'db.test.ecc',
          ecocCollection: 'db.test.ecoc',
          fields: [
            {
              path: 'a',
              keyId: UUID("ace57ba5-50db-4002-b50e-c360bec9614b"),
              bsonType: 'string'
            }
          ]
        }
      }
    },
    documents: [ { foo: 'bar' } ]
  },
  ok: 1
}
```

MONGOCRYPT-405 requires modifying the outgoing `encryptionInformation`. This PR proposes always removing the `encryptionInformation` from the mongocryptd reply, rather than attempting to reuse it.

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -643,11 +643,6 @@ _fle2_finalize (mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out)
       /* Append 'encryptionInformation' to the original command. */
       bson_init (&converted);
       bson_copy_to (&original_cmd_bson, &converted);
-      if (!_fle2_append_encryptionInformation (
-             &converted, ectx->ns, &encrypted_field_config_bson, ctx->status)) {
-         bson_destroy (&converted);
-         return _mongocrypt_ctx_fail (ctx);
-      }
    } else {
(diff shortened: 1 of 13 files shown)


vector-of-bool (approved):
LGTM!

PR #312, opened 2022-04-21
MONGOCRYPT-385 Add support for AES-CTR on Mac

brushless-glitch:
This adds support for AES-CTR with CommonCrypto.


kevinAlbs (approved):
LGTM!

PR #313, opened 2022-04-22
Do not skip NEED_MONGO_COLLINFO state with bypass_query_analysis

kevinAlbs:
# Summary

- Do not skip the `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` state when the `bypass_query_analysis` option is true.

# Background & Motivation

If the target collection has an associated `encrypted_field_config`, it must be appended to the outgoing command with `_fle2_append_encryptionInformation`. The server uses the `encrypted_field_config` to process FLE 2 payloads.

The `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` state requests the driver to run `listCollections`. It is used to obtain a server-side `$jsonSchema` (for FLE 1 `schema`) or a server-side `encryptedFields` (for FLE 2 `encrypted_field_config`).

### DIFF
--- a/test/test-mongocrypt-ctx-encrypt.c
+++ b/test/test-mongocrypt-ctx-encrypt.c
@@ -1921,39 +1921,86 @@ _test_encrypt_with_bypassqueryanalysis (_mongocrypt_tester_t *tester)
    mongocrypt_t *crypt;
    mongocrypt_ctx_t *ctx;
 
-   crypt = mongocrypt_new ();
-   ASSERT_OK (
-      mongocrypt_setopt_kms_providers (
-         crypt,
-         TEST_BSON (
-            "{'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'bar'}}")),
-      crypt);
-   ASSERT_OK (mongocrypt_setopt_encrypted_field_config_map (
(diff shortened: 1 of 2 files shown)


test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -1921,39 +1921,86 @@ _test_encrypt_with_bypassqueryanalysis (_mongocrypt_tester_t *tester)
...
+      ASSERT_OK (
+         mongocrypt_setopt_kms_providers (
+            crypt,
+            TEST_BSON (
+               "{'aws': {'accessKeyId': 'foo', 'secretAccessKey': 'bar'}}")),
+         crypt);
+      ASSERT_OK (mongocrypt_setopt_encrypted_field_config_map (
+                    crypt, TEST_BSON ("{'db.coll': {'foo': 'bar'}}")),
+                 crypt);
+      mongocrypt_setopt_bypass_query_analysis (crypt);
```
eramongodb, line 1936: Did this call to `mongocrypt_setopt_bypass_query_analysis()` have any effect prior to the addition of the `ctx->crypt->opts.bypass_query_analysis` check in `mongocrypt_ctx_encrypt_init()`?

eramongodb (approved):
Other than a clarification question, LGTM.

PR #314, opened 2022-04-22
MONGOCRYPT-405 Append `deleteTokens` to required commands

kevinAlbs:
# Summary

- Append `deleteTokens` to the outgoing `encryptionConfig` for required commands.
- Check if `marked_cmd` is empty before using in `_fle2_finalize`.

# Background & Motivation

Test data was obtained from [server tests](https://gist.github.com/kevinAlbs/3ccc30754755dcf543f14f83340b7013).

Confirmed [in slack](https://mongodb.slack.com/archives/C0316KB5VM4/p1650551700482719) that `deleteTokens` must only be applied to `update`, `delete`, and `findAndModify` commands.

Can no longer make the assumption that `!nothing_to_do` means that `marked_cmd` is non-empty. Keys may have been requested for `deleteTokens`.


vector-of-bool (approved):
LGTM! Nothing stands out to me.

PR #315, opened 2022-04-23
MONGOCRYPT-409 + MONGOCRYPT-407 Add support for FLE 2 explicit encryption

kevinAlbs:
# Summary

- Add support for FLE 2 explicit encryption.
- Add new options for FLE 2 explicit encryption:
    - `mongocrypt_ctx_setopt_index_type`
    - `mongocrypt_ctx_setopt_contention_factor`
    - `mongocrypt_ctx_setopt_index_key_id`
    - `mongocrypt_ctx_setopt_query_type`

# Background and motivation

## index_key_id
FLE 2 fields have two keys: the `IndexKey` (aka `S_Key`) and `UserKey` (aka `K_Key`). [Reference: Keys and Tokens](https://github.com/markbenvenuto/mongo-enterprise-modules/blob/e9954e024f7bfd214c779e7f937776fe898072d4/fle_protocol.md#reference-keys-and-tokens) has a brief description:

> S_KeyId or IndexKey is a per-field key material. It is 96 bytes since that is the same size as the original FLE 1 value. From this index key, a normal of tokens are derived via HMAC as described below.
> K_KeyId is either chosen by the user or is S_KeyId/IndexKey.
> Since it is possible for IndexKey == UserKey, we choose the key and HMAC components assuming they are shared.

This answer over [slack](https://mongodb.slack.com/archives/C0316KB5VM4/p1649427585223719) motivated an additional `mongocrypt_ctx_setopt_index_key_id` option for explicit encryption:
> For implicit encryption, FLE2EncryptionPlaceholder sets userKeyId = indexKeyId.
> Only explicit encryption can set userKeyId to something that is not indexKeyId.

An `IndexKey` option is now suggested in the [scope](https://docs.google.com/document/d/1duvPDqbFFEvSjTWX_8Vp-u_IjGZGdV3ZyrzCjrjbn2E/edit).

## query_type

The query type option is to support the proposed `GenerateEquality` driver helper. MONGOCRYPT-407 proposed a separate context. But the behavior would be nearly identical to explicit encryption. Assuming `GenerateEquality` ends up in the specification, drivers can use the explicit encryption context to implement it.


test/test-mongocrypt-ctx-setopt.c
```diff
@@ -697,6 +697,119 @@ _test_setopt_for_explicit_encrypt (_mongocrypt_tester_t *tester)
    ALGORITHM_FAILS ("bad-algo", -1, "unsupported algorithm");
    EX_ENCRYPT_INIT_FAILS (bson, "unsupported algorithm");
 
+   /* It is an error to set the FLE 1 keyAltName option with any of the FLE 2
+    * options (index_type, index_key_id, or contention_factor). */
```
eramongodb, line 701:
```suggestion
    * options (index_type, index_key_id, contention_factor, or query_type). */
```

test/test-mongocrypt-ctx-setopt.c
```diff
@@ -697,6 +697,119 @@ _test_setopt_for_explicit_encrypt (_mongocrypt_tester_t *tester)
...
+   {
+      REFRESH
```
eramongodb, line 703: Unrelated to PR, but `REFRESH` macro (and others in this file) should probably be adjusted at some point to consistently require a terminating semicolon:
```diff
- #define REFRESH                  \
-    mongocrypt_ctx_destroy (ctx); \
-    ctx = mongocrypt_ctx_new (crypt);
+ #define REFRESH                         \
+    do {                                 \
+       mongocrypt_ctx_destroy (ctx);     \
+       ctx = mongocrypt_ctx_new (crypt); \
+    } while (0)
```

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -2163,6 +2163,373 @@ _test_encrypt_fle2_unindexed_encrypted_payload (_mongocrypt_tester_t *tester)
...
+         ASSERT_OK (mongocrypt_ctx_mongo_feed (
+                       ctx, TEST_FILE ("./test/data/keys/"
+                                       "12345678123498761234123456789012-local-"
+                                       "document.json")),
+                    ctx);
+         ASSERT_OK (mongocrypt_ctx_mongo_feed (
+                       ctx, TEST_FILE ("./test/data/keys/"
+                                       "ABCDEFAB123498761234123456789012-local-"
+                                       "document.json")),
+                    ctx);
```
eramongodb, line 2247:
```suggestion
         ASSERT_OK (mongocrypt_ctx_mongo_feed (
                       ctx,
                       TEST_FILE ("./test/data/keys/"
                                  "12345678123498761234123456789012-local-"
                                  "document.json")),
                    ctx);
         ASSERT_OK (mongocrypt_ctx_mongo_feed (
                       ctx,
                       TEST_FILE ("./test/data/keys/"
                                  "ABCDEFAB123498761234123456789012-local-"
                                  "document.json")),
                    ctx);
```
Clang-Format. Also applies to other instances of loading `./test/data/keys/...` below.

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -2163,6 +2163,373 @@ _test_encrypt_fle2_unindexed_encrypted_payload (_mongocrypt_tester_t *tester)
...
+                       "'subType': '06' } }}"),
+            got);
+         mongocrypt_binary_destroy (got);
+      }
+
+      mongocrypt_ctx_destroy (ctx);
+      mongocrypt_destroy (crypt);
+   }
+   /* Test Indexed with non-zero ContentionFactor. */
+   {
```
eramongodb, line 2339:
```suggestion
   }

   /* Test Indexed with non-zero ContentionFactor. */
   {
```
Consistent spacing between test cases.

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -2163,6 +2163,373 @@ _test_encrypt_fle2_unindexed_encrypted_payload (_mongocrypt_tester_t *tester)
...
+                       "'subType': '06' } }}"),
+            got);
+         mongocrypt_binary_destroy (got);
+      }
+
+      mongocrypt_ctx_destroy (ctx);
+      mongocrypt_destroy (crypt);
+   }
+   /* Test that omitted index_key_id defaults to using user_key_id. */
+   {
```
eramongodb, line 2408:
```suggestion
   }

   /* Test that omitted index_key_id defaults to using user_key_id. */
   {
```
Consistent spacing between test cases.

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -2163,6 +2163,373 @@ _test_encrypt_fle2_unindexed_encrypted_payload (_mongocrypt_tester_t *tester)
...
+                                     "ABCDEFAB123498761234123456789012");
+   _mongocrypt_buffer_copy_from_hex (&index_key_id,
+                                     "12345678123498761234123456789012");
+
+   /* Test Unindexed. */
+   {
+#define RNG_DATA \
+   "\x4d\x06\x95\x64\xf5\xa0\x5e\x9e\x35\x23\xb9\x8f\x57\x5a\xcb\x15"
+      _test_rng_data_source source = {
+         .buf = {.data = (uint8_t *) RNG_DATA, .len = sizeof (RNG_DATA)}};
```
eramongodb, line 2217: `sizeof (RNG_DATA)` includes null terminator. Is this acceptable? This question also applies to other instances of `sizeof (RNG_DATA)` below.

eramongodb (approved):
Some suggestions and comments, but otherwise, LGTM.

test/test-mongocrypt-ctx-encrypt.c
kevinAlbs: Good eye. Preferably no, since the test asserts if more random data was requested than expected. Updated to subtract one.

test/test-mongocrypt-ctx-setopt.c
kevinAlbs: Good point. That is small enough to do here. Done.

PR #317, opened 2022-04-26
Fix: Getting working directory adds a NUL char

vector-of-bool:
This is a source of failure with loading csfle from a relative path: The wchar-narrowing function accidentally added a NUL char.


src/mlib/str.h
```diff
@@ -825,13 +828,15 @@ mstr_win32_narrow (const wchar_t *wstring)
...
    }
-   mstr_mut ret = mstr_new ((size_t) length);
+   // Allocate a new string, not including the null terminator
+   mstr_mut ret = mstr_new ((size_t) (length - 1));
    int got_len = WideCharToMultiByte (CP_UTF8,
-                                      WC_ERR_INVALID_CHARS,
+                                      wcflags,
                                       wstring,
-                                      -1,
+                                      length,
```
kevinAlbs, line 836:
```suggestion
                                      wcslen (wstring),
```

IIUC `length` is the required size of `ret.data`. This argument, `cchWideChar` is:

> Size, in characters, of the string indicated by lpWideCharStr.

I think if the `wstring` contains a character that encodes to multi-byte UTF-8, this may be an incorrect length for `wstring`.

If wstring is null terminated.

kevinAlbs (approved):
Looks good with a possible fix.

src/mlib/str.h
vector-of-bool: I'll add a proper test using some utf16 strings to make sure.

PR #320, opened 2022-04-26
MONGOCRYPT-423 Only send EncryptionInformation if needed

kevinAlbs:
# Summary
- Do not send encryptionInformation under these conditions:
   - The outgoing command has no `FLE2FindEncryptedPayload`, `FLE2InsertUpdatePayload`.
   - The command is: find, aggregate, distinct, count, or insert.

# Background & Motivation

See MONGOCRYPT-423 for the rationale.

`encryptionInformation` is required for the server to process `FLE2FindEncryptedPayload`, `FLE2InsertUpdatePayload`. `encryptionInformation` is required for the server to process update and delete operations on indexed fields.


src/mongocrypt-ctx-encrypt.c
```diff
@@ -938,6 +938,75 @@ generate_delete_tokens (_mongocrypt_crypto_t *crypto,
...
+}
+
+static bool
+must_omit_encryptionInformation (const char *command_name,
+                                 const bson_t *command,
+                                 bool *must_omit /* out */,
+                                 mongocrypt_status_t *status)
+{
+   const char *eligible_commands[] = {
+      "find", "aggregate", "distinct", "count", "insert"};
```
kevinAlbs, line 973: @markbenvenuto are these conditions for omitting `encryptionInformation` correct?

1. The command is `find`, `aggregate`, `distinct`, `count`, or `insert.
2. The command has no `FLE2FindEncryptedPayload` or `FLE2InsertUpdatePayload`.

src/mongocrypt-ctx-encrypt.c
```diff
@@ -938,6 +938,75 @@ generate_delete_tokens (_mongocrypt_crypto_t *crypto,
...
+   if (in->data[0] == MC_SUBTYPE_FLE2FindEqualityPayload) {
+      *out = true;
+      return true;
+   }
+
+   return true;
+}
+
+static bool
+must_omit_encryptionInformation (const char *command_name,
```
vector-of-bool, line 967: This naming spooks me, because the return value is not actually the result value, which are both `bool`, yet the name implies the result is related to what it computes. I'm not sure what a better name might be, though. Maybe return a `struct` with two named values instead? Or maybe an enum of "yes, no, error" to be checked at the call site?

vector-of-bool (commented):
Looks okay, but with one comment about bool shuffling

src/mongocrypt-ctx-encrypt.c
kevinAlbs: I like the idea of returning a struct. Done.

vector-of-bool (approved):

src/mongocrypt-ctx-encrypt.c
kevinAlbs: Confirmed over slack.

PR #324, opened 2022-04-28
feat(NODE-4218): add aes256-ctr support

addaleax:
##### feat(NODE-4218): add aes256-ctr support

This is required for FLE2 support. Diff best viewed with whitespace changes hidden because it moves a function to an outer scope :)

##### fix: pass promoteValues: false where encryptedFields is processed

(drive by, can be a separate PR if necessary)

This is required because `encryptedFields.fields[].queries[].contention`
must be a Long in order for the server/mongocryptd/the csfle
shared library to be able to process it.

### DIFF
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -252,6 +252,8 @@ module.exports = function (modules) {
       const stateMachine = new StateMachine({
         bson,
         ...options,
+        promoteValues: false,
+        promoteLongs: false,
         proxyOptions: this._proxyOptions,
         tlsOptions: this._tlsOptions
       });
--- a/bindings/node/lib/cryptoCallbacks.js
+++ b/bindings/node/lib/cryptoCallbacks.js
@@ -1,33 +1,25 @@
 'use strict';
 const crypto = require('crypto');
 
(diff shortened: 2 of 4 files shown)


durran (approved):
I would have liked a test added to verify the the encrypt/decrypt callbacks were called but noticed all the cryptoCallbacks tests are all skipped and I think it's another ticket to look into why those are skipped and what we need to do to get them running again... So I'd say that's out of scope here then. Nice refactor to follow the Hmac pattern with `makeAES256Hook`. LGTM

durran:
Based on conversations with @addaleax and @kevinAlbs it wouldn't currently be possible to actually test this in its current state, as we need to implement the FLE2 explicit encryption API or some binding into `mongocrypt_ctx_setopt_index_type` in order to add the tests for this. That spec is not yet merged so a second PR will be opened with tests that also cover the AES-256-CTR cases for this PR. So I'm sticking with the approval to get this is now and we can verify on the next PR.

addaleax:
#327 adds test coverage here, including one tests that explicitly asserts the exact FLE2 AES256-CTR-encrypted result. But yeah, that spec isn’t merged yet, so I think a separate PR is fine for now

dariakp (approved):

PR #325, opened 2022-04-29
MONGOCRYPT-424 Process compact in auto encryption

kevinAlbs:
# Summary

- Process `compactStructuredEncryptionData` command as part of encryption context.
- Store `bypass_query_analysis` on `_mongocrypt_ctx_encrypt_t`.

# Background & Motivation

The motivation is to remove the proposed `ClientEncryption.Compact` helper proposed in the [DBX Scope: FLE 2.0 Support](https://docs.google.com/document/d/1duvPDqbFFEvSjTWX_8Vp-u_IjGZGdV3ZyrzCjrjbn2E/edit). Processing a `compactStructuredEncryptionData` command can be done in automatic encryption. Drivers do not need to wrap a new context, or expose new API.

compactionTokens for unindexed fields (where `mc_EncryptedField_t::has_queries` is false) are required. Integration tests receive this error when compactionTokens are omitted for unindexed fields:
> (Location6346806) Compaction tokens object is missing compaction token for the encrypted path 'encryptedUnindexed'

The `compactStructuredEncryptionData` does not need `mongocryptd` / `csfle`. Sending to `mongocryptd` results in an error. Storing `bypass_query_analysis` on `_mongocrypt_ctx_encrypt_t` enables setting `bypass_query_analysis` to true for only the lifetime of the context.


src/mongocrypt-ctx-encrypt.c
```diff
@@ -357,6 +357,43 @@ _fle2_collect_keys_for_deleteTokens (mongocrypt_ctx_t *ctx)
...
+ * compactionTokens. compactionTokens is only applicable to FLE 2. */
+static bool
+_fle2_collect_keys_for_compact (mongocrypt_ctx_t *ctx)
+{
+   _mongocrypt_ctx_encrypt_t *ectx = (_mongocrypt_ctx_encrypt_t *) ctx;
+
+   /* compactionTokens are only appended for FLE 2. */
+   if (_mongocrypt_buffer_empty (&ectx->encrypted_field_config)) {
+      return true;
+   }
```
vector-of-bool, line 370: I've seen this "check for fle2" a few times now spelled as checking whether `encrypted_field_config` is empty. It may ease readability to abstract this into a named function that does the check i.e. `context_uses_fle2()` rather than relying on comments to explain what the check is doing.

vector-of-bool (approved):
LGTM, but one comment about a possible additional level-of-indirection that may be useful for readability across several past changes.

src/mongocrypt-ctx-encrypt.c
kevinAlbs: Good idea. Added `context_uses_fle2()`

PR #326, opened 2022-04-30
MONGOCRYPT-426 Fix processing of `maxContentionCounter` in `FLE2EncryptionPlaceholder`

kevinAlbs:
# Summary

- Choose a random contentionFactor in the bounds on insert.
- Error with a negative `maxContentionCounter` on insert.

# Background & Motivation

The `maxContentionCounter` (also referred to as `maxContentionFactor`) is used to increase entropy in inserted data. On insert, a `contentionCounter` (also referred to as `contentionFactor`) must be chosen as a random integer in the range `[0, maxContentionCounter`] for deriving tokens. libmongocrypt incorrectly always chooses `contentionCounter == maxContentionCounter`.

Here is the server code describes choosing the random `contentionFactor` from `maxContentionCounter`: https://github.com/mongodb/mongo/commit/52e26913ceedc6cf87780922bae90f12d9cb62ff

The [Reference: Keys and Tokens](https://github.com/markbenvenuto/mongo-enterprise-modules/blob/fle2/fle_protocol.md#reference-keys-and-tokens) also notes the expected behavior:

> u = or [1,r] where r = contention factor
> u == 0 if field has no contention otherwise u = random secure sample {1, .. max contention}.

To validate correctness, the test suggested in [this comment](https://github.com/mongodb/specifications/pull/1200#discussion_r862089938) is implemented in the Go driver [here](https://github.com/kevinAlbs/mongo-go-driver/pull/3/files#diff-55617555c17bae6fadeb3e9a0c0e3a10837496998c75937eba67712b871f58eeR151-R211).


src/mongocrypt-crypto.c
```diff
@@ -1658,3 +1658,35 @@ _mongocrypt_fle2_do_decryption (_mongocrypt_crypto_t *crypto,
...
+      return false;
+   }
+
+   int64_t rand_i64;
+   memcpy (&rand_i64, rand_i64_buf.data, rand_i64_buf.len);
+
+   /* Zero the leading bit to ensure rand_i64 is non-negative. */
+   rand_i64 &= (~(1ull << 63));
+
+   *out = rand_i64 % exclusive_upper_bound;
```
vector-of-bool, line 1688: Do we want strongly random numbers? This will introduce a subtle bias for non-powers-of-two `exclusive_upper_bound`.

src/mongocrypt-marking.c
```diff
@@ -457,13 +457,31 @@ _mongocrypt_fle2_placeholder_to_insert_update_ciphertext (
...
+   if (placeholder->maxContentionCounter < 0) {
+      CLIENT_ERR ("maxContentionCounter must be non-negative, got: %" PRId64,
+                  placeholder->maxContentionCounter);
+      goto fail;
+   }
+   if (placeholder->maxContentionCounter > 0) {
+      /* Choose a random contentionFactor in the inclusive range [0,
+       * placeholder->maxContentionCounter] */
+      if (!_mongocrypt_random_int64 (crypto,
+                                     placeholder->maxContentionCounter + 1,
```
vector-of-bool, line 470: If `maxContentionCounter` is `INT64_MAX`, that'll overflow this parameter, which will be UB, and then later a divide-by-zero with the `%` operator.

src/mongocrypt-crypto.c
vector-of-bool: Scratch the divide-by-zero, since that's checked, but still might overflow

src/mongocrypt-marking.c
vector-of-bool: Scratch the divide-by-zero, since that's checked, but still might overflow.

vector-of-bool (commented):
Looks good other than two comments about random generation.

src/mongocrypt-crypto.c
kevinAlbs: Great question. I was not aware about the modulo bias.

I am not sure if strongly random numbers are a requirement. The server uses a [secure random number generator](https://github.com/mongodb/mongo/blob/0ba7cf78b7d59497eafd38b5aafeac29e8484cf1/src/mongo/platform/random.h#L185). On Unix it is sourced from `/dev/urandom` into `std::uniform_int_distribution`.

@markbenvenuto is strongly random numbers a requirement for contention counter generation?

One straightforward looking option may be to implement something like [arc4_random_uniform](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/lib/libc/crypt/arc4random_uniform.c#L33).

kevinAlbs: Updated to add a `_mongocrypt_random_uint64` based on the linked implementation of `arc4random_uniform`.

src/mongocrypt-marking.c
```diff
@@ -457,13 +457,39 @@ _mongocrypt_fle2_placeholder_to_insert_update_ciphertext (
...
 
-   if (!_mongocrypt_fle2_placeholder_common (kb,
-                                             &common,
-                                             &placeholder->index_key_id,
-                                             &value,
-                                             true, /* derive tokens using counter */
-                                             placeholder->maxContentionCounter,
-                                             status)) {
+   int64_t contentionFactor = 0;
+   if (placeholder->maxContentionCounter < 0) {
```
sgolemon, line 461: Shouldn't this check occur in the palceholder parse function?

src/mongocrypt-marking.c
```diff
@@ -457,13 +457,39 @@ _mongocrypt_fle2_placeholder_to_insert_update_ciphertext (
...
+      CLIENT_ERR ("maxContentionCounter must be non-negative, got: %" PRId64,
+                  placeholder->maxContentionCounter);
+      goto fail;
+   }
+   if (placeholder->maxContentionCounter > 0) {
+      if (placeholder->maxContentionCounter < 0) {
```
sgolemon, line 467: How is this check expected to pass?

It was already checked on line 461, and it's antithetical to the check on line 466

src/mongocrypt-marking.c
```diff
@@ -457,13 +457,39 @@ _mongocrypt_fle2_placeholder_to_insert_update_ciphertext (
...
+         CLIENT_ERR ("Expected maxContentionFactor >= 0");
+         goto fail;
+      }
+      if (placeholder->maxContentionCounter == INT64_MAX) {
```
sgolemon, line 471: Again, this probably belongs in the parser, with at most a follow ASSERT here.

src/mongocrypt-marking.c
kevinAlbs: [_fle2_finalize_explicit](https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt-ctx-encrypt.c#L1280) constructs a `_mongocrypt_marking_t` without parsing. This is a common code path between explicit and automatic encryption.

Moved this check to a common helper to error earlier.

markbenvenuto (approved):

vector-of-bool (approved):

sgolemon (approved):

PR #327, opened 2022-05-04
feat(NODE-4220): add FLE2 explicit encryption API

addaleax:
~~This does not cover the currently proposed set of spec prose
tests, since it seems like those would typically be placed
into the main driver repository.~~

~~It does, however, provide test coverage for NODE-4218 / https://github.com/mongodb/libmongocrypt/pull/324 (still in review).~~

I decided to put the spec tests here. Not sure if that’s customary, but to me it just makes sense to put code and tests that get changed together in the same place.

Reviewers, please feel free to push your suggestions to this branch directly.

### DIFF
--- a/bindings/node/test/clientEncryption.test.js
+++ b/bindings/node/test/clientEncryption.test.js
@@ -631,13 +631,13 @@ describe('ClientEncryption', function () {
     });
 
     it('Case 5: can roundtrip encrypted unindexed', async function () {
-      const payload = await clientEncryption.encrypt('encrypted indexed value', {
+      const payload = await clientEncryption.encrypt('encrypted unindexed value', {
         keyId: KEY1_ID,
         algorithm: 'Unindexed'
       });
       const decrypted = await clientEncryption.decrypt(payload);
 
-      expect(decrypted).to.deep.equal('encrypted indexed value');
+      expect(decrypted).to.deep.equal('encrypted unindexed value');
     });
   });
 });


bindings/node/package.json
```diff
@@ -52,6 +52,7 @@
     "jsdoc-to-markdown": "^7.1.1",
     "mocha": "^9.2.0",
     "mongodb": "^4.3.1",
+    "mongodb-client-encryption": "file:.",
```
addaleax, line 55: This is so the driver package can pick up this one, so that the auto-encryption-enabled client in the spec tests works.

bindings/node/lib/clientEncryption.js
```diff
@@ -299,6 +302,14 @@ module.exports = function (modules) {
...
         contextOptions.keyAltName = bson.serialize({ keyAltName });
       }
+      if (options.algorithm === 'Indexed') {
+        delete contextOptions.algorithm;
+        contextOptions.indexType = 'Equality';
+      }
+      if (options.algorithm === 'Unindexed') {
+        delete contextOptions.algorithm;
+        contextOptions.indexType = 'None';
+      }
```
durran, line 312: Let me know if I missed it somewhere but the spec states `contentionFactor` must default to 0 and also an error must be raised if it is set and the algorithm is not `'Indexed'`. I don't think I see that in this PR but let me know if I'm wrong.

bindings/node/test/clientEncryption.test.js
```diff
@@ -330,4 +368,276 @@ describe('ClientEncryption', function () {
...
+    });
+    const KEY1_ID = KEY1_DOCUMENT._id;
+    const LOCAL_MASTERKEY =
+      'Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5aXRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk';
+
+    let keyVaultClient;
+    let encryptedClient;
+    let clientEncryption;
+
+    async function setup() {
```
durran, line 453: The spec states that the explicit encryption tests must not run against a standalone server. Could we create a ticket to address that separately as there's a good bit of setup we need to do to make that happen.

bindings/node/test/clientEncryption.test.js
```diff
@@ -330,4 +368,276 @@ describe('ClientEncryption', function () {
...
+        keyId: KEY1_ID,
+        algorithm: 'Indexed'
+      });
+      const decrypted = await clientEncryption.decrypt(payload);
+
+      expect(decrypted).to.deep.equal('encrypted indexed value');
+    });
+
+    it('Case 5: can roundtrip encrypted unindexed', async function () {
+      const payload = await clientEncryption.encrypt('encrypted indexed value', {
```
durran, line 634: indexed -> unindexed to match the spec

bindings/node/test/clientEncryption.test.js
```diff
@@ -330,4 +368,276 @@ describe('ClientEncryption', function () {
...
+        keyId: KEY1_ID,
+        algorithm: 'Unindexed'
+      });
+      const decrypted = await clientEncryption.decrypt(payload);
+
+      expect(decrypted).to.deep.equal('encrypted indexed value');
```
durran, line 640: indexed -> unindexed to match the spec

durran (changes requested):

bindings/node/lib/clientEncryption.js
addaleax: The spec is being descriptive here, not prescriptive – i.e. these things are handled and tested by libmongocrypt. It would actually be an error to explicitly set a default value for `contentionFactor` in the `'Unindexed'` case, even if that value matches the default.

bindings/node/test/clientEncryption.test.js
addaleax: Sure – https://jira.mongodb.org/browse/NODE-4236 :+1:

bindings/node/lib/clientEncryption.js
durran: Thanks for the explanation.

bindings/node/test/clientEncryption.test.js
durran: Thanks!

PR #328, opened 2022-05-05
A csfle-markup utility

vector-of-bool:
This is a small single-file command-line app that eats a JSON document and feeds it to `csfle`, then spits the result back out. Handy for debugging.

### DIFF
--- a/src/csfle-markup.cpp
+++ b/src/csfle-markup.cpp
@@ -1,3 +1,7 @@
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
 #include <bson/bson.h>
 
 extern "C" {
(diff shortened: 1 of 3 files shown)


src/csfle-markup.cpp
```diff
@@ -0,0 +1,230 @@
...
+      fprintf (stderr,
+               "Failed to analyze the given query: %s\n",
+               mongo_csfle_v1_status_get_explanation (status));
+      return 8;
+   }
+   auto del_data = DEFER ({ mongo_csfle_v1_bson_free (data_ptr); });
+
+   bson_t *output = bson_new_from_data (data_ptr, data_len);
+   assert (output);
+   auto del_output = DEFER ({ bson_free (output); });
```
kevinAlbs, line 209:
```suggestion
   auto del_output = DEFER ({ bson_destroy (output); });
```

kevinAlbs (approved):
LGTM, this may come in handy if we need to debug differences between csfle and mongocryptd again.

I like the DEFER pattern.

PR #330, opened 2022-05-09
CSHARP-4153: Add FLE 2 API to AutoEncryptionOpts.

DmitryLukyanov:


BorisDog (approved):
LGTM. Notice the minor comment.

bindings/cs/MongoDB.Libmongocrypt/CryptOptions.cs
```diff
@@ -41,6 +43,18 @@ public CryptOptions(IEnumerable<KmsCredentials> credentials) : this(credentials,
             Schema = schema;
         }
 
+        public CryptOptions(
```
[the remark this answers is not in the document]
DmitryLukyanov: done

bindings/cs/MongoDB.Libmongocrypt/PinnedBinary.cs
```diff
@@ -24,6 +24,23 @@ namespace MongoDB.Libmongocrypt
     /// <seealso cref="Binary" />
     internal class PinnedBinary : Binary
     {
+        #region static
+        internal static void HandleAsPinnedBinary<THandle>(THandle handle, byte[] bytes, Status status, Func<THandle, BinarySafeHandle, bool> handleFunc) where THandle : CheckableSafeHandle
```
BorisDog, line 28: minor: `Handle` naming is a bit unclear. But I don't have better suggestions, maybe `HandleCheck`?

bindings/cs/MongoDB.Libmongocrypt/CryptClient.cs
```diff
@@ -136,33 +123,55 @@ public CryptContext StartExplicitEncryptionContextWithKeyId(byte[] keyId, string
...
-                }
+                PinnedBinary.HandleAsPinnedBinary(handle, keyId, _status, (h, pb) => Library.mongocrypt_ctx_setopt_key_id(h, pb));
+            }
+            else if (keyAltName != null)
+            {
+                PinnedBinary.HandleAsPinnedBinary(handle, keyAltName, _status, (h, pb) => Library.mongocrypt_ctx_setopt_key_alt_name(h, pb));
             }
 
-            handle.Check(_status, Library.mongocrypt_ctx_setopt_algorithm(handle, encryptionAlgorithm, -1));
+            switch (encryptionAlgorithm)
```
BorisDog, line 149: Does this need to be case-sensitive, and why?

DmitryLukyanov: good question, initially I was going to agree with you and change it to support case insensitive check here, but then I noticed that handling old algorithm names are case sensitive on c level https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt-ctx.c#L292. So I think we should follow this logic for new algorithms as well. Making it case insensitive should happen together with changing a libmongocrypt level as well which I think makes sense but it should be a diff ticket

DmitryLukyanov: NOTE: spec says says about case sensitive check as well

bindings/cs/MongoDB.Libmongocrypt/PinnedBinary.cs
DmitryLukyanov: The right name here probably should be like: `RunWithBytesAsPinnedBinary`, maybe we can simplify it to: `RunAsPinnedBinary`?

PR #334, opened 2022-05-11
feat(NODE-4241): add support for rewrapManyDataKey and keyMaterial

addaleax:
spec PR: https://github.com/mongodb/specifications/pull/1213

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -375,12 +375,25 @@ export class ClientEncryption {
     options: ClientEncryptionCreateDataKeyProviderOptions
   ): Promise<Binary>;
 
+  /** Alias for @see createDataKey. */
+  createKey(
+    provider: ClientEncryptionDataKeyProvider,
+    callback: ClientEncryptionCreateDataKeyCallback
+  ): void;
+
+  /** Alias for @see createDataKey. */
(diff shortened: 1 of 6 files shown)


bindings/node/index.d.ts
```diff
@@ -296,6 +296,25 @@ export interface ClientEncryptionCreateDataKeyProviderOptions {
...
    */
   keyAltNames?: string[] | undefined;
+
+  /** @experimental */
+  keyMaterial?: Buffer | Binary;
+}
+
+/** @experimental */
+export interface ClientEncryptionRewrapManyDataKeyProviderOptions {
+  provider?: ClientEncryptionDataKeyProvider;
```
durran, line 306: The spec doesn't have this as optional - is there a case where it's not provided?

durran (changes requested):
Spec test changes part of NODE-4110.

@addaleax Do we need to wait on the spec to merge first before merging this?

addaleax:
> @addaleax Do we need to wait on the spec to merge first before merging this?

@durran Up to you – it would make our lives a bit easier if it's released earlier, I would be fine with it as long as everything is tagged as `@experimental` anyway 🙂

durran (approved):

PR #337, opened 2022-05-14
MONGOCRYPT-427 Fix explain

kevinAlbs:
# Summary

- Fix handling of the `explain` command for FLE 2.
   - Append "encryptionInformation" in correct location.
   - Strip "encryptionInformation" nested inside `explain`.
- Fix handling of the `explain` command for FLE 1 with `csfle`
  - Use command database name for value of "$db".
  - Append "$db" nested inside `explain`.

Integration tests were run in the C driver locally with both mongocryptd and csfle, and [on Evergreen here](https://spruce.mongodb.com/version/627ff0030ae6067da396c836/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

# Background & Motivation

These are the expectations of "explain" for csfle and mongocryptd found through experimenting.

## FLE1 explain expectations

mongocryptd and csfle expect "jsonSchema" as a sibling of "explain". Example:

```js
{
    "explain": {
        "find": "foo",
        "$db": "db" // Required by csfle. Prohibited by mongocryptd.
    },
    "jsonSchema": { },
    "isRemoteSchema": false,
    "$db": "db" // Appended by driver for mongocryptd. Appended by libmongocrypt for csfle.
}
```

csfle requires a "$db" field be nested in the "explain" document matching the top-level "$db".

Neither csfle or mongocryptd return "jsonSchema" or "$db" in the reply.

## FLE2 explain expectations

mongocryptd expects "encryptionInformation" as a sibling of "explain":

```js
{
    "explain": {
        "find": "foo"
    },
    "encryptionInformation": { "fields": [] },
    "$db": "db" // Appended by driver.
}
```

csfle expects "encryptionInformation" nested in "explain":

```js
{
    "explain": {
        "find": "foo",
        "encryptionInformation": { "fields": [] },
        "$db": "db"
    },
    "$db": "db"
}
```

mongocryptd and csfle return "encryptionInformation" nested in "explain". Neither csfle nor mongocryptd return "$db" in the reply.

csfle requires a "$db" field be nested in the "explain" document matching the top-level "$db".

mongod and mongos expect "encryptionInformation" nested in "explain".

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -172,6 +172,132 @@ _fle2_append_encryptionInformation (bson_t *dst,
    return true;
 }
 
+typedef enum { MC_TO_CSFLE, MC_TO_MONGOCRYPTD, MC_TO_MONGOD } mc_cmd_target_t;
+
+/**
+ * @brief Add "encryptionInformation" to a command.
+ *
+ * @param cmd_name The name of the command.
+ * @param cmd The command being rewritten. It is an input and output.
+ * @param ns The <db>.<collection> namespace for the command.
+ * @param encryptedFieldConfig The "encryptedFields" document for the
+ * collection.
+ * @param deleteTokens Delete tokens to append to "encryptionInformation". May
+ * be NULL.
+ * @param coll_name The collection name.
+ * @param db_name The database name.
+ * @param cmd_target The intended destination of the command. csfle,
+ * mongocryptd, and mongod have different requirements for the location of
+ * "encryptionInformation" and presence of a "$db" field.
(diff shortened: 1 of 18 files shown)


vector-of-bool (approved):
This LGTM, but it's an unfortunately very hairy workaround. Is there a SERVER ticket to consolidate this to a single consistent behavior? Or at least investigate why it's so strange?

kevinAlbs:
> This LGTM, but it's an unfortunately very hairy workaround. Is there a SERVER ticket to consolidate this to a single consistent behavior? Or at least investigate why it's so strange?

Filed SERVER-66508 to request "explain" be made consistent between csfle and mongocryptd.

PR #340, opened 2022-05-18
Ensure we handle all enums in a switch and have a default case

vector-of-bool:
Fixes MONGOCRYPT-430

A bogus value for `queryType` would break invariants of the encryption context. Instead, catch that early and return an error.


vector-of-bool:
GCC caught more missing `default:` labels that I had locally. Could you take a look again to make sure these aren't going to cause trouble?

addaleax (approved):

kevinAlbs (approved):
LGTM. The test-java task failures look unrelated.

PR #341, opened 2022-05-19
MONGOCRYPT-429 process create, collMod, and createIndexes commands

kevinAlbs:
# Summary

- Do not bypass `create`, `createIndexes`, or `collMod`.
- Run `isMaster` on `mongocryptd` for `create` and `createIndexes` to determine `maxWireVersion`.
- Do not send `listCollections` for `create`.

Integration tests proposed in https://github.com/mongodb/specifications/pull/1227 were run with the [Go driver](https://github.com/kevinAlbs/mongo-go-driver/pull/6) and the [C driver](https://github.com/mongodb/mongo-c-driver/pull/1005) using this branch.

# Background & Motivation

DRIVERS-2309 requests running "create", "createIndexes", and "collMod" commands through query analysis (`mongocryptd` or `csfle` shared library).

The "create" and "collMod" commands may contain sensitive data in `validators`:

```js
{
    "create": "encryptedCollection",
    "validator": {
        // "a", "b", and "c" are sensitive.
        "encryptedField": { "$in": [ "a", "b", "c" ]}
    }
}
```

The "createIndexes" field may contain sensitive data in the `partialFilterExpression`:

```js
{
    "createIndexes": "encryptedCollection",
    "indexes": {
        "name": "name",
        "key": { "name": 1 },
        "partialFilterExpression": {
            // "a", "b", and "c" are sensitive.
            "encryptedField": { "$in": [ "a", "b", "c" ]}
        }
    }
}
```

## Version Checking

`mongocryptd` versions older than 6.0.0-rc5 return an error on `create` and `createIndexes`:

```
> db.runCommand({create: "foo", jsonSchema: {}})
{
	"ok" : 0,
	"errmsg" : "no such command: 'create'",
	"code" : 59,
	"codeName" : "CommandNotFound"
}
MongoDB Enterprise > db.version()
5.0.5
```

The proposed solution is to send an `isMaster` command to determine the version of `mongocryptd`. If `mongocryptd` is older than 6.0.0, `libmongocrypt` keeps the old behavior and does not run `create` or `createIndexes` through query analysis.

## Not sending `listCollections` in `create`

The collection is not expected to have a server side schema if it is being created. The caveat is that it adds another special case to the state machine.


src/mongocrypt-ctx-encrypt.c
```diff
@@ -451,11 +451,7 @@ _fle2_collect_keys_for_deleteTokens (mongocrypt_ctx_t *ctx)
       return true;
    }
 
-   const char *cmd_name = get_command_name (&ectx->original_cmd, ctx->status);
-   if (!cmd_name) {
-      _mongocrypt_ctx_fail (ctx);
-      return false;
-   }
+   const char *cmd_name = ectx->cmd_name;
```
vector-of-bool, line 454: I like the deduplication of this block.

vector-of-bool (approved):
LGTM

PR #343, opened 2022-05-23
MONGOCRYPT-431 Preserve $db

kevinAlbs:
# Summary

- If $db is included in the input command, preserve it in the output command.
- If $db is not included in the input command, do not add it to the output command.
- Always omit the $db to the command for mongocryptd.

# Background & Motivation

C# expects $db to be on the command after encryption. This was done by [injecting $db into the response from mongocryptd](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs#L268-L278). That is no longer possible when using the `csfle` shared library. The `MONGOCRYPT_CTX_NEED_MONGO_MARKINGS` state is not entered for `csfle`.

Other drivers append $db to the command after encryption. See [Do drivers append $db before or after encryption](https://docs.google.com/spreadsheets/d/18O6cfAQKC3OIUMW4_prWnpaLBMDh-lmv1F1RNUn0Sa0/edit#gid=0).

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -607,7 +607,11 @@ _fle2_mongo_op_markings (mongocrypt_ctx_t *ctx, bson_t *out)
       return _mongocrypt_ctx_fail (ctx);
    }
 
-   bson_copy_to (&cmd_bson, out);
+   // If input command included $db, do not include it in the command to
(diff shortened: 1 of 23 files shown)


vector-of-bool (approved):
An unfortunate edge case, but LGTM.

PR #346, opened 2022-05-23
MONGOCRYPT-316 bypass hello, buildInfo, getCmdLineOpts, and getLog

kevinAlbs:
# Summary

Resolves MONGOCRYPT-316 and MONGOCRYPT-308.

- Bypass auto encryption for the commands: `hello`, `buildInfo`, `getCmdLineOpts` and `getLog`.

# Background & Motivation

`mongocrypt_ctx_encrypt_init` uses an allow-list to determine how to handle commands in automatic encryption.
- Commands expected to contain no sensitive information may be bypassed. Bypassed commands are sent as-is.
- Eligible commands are sent to query analysis.
- Commands not listed result in an error.


kevinAlbs:
> That lengthy chain of `strcmp` is looking like it could use a transformation to an array/set/map. But I do not consider it a blocker. LGTM.

Agreed.

PR #347, opened 2022-05-24
test(NODE-4254): use majority write concern

durran:
FLE2 tests need to use majority write concern.

### DIFF
--- a/bindings/node/test/clientEncryption.test.js
+++ b/bindings/node/test/clientEncryption.test.js
@@ -178,7 +178,9 @@ describe('ClientEncryption', function () {
       expect(dataKey._bsontype).to.equal('Binary');
 
       // Remove and re-insert with a fixed UUID to guarantee consistent output
-      const doc = (await keyVaultColl.findOneAndDelete({ _id: dataKey })).value;
+      const doc = (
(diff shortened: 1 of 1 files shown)


bindings/node/test/clientEncryption.test.js
```diff
@@ -178,7 +178,9 @@ describe('ClientEncryption', function () {
       expect(dataKey._bsontype).to.equal('Binary');
 
       // Remove and re-insert with a fixed UUID to guarantee consistent output
-      const doc = (await keyVaultColl.findOneAndDelete({ _id: dataKey })).value;
+      const doc = (
+        await keyVaultColl.findOneAndDelete({ _id: dataKey }, { writeConcern: { w: 'majority' } })
+      ).value;
       doc._id = new BSON.Binary(Buffer.alloc(16), 4);
       await keyVaultColl.insertOne(doc);
```
addaleax, line 185: This one is probably the one where it affects the rest of the test the most?

addaleax (approved):

bindings/node/test/clientEncryption.test.js
durran: Yeah I think this one is the most and the other is just 1.

PR #352, opened 2022-05-31
[MONGOCRYPT-434] Rename "csfle" to "crypt_shared"

vector-of-bool:
This changeset replaces almost all references of the "csfle" dynamic library
with "crypt_shared", as per DRIVERS-2338/MONGOCRYPT-434.
This change includes renames to both the public and internal API
references. Some filenames and test names remain unchanged.

Also see: https://github.com/mongodb/specifications/pull/1236

This change also includes updates to the Node and Python bindings to use the new names and APIs

Note: Using these changes depend on updates to the downloads manifest that will also rename the `csfle` artifact to `crypt_shared`. The changes have so far been manually verified by using a `mongo_crypt_v1.so` library downloaded by hand.

### DIFF
--- a/bindings/node/test/autoEncrypter.test.js
+++ b/bindings/node/test/autoEncrypter.test.js
@@ -396,7 +396,7 @@ describe('AutoEncrypter', function () {
         }
       });
 
-      expect(this.mc).to.have.property('csfleVersionInfo', null);
+      expect(this.mc).to.have.property('cryptSharedLibVersionInfo', null);
 
       const localMcdm = this.mc._mongocryptdManager;
       sandbox.spy(localMcdm, 'spawn');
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM. Please wait to merge until the new shared library is available to download to ensure tests pass.

bindings/node/lib/autoEncrypter.js
```diff
@@ -126,34 +126,39 @@ module.exports = function (modules) {
         mongoCryptOptions.logger = options.logger;
       }
 
-      if (options.extraOptions && options.extraOptions.csflePath) {
-        mongoCryptOptions.csflePath = options.extraOptions.csflePath;
+      if (options.extraOptions && options.extraOptions.cryptSharedLibPath) {
+        mongoCryptOptions.cryptSharedLibPath = options.extraOptions.cryptSharedLibPath;
```
addaleax, line 130: Thank you for addressing this as well! :blue_heart:

@nbbeeken @durran @dariakp @baileympearson Heads up that this is a breaking change from 2.1.0 – that should be okay on our side, and I don’t think the team really announced shared library support, so I’d be okay with not introducing aliases here.

I’ve opened https://github.com/mongodb/node-mongodb-native/pull/3280 to adjust the typings.

addaleax (approved):

PR #354, opened 2022-06-01
MONGOCRYPT-435 do not run ismaster check if bypassing query analysis

kevinAlbs:
# Summary

- Do not run ismaster check of mongocryptd if bypassing query analysis

# Background & Motivation

MONGOCRYPT-429 added logic to run an `isMaster` command on mongocryptd when automatically encrypting the `create` and `createIndexes` command. The `create` and `createIndexes` commands are only supported on mongocryptd >= 6.0.0-rc5. The `isMaster` check is used to determine if mongocryptd supports `create` and `createIndexes`.

If query analysis is bypassed is set, a context expected not to enter the `MONGOCRYPT_CTX_NEED_MONGO_MARKINGS` state.

This change is validated with a C driver specification test here: https://github.com/mongodb/mongo-c-driver/pull/1023


vector-of-bool (approved):
LGTM

PR #358, opened 2022-06-03
PYTHON-3285 Support explicit queryable encryption

ShaneHarvey:


bindings/python/pymongocrypt/explicit_encrypter.py
```diff
@@ -17,19 +17,37 @@
...
+          - `contention_factor` (int): The contention factor to use
+            when the algorithm is "Indexed".
+
+        .. versionchanged:: 1.3
+           Added the `index_key_id`, `query_type`, and `contention_factor` parameters.
         """
         self.algorithm = algorithm
         self.key_id = key_id
         self.key_alt_name = key_alt_name
+        self.index_key_id = index_key_id
```
blink1073, line 41: Should we validate this input as well?

bindings/python/test/test_mongocrypt.py
```diff
@@ -476,14 +477,74 @@ def _test_encrypt_decrypt(self, key_id=None, key_alt_name=None):
...
+        with self.assertRaisesRegex(MongoCryptError, "algorithm"):
+            encrypter.encrypt(encoded_val, "Invalid", key_id)
+        # Invalid index_key_id type.
+        with self.assertRaises(TypeError):
+            encrypter.encrypt(encoded_val, "Indexed", key_id, index_key_id=1)
+        with self.assertRaises(MongoCryptError):
+            encrypter.encrypt(encoded_val, "Indexed", key_id, index_key_id=b'1234')
+        # Invalid query_type type.
+        with self.assertRaisesRegex(TypeError, "query_type"):
+            encrypter.encrypt(encoded_val, "Indexed", key_id, query_type='not a int')
```
blink1073, line 504: nit: not "an" int?

bindings/python/pymongocrypt/explicit_encrypter.py
ShaneHarvey: Good question. Let's treat it the same way as the existing "key_id" which is validated in PyMongo.

bindings/python/test/test_mongocrypt.py
ShaneHarvey: done.

blink1073 (approved):
LGTM!

PR #364, opened 2022-06-06
fix(NODE-4251): fix markCommand deserialization

durran:
Fixes the int/long mismatch when providing a contention value for equality queries.

### DIFF
--- a/bindings/node/lib/stateMachine.js
+++ b/bindings/node/lib/stateMachine.js
@@ -432,15 +432,16 @@ module.exports = function (modules) {
      */
     markCommand(client, ns, command, callback) {
       const bson = this.bson;
+      const options = { promoteLongs: false };
       const dbName = databaseNamespace(ns);
-      const rawCommand = bson.deserialize(command, { promoteLongs: false, promoteValues: false });
+      const rawCommand = bson.deserialize(command, options);
 
-      client.db(dbName).command(rawCommand, (err, response) => {
+      const commandOpts = rawCommand.encryptedFields ? options : {};
+      client.db(dbName).command(rawCommand, commandOpts, (err, response) => {
         if (err) {
           callback(err, null);
           return;
         }
-
         callback(err, bson.serialize(response, this.options));
       });
     }


bindings/node/lib/stateMachine.js
```diff
@@ -432,15 +432,16 @@ module.exports = function (modules) {
      */
     markCommand(client, ns, command, callback) {
       const bson = this.bson;
+      const options = { promoteLongs: false };
```
addaleax, line 435: Just for context, I put in the `promoteValues: false` here to be explicit about the fact that we always want to be sure to forward the *exact* message from mongocryptd to libmongocrypt.

bindings/node/lib/stateMachine.js
```diff
@@ -432,15 +432,16 @@ module.exports = function (modules) {
...
       const dbName = databaseNamespace(ns);
-      const rawCommand = bson.deserialize(command, { promoteLongs: false, promoteValues: false });
+      const rawCommand = bson.deserialize(command, options);
 
-      client.db(dbName).command(rawCommand, (err, response) => {
+      const commandOpts = rawCommand.encryptedFields ? options : {};
```
addaleax, line 439: Why is this conditional on encryptedFields usage?

addaleax (commented):
@dariakp I think this is reasonably covered through spec tests, yeah

bindings/node/lib/stateMachine.js
durran: Because when I did it on all commands, our driver started having issues with values not being promoted where it expected it, or the types were getting promoted incorrectly. One example of this was cursor ids coming back as int32s instead of longs. This change is basically temporary until the server allows int/long to be interchangeable and then I'll revert this change.

addaleax: @durran Correct me if I'm wrong, but this flag controls only the decoding of  response for a request from libmongocrypt, back to libmongocrypt. Without this flag, we run the risk of changing the BSON types that libmongocrypt sees to be something other than what mongocryptd reported – isn't that a bug? and something that would mismatch the shared library behavior as well (where bson is passed through verbatim)?

If the time allows for it, I'd like to dig a bit deeper here.

durran: You are correct and I just brought it back for that case in the deserialization of the command and just use `{ promoteLongs: false }` on the subsequent command call.

bindings/node/lib/stateMachine.js
durran: I brought it back now.

bindings/node/lib/stateMachine.js
addaleax: @durran What I’m having trouble with here is that there are commands that fail when you turn type promotion off for all command responses. That sounds very much like there’s a bug somewhere else that’s being masked by the promotion of values, and would prefer to dig into what that bug is. The expectation by libmongocrypt is generally that it will receive the *exact* response from mongocryptd. (I’m also happy to chat about this tomorrow in person if you prefer.)

> This change is basically temporary until the server allows int/long to be interchangeable and then I'll revert this change.

The point is that this shouldn’t happen; it’s a _good_ thing that int/long not being interchangeable exposes bugs here.

durran: @addaleax Yes let's chat tomorrow.

addaleax (approved):
LGTM, and thank you for your patience! 😊

durran:
> LGTM, and thank you for your patience! 😊

Thank you for being thorough!

PR #365, opened 2022-06-07
Use mongo-c-driver as a subdirectory for obtaining libbson

vector-of-bool:


.evergreen/build_all.sh
```diff
@@ -18,20 +18,32 @@ evergreen_root="$(pwd)"
...
+    # Amazon Linux 2 (arm64) has a very old system CMake we want to ignore
+    IGNORE_SYSTEM_CMAKE=1 . ${evergreen_root}/libmongocrypt/.evergreen/find-cmake.sh
+    # Check if on macOS with arm64. Use system cmake. See BUILD-14565.
+    OS_NAME=$(uname -s | tr '[:upper:]' '[:lower:]')
+    MARCH=$(uname -m | tr '[:upper:]' '[:lower:]')
+    if [ "darwin" = "$OS_NAME" -a "arm64" = "$MARCH" ]; then
+        CMAKE=cmake
+    fi
+fi
+
```
vector-of-bool, line 37: This was inlined from `build_install_bson.sh`, which was deleted.

.evergreen/config.yml
```diff
@@ -23,13 +23,6 @@ functions:
           set -o errexit
           chmod u+x libmongocrypt/.evergreen/*.sh
           ./libmongocrypt/.evergreen/print-env-info.sh
-    - command: s3.get
-      params:
-        aws_key: '${aws_key}'
-        aws_secret: '${aws_secret}'
-        remote_file: '${project}/${branch_name}/${libmongocrypt_s3_suffix}/mongo-c-driver-src.tar.gz'
-        bucket: mciuploads
-        extract_to: mongo-c-driver
```
vector-of-bool, line 32: Every individual task now downloads its own copy of the mongo-c-driver source, no longer blocking all tasks until the `prep-c-driver-source` task.

.evergreen/find-cmake.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
```
vector-of-bool, line 1: This file was copied from mongo-c-driver directly.

.evergreen/linker-tests.sh
```diff
@@ -77,18 +77,13 @@ $CMAKE --build . --target install --config RelWithDebInfo
...
+LIBBSON2_SRC_DIR="$(system_path "$PWD")"
 
 # Build libmongocrypt, static linking against libbson2
 cd $linker_tests_root/libmongocrypt-cmake-build
-PREFIX_PATH="$(system_path $linker_tests_root/install/bson2)"
 INSTALL_PATH="$(system_path $linker_tests_root/install/libmongocrypt)"
 SRC_PATH="$(system_path $libmongocrypt_root)"
-$CMAKE -DCMAKE_BUILD_TYPE=RelWithDebInfo $ADDITIONAL_CMAKE_FLAGS -DCMAKE_PREFIX_PATH="$PREFIX_PATH" -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" "$SRC_PATH"
+$CMAKE -DCMAKE_BUILD_TYPE=RelWithDebInfo "-DMONGOCRYPT_MONGOC_DIR=$LIBBSON2_SRC_DIR" $ADDITIONAL_CMAKE_FLAGS -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" "$SRC_PATH"
 $CMAKE --build . --target install --config RelWithDebInfo
```
vector-of-bool, line 87: We no longer need to install the libbson2 with the patch, we just need to build libmongocrypt and point it at the patched version of mongo-c-driver.

CMakeLists.txt
```diff
@@ -12,19 +12,21 @@ endif ()
 
 set (CMAKE_C_STANDARD 99)
 
-option (ENABLE_SHARED_BSON "Dynamically link libbson (default is static)" OFF)
```
vector-of-bool, line 15: This option was moved to `ImportBSON.cmake`, and renamed to `USE_SHARED_LIBBSON`.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+   if (NOT COMMAND FetchContent_Declare)
+      # We need FetchContent in order to download the project.
+      message (FATAL_ERROR
+            "No MONGOCRYPT_MONGOC_DIR setting was defined, and the FetchContent.cmake "
+            "module is not available. Upgrade your CMake version, or provide a "
+            "MONGOCRYPT_MONGOC_DIR path to a mongo-c-driver directory (This is required "
+            "for libmongocrypt to find a libbson to use and link against).")
+   endif ()
+   include (FetchMongoC)
+   # The FetchMongoC module defines a MONGOCRYPT_MONGOC_DIR for use to use
```
rcsanchez97, line 106: This looks like a typo in the comment.

rcsanchez97 (commented):
Looks good after a quick first glance.  I happened to notice a small typo, which I noted in a line comment.  I'll take another closer look tomorrow or Monday.

cmake/mongocrypt-config.cmake
```diff
@@ -1,3 +1,58 @@
 include(CMakeFindDependencyMacro)
 find_dependency(kms_message 0.0.1)
 include("${CMAKE_CURRENT_LIST_DIR}/mongocrypt_targets.cmake")
+
+# Find the libbson that will be used for static linking
+if (DEFINED MONGOCRYPT_LIBBSON_STATIC_USE)
```
kevinAlbs, line 6: Is there a use case for consumers defining `MONGOCRYPT_LIBBSON_STATIC_USE` or `MONGOCRYPT_LIBBSON_SHARED_USE`. Does this override the target names?

.evergreen/linker-tests.sh
kevinAlbs: Nice

.evergreen/config.yml
kevinAlbs: This is fantastic.

.evergreen/prep_c_driver_source.sh
```diff
@@ -7,6 +7,6 @@ set -o errexit
 MONGO_C_DRIVER_VERSION=1.17.0
 
 # Force checkout with lf endings since .sh must have lf, not crlf on Windows
-git clone git@github.com:mongodb/mongo-c-driver.git --config core.eol=lf --config core.autocrlf=false --depth=1 --branch $MONGO_C_DRIVER_VERSION
+git clone https://github.com/mongodb/mongo-c-driver.git --config core.eol=lf --config core.autocrlf=false --depth=1 --branch $MONGO_C_DRIVER_VERSION
```
kevinAlbs, line 10: Curious, was using `https` necessary to avoid errors on clone? I noticed some hosts error cloning with ssh: https://jira.mongodb.org/browse/BUILD-15101

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+endif ()
+
+message (STATUS "Using [${MONGOCRYPT_MONGOC_DIR}] as a sub-project for libbson")
+
+function (_import_bson_add_subdir)
```
kevinAlbs, line 111: There is a task failure on the macOS m1 build [task failure](https://spruce.mongodb.com/task/libmongocrypt_macos_build_and_test_and_upload_patch_9b0cf4476d04606426aed5f979e0d8222ceb971c_62a3a40357e85a6082c7b3f3_22_06_10_20_05_24/logs?execution=0) 

> ld: warning: ignoring file /opt/homebrew/Cellar/zstd/1.4.9/lib/libzstd.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64

zstd may not be available for x86_64. I suggest using `ENABLE_ZSTD=OFF` to disable.

kevinAlbs (commented):
I like this. I think it simplifies the most common build configuration. I left comments and questions inline, but no major concerns.

eramongodb (changes requested):
Troubled by behavior of `add_subdirectory()` of the C Driver repository on the configuration process of libmongocrypt.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+   # Disable AWS_AUTH, to prevent it from building the kms-message symbols, which we build ourselves
+   set (ENABLE_MONGODB_AWS_AUTH OFF CACHE BOOL "Disable kms-message content in mongoc for libmongocrypt" FORCE)
+   # Disable install() for the libbson static library. We'll do it ourselves
+   set (ENABLE_STATIC BUILD_ONLY)
+   # Disable over-alignment of bson types
+   set (ENABLE_EXTRA_ALIGNMENT ${_extra_alignment_default} CACHE BOOL "Toggle extra alignment of bson_t")
+   # Add the subdirectory as a project. EXCLUDE_FROM_ALL to inhibit building and installing of components unless requested
+   add_subdirectory ("${MONGOCRYPT_MONGOC_DIR}" _mongo-c-driver EXCLUDE_FROM_ALL)
```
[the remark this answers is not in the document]
vector-of-bool: I remember seeing this issue previously. The mongo-c project overrides the message() command to be a no-op, for some reason. This was removed in a later version, but remains in 1.17.0.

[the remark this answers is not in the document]
eramongodb: Confirmed. It seems like this issue has been resolved since [C Driver release version 1.17.3](https://github.com/mongodb/mongo-c-driver/compare/1.17.2...1.17.3). Good to know.

Is there a reason the C Driver version is currently being pinned to 1.17.0 and not a newer release?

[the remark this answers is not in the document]
vector-of-bool: No specific reason other than "it already was". There are some platform issues regarding the newer Mongo-c versions because of atomics, but bumping to a bugfix of 1.17 should be fine.

kms-message/CMakeLists.txt
```diff
@@ -264,21 +264,18 @@ if (NOT DISABLE_NATIVE_CRYPTO)
...
 
    add_executable(test_kms_gcp_online test/test_kms_gcp_online.c test/test_kms_util.c test/test_kms_online_util.c)
    target_include_directories(test_kms_gcp_online PRIVATE ${PROJECT_SOURCE_DIR}/src)
    target_compile_definitions(test_kms_gcp_online PRIVATE ${KMS_MESSAGE_DEFINITIONS})
-   target_link_libraries(test_kms_gcp_online mongo::mongoc_shared)
+   target_link_libraries(test_kms_gcp_online _mongocrypt::mongoc)
    target_link_libraries(test_kms_gcp_online kms_message_static)
-endif ()
\ No newline at end of file
+endif ()
```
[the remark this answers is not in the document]
eramongodb: Disregard: duplicate comment on outdated revision of file.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+     MONGOCRYPT_MONGOC_DIR *will not* be included in the mongo::mongocrypt library.
+   - The installed libbson_for_shared will dynamically link to a libbson on the user's system by
+     using a find_library() call.
+
+   In both of the above cases, libbson_for_static will require that the final consumer
+   provide their own definitions of the libbson symbols, regardless of the value
+   of USE_SHARED_LIBBSON.
+]]
+
+include(CheckCSourceCompiles)
```
rcsanchez97, line 51:
```suggestion
include (CheckCSourceCompiles)
```

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+include(CMakePushCheckState)
+
+cmake_push_check_state ()
+   # Even though we aren't going to use the system's libbson, try to detect whether it has
```
rcsanchez97, line 55: To support the building of Debian packages and RPMs, there has to be an option to build against the system-installed libbson.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+   # extra-alignment enabled. We want to match that setting as our default, for convenience
+   # purposes only.
+   find_path(SYSTEM_BSON_INCLUDE_DIR bson/bson.h PATH_SUFFIXES libbson-1.0)
```
rcsanchez97, line 58:
```suggestion
   find_path (SYSTEM_BSON_INCLUDE_DIR bson/bson.h PATH_SUFFIXES libbson-1.0)
```

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+   if (SYSTEM_BSON_INCLUDE_DIR AND NOT DEFINED ENABLE_EXTRA_ALIGNMENT)
+      set(CMAKE_REQUIRED_INCLUDES "${SYSTEM_BSON_INCLUDE_DIR}")
```
rcsanchez97, line 60:
```suggestion
      set (CMAKE_REQUIRED_INCLUDES "${SYSTEM_BSON_INCLUDE_DIR}")
```

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+      set (_extra_alignment_default OFF)
+      check_c_source_compiles([[
+         #include <bson/bson.h>
+
+         int main() { }
+      ]] HAVE_SYSTEM_LIBBSON)
+
+      if (HAVE_SYSTEM_LIBBSON)
+         # We have a libbson, check for extra alignment
+         check_c_source_compiles([[
```
rcsanchez97, line 70:
```suggestion
         check_c_source_compiles ([[
```

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+
+            int main() {}
+         ]] SYSTEM_LIBBSON_IS_EXTRA_ALIGNED)
+         if (SYSTEM_LIBBSON_IS_EXTRA_ALIGNED)
+            # Extra aligned! We'll use extra alignment by default.
+            set (_extra_alignment_default ON)
+         endif ()
+      endif ()
+   endif ()
+cmake_pop_check_state()
```
rcsanchez97, line 85:
```suggestion
cmake_pop_check_state ()
```

cmake/ImportBSON.cmake
rcsanchez97, line 106: It just hit me what you probably meant to say here.

```suggestion
   # The FetchMongoC module defines a MONGOCRYPT_MONGOC_DIR for us to use
```

cmake/ImportBSON.cmake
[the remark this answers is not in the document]
rcsanchez97: Wow, that's a throwback.  Yes, it was a rather unfortunate part of the build.  I also see no issue to bumping to the latest 1.17 (which happens to be 1.17.7).

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+# Workaround: Embedded mongoc_static does not set its INCLUDE_DIRECTORIES for user targets
+target_include_directories (mongoc_static
+   PUBLIC
+      "$<BUILD_INTERFACE:${MONGOCRYPT_MONGOC_DIR}/src/libmongoc/src>"
+      "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/_mongo-c-driver/src/libmongoc/src/mongoc>"
+   )
+
+# Put the libbson dynamic library into the current binary directory (plus possible config suffix).
+# This ensures that libbson DLL will resolve on Windows when it searches during tests
+set_property(TARGET bson_shared PROPERTY RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
```
rcsanchez97, line 158:
```suggestion
set_property (TARGET bson_shared PROPERTY RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
```

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+
+if (ENABLE_STATIC)
+   # We are going to build a static libmongocrypt.
+   # We want the static libbson target from the embedded mongoc. Enable the static library as
+   # part of "all", and install the archive alongside the rest of our static libraries.
```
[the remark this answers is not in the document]
rcsanchez97: For reference, up until now we have only included the static libbson, as `libbson-static-1.0.a`, in the Debian packages and RPMs that we build for publication in our PPA.  The specific reason for this is that we do not publish libmongoc/libbson packages via PPA and so we must account for the very likely scenario where a libmongocrypt user is installing from the PPA to a platform that doesn't have a recent libbson or doesn't have it at all.

Making the libmongocrypt build install a `libbson-static-for-libmongocrypt.a` feels like not the right thing.  I'd rather see us go the plain vanilla external project route.  If the user does not provide an adequate libbson (either in system directories or in some other user-specified location), then we download the C driver, build the necessary components and install from the C driver build to the installation target specified by the user for the top-level libmongocrypt build.  We can then document for the user that if a suitable libbson isn't provided, that one will be fetched by the libmongocrypt build and will be installed alongside the libmongocrypt components.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+   # (Useful for some users for convenience of static-linking libmongocrypt: CDRIVER-3187)
+   set_target_properties(bson_static PROPERTIES
```
rcsanchez97, line 165:
```suggestion
   set_target_properties (bson_static PROPERTIES
```

cmake/mongocrypt-config.cmake
```diff
@@ -1,3 +1,58 @@
...
+    set (_static "${MONGOCRYPT_LIBBSON_STATIC_USE}")
+elseif (TARGET mongo::bson_static)
+    set (_static mongo::bson_static)
+else ()
+    find_library(
```
rcsanchez97, line 11:
```suggestion
    find_library (
```

cmake/mongocrypt-config.cmake
```diff
@@ -1,3 +1,58 @@
...
+#   need to be resolved before the library can be used.
+#
+#   (Note: static libmongocrypt *always* needs to link against an external libbson, as it does not
+#    embed the libbson symbols.)
+set (_using_shared_libbson "@USE_SHARED_LIBBSON@")
+
+if (NOT _using_shared_libbson)
+    # The libmongocrypt shared library already includes embedded libbson symbols, so there is
+    # no usage requirements for a libbson
+else()
```
rcsanchez97, line 42:
```suggestion
else ()
```

cmake/ImportBSON.cmake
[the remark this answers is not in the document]
vector-of-bool: The `libbson-static-for-libmongocrypt` file is to replace the `libbson-static` that was being copied and installed manually by the build shell scripts (The one created for CDRIVER-3187). The reason it is given the specific name here is because the Debian package build (correctly) errors because there's already a package (the actual libbson-dev package) that installs a file at the exact same filepath.

As for how to handle linking `mongocrypt_static` properly when libbson is not easily available, that's a good question. Previously the installed libmongocrypt would just rely on the `find_package()` result to propagate correctly, which isn't reliable in general:

- If the `find_package([some-libbson])` in `libmongocrypt` generated import targets (i.e. `mongo::bson[_static]`), the exported libmongocrypt targets would have link libraries of `mongo::bson[_static]`, and it was up to the user to ensure that those were visible by the time their CMake project finished configuring (i.e. the user had to `find_package()` to get libbson).
- If the `find_package()` generated the old `${BSON_FOO}` variables, then the generated libmongocrypt targets would have each of the libbson attributes repeated as interface properties and would have an explicit link to whatever the found libbson library file was found. I don't think this codepath is very common since most users are probably on a new enough libbson that the `mongo::bson` targets are installed.

In this changeset, mongocrypt and mongocrypt_static link-only to targets `_mongocrypt-libbson_for_shared` and `_mongocrypt-libbson_for_static`, respectively. These are interface targets that are constructed such that their installed version exports no usage requirements in the `mongocrypt-targets.cmake` files. Instead, `mongocrypt-config.cmake` does additional work to ensure that these interface targets point to the proper libbson files so that the link will execute successfully.

Of course, as in your attempt, this fails to work, because `mongocrypt-config.cmake` cannot find a `libbson` static library to attach to `_mongocrypt-libbson_for_static`. Do you have a `libbson-static-1.0.a` library installed? If so, the `find_library()` call should have succeeded. If the `find_library()` call fails, we _could_ issue a warning message to users, but it may be irrelevant for some users that do not wish to link the libmongocrypt statically.

Handling this properly is difficult when you attempt to build and install both static and dynamic libraries as a single package and cannot assume the way that the user wishes to link their final executable.  For example:

1. If a user wants to statically link with libmongocrypt, then:
   1. Static libmongocrypt _always_ requires libbson definitions to be provided when CMake generates the runtime binaries.
   2. If the user wishes to link libbson statically, then `mongo::mongocrypt_static` must link_libraries to a static libbson.
   3. If the user wishes to link libbson dynamically, then `mongo::mongocrypt_static` must link_libraries to a dynamic libbson.
   4. If the user doesn't care about linking to libbson for their final application: Too bad: the static libmongocrypt _requires_ that the libbson definitions be provided. We can attempt to do this transparently for the user in `mongocrypt-config.cmake`, but should the default be to statically link to libbson? Or dynamically? In the current changeset, this defaults to try `mongo::bson_static`, or to fall back and link against a `libbson-static-1.0` via `find_library()`.
2. If the user wants to dynamically link libmongocrypt:
   1. If the libmongocrypt was built with `USE_SHARED_LIBBSON`, then `mongo::mongocrypt` requires that the final application link to a libbson dynamic library, as the `mongo::mongocrypt` binary contains an `SONAME` reference to libbson. Statically linking with libbson is not an option for the user in this case.
   2. If the libmongocrypt was build with `USE_SHARED_LIBBSON=OFF` (the default), then there are no requirements on transitively linking libbson, and the user can do as they wish.

Alternatively, in the case of `mongo::mongocrypt_static`, we could forego attempting to solve this entirely and have a documentation-only requirement that the user bring-their-own libbson symbols for their runtime binaries. This would generate link errors unless the user is linking to libbson in some way. This would be the simplest solution, actually, and now I'm leaning towards it.

.evergreen/prep_c_driver_source.sh
vector-of-bool: Using SSH cloning is only available if the user has the github.com SSH key authorized, which won't be the case on fresh systems (e.g. I ran into issues using SSH cloning in a Earthly environment, whereas https cloning will "just work")

rcsanchez97: However, it is important to note that for CI some of the repos we clone are not accessible without authentication.  I believe everything under the `mongodb` namespace is available for anonymous access, but I'm not sure all of the others are.  Just something to keep in mind.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+function (_import_bson_add_subdir)
+   # Disable AWS_AUTH, to prevent it from building the kms-message symbols, which we build ourselves
+   set (ENABLE_MONGODB_AWS_AUTH OFF CACHE BOOL "Disable kms-message content in mongoc for libmongocrypt" FORCE)
+   # Disable install() for the libbson static library. We'll do it ourselves
+   set (ENABLE_STATIC BUILD_ONLY)
+   # Disable over-alignment of bson types
+   set (ENABLE_EXTRA_ALIGNMENT ${_extra_alignment_default} CACHE BOOL "Toggle extra alignment of bson_t")
+   # Add the subdirectory as a project. EXCLUDE_FROM_ALL to inhibit building and installing of components unless requested
+   add_subdirectory ("${MONGOCRYPT_MONGOC_DIR}" _mongo-c-driver EXCLUDE_FROM_ALL)
+endfunction ()
```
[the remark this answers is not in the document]
vector-of-bool: Because of `EXCLUDE_FROM_ALL`, all of the non-required components in `mongo-c-driver` will not be built unless they are explicitly built or linked by something that _is_ built. This allows us to omit toggling tests/examples/mongoc.

cmake/ImportBSON.cmake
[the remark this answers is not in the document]
rcsanchez97: I would also lean toward the simpler documentation-only solution.  The user who is consuming libmongocrypt directly is most likely also consuming the C and/or C++ driver directly.  It seems unlikely that they would be trying to build only libmongocrypt without the need for any other driver support (not saying it would never happen, just that it seems unlikely to be common).  The other languages (Ruby, Node, etc.,) provide the appropriate shared and/or static libraries within their packages (which they obtain from the Evergreen build), so expecting that someone who directly consumes libmongocrypt to also provide their own libbson (static and/or dynamic as suits their particular use case) seems like a perfectly reasonable expectation, especially if in the common case they would already have to build it.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+     the MONGOCRYPT_MONGOC_DIR.
+   - mongo::mongocrypt will be built with a dynamic link requirement on a libbson dynamic
+     library, which must be resolved at runtime by consumers. The translation units from the
+     MONGOCRYPT_MONGOC_DIR *will not* be included in the mongo::mongocrypt library.
+   - The installed libbson_for_shared will dynamically link to a libbson on the user's system by
+     using a find_library() call.
+
+   In both of the above cases, libbson_for_static will require that the final consumer
+   provide their own definitions of the libbson symbols, regardless of the value
+   of USE_SHARED_LIBBSON.
```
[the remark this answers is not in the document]
vector-of-bool: > @kevinAlbs: To confirm, is it is no longer possible to use a system install (e.g. link to /usr/local/lib/libbson.so)?

It is, but also isn't. In the case of `USE_SHARED_LIBBSON`, generated files will contain a dynamic-link requirement to a libbson dynamic library, referenced by a portable name (i.e. ELF uses the SONAME, DLL uses the filename, etc.). It is up to the runtime linker to decide what file will satisfy the reference and resolve the external symbols. The main purpose of the "linking" the dynamic libmongocrypt library to a libbson dynamic library is to embed this dynamic-link reference and ensure that the referrenced symbols actually exist in the dynamic library, but the dynamic link does not pin that dynamic library reference nor the resolved symbols to any particular instance of the dynamic library. In the current revision, the `USE_SHARED_LIBBSON` will compile the libbson from the subproject and then link against it. This will generate the appropriate runtime linking rules that a system's libbson will also be able to satisfy the dynamic link. Because libbson is very stable, and because libmongocrypt here links to libbson 1.17, any newer libbson-1.0 will satisfy the link, including a system's libbson.

---

> @rcsanchez97: It looks like there needs to be a way to tell the build "use the system libbson and don't go out to the network looking for it".

Is it required that the actual build-time `ld` invocation _specifically use_ the `libbson` installed in `/usr/lib`, or is the requirement that the generated runtime linker request be _satisfyable_ by the `libbson` in `/usr/lib`?

- If the former:
  - The `ImportBSON.cmake` can be tweaked to instead define `_mongocrypt-libbson_for_{static,shared}` to link against the `mongo::libbson` targets resolved via a `find_package()` call, as they were before. This would still require the intermediate build step to install a partial `mongo-c-driver` that omits the libmongocrypt dependency.
- If the latter:
  - The generated source-dist can include the fetch snapshot of `mongo-c-driver` and `ImportBSON.cmake` can `add_subdirectory()` that directory without invoking `FetchBSON`. This design would alleviate the requirement for the separate `pkg.mongo-c-driver.no-libmongocrypt` step, and the resulting binaries would still be satisfied by the dynamic `libbson` from the `libbson-1.0` package.

cmake/ImportBSON.cmake
vector-of-bool: [See this comment](https://github.com/mongodb/libmongocrypt/pull/365/files#r896084394)

cmake/ImportBSON.cmake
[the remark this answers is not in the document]
rcsanchez97: In the Debian package and RPM builds (the official builds, not the packaging builds we run in Evergreen to create packages for the PPA), the process involves first installing the build-time dependencies (including libbson from a package which can be installed via apt or dnf, as applicable to the platform), then building libmongocrypt against that specific libbson.  In theory, I can't think of a reason why the build must be limited to linking to the specific libbson installed in `/usr/lib` by the build process.  However, in practice the limitation that the build complete without accessing remote network resources means that the only available libbson will be in `/usr/lib` when building Debian packages and RPMs.

The Debian and Fedora policies would not allow for the solution you are proposing.  I think the simplest way to state the requirement is "there is must be a way to build libmongocrypt with only system components and the components included in the libmongocrypt source tree itself (i.e., no remote network access from the build)."  Note that the C++ driver has a similar situation with the polyfill.  There is one build configuration which downloads MNMLSTC (and which is therefore unsuitable for Debian package and RPM builds), and another that allows building the C++ driver against and already installed Boost.

I have no objection to making the experience simpler for libmongocrypt users, but we will need a patch that allows us to meet the requirements of distribution packaging policies.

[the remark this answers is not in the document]
vector-of-bool: I've added a special condition in `ImportBSON` that can be used to override the behavior for use by package maintainers.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,170 @@
...
+target_link_libraries (_mongocrypt-libbson_for_static INTERFACE $<BUILD_INTERFACE:bson_static>)
+
+# And an alias to the mongoc target for use in some test cases
+add_library (_mongocrypt::mongoc ALIAS mongoc_static)
+# Workaround: Embedded mongoc_static does not set its INCLUDE_DIRECTORIES for user targets
+target_include_directories (mongoc_static
+   PUBLIC
+      "$<BUILD_INTERFACE:${MONGOCRYPT_MONGOC_DIR}/src/libmongoc/src>"
+      "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/_mongo-c-driver/src/libmongoc/src/mongoc>"
+   )
```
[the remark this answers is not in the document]
vector-of-bool: I see no benefit to provide end-users (not package-maintainers) the option to import the system's libbson and libmongoc as a supported configuration option. If a user wishes to provide a tweaked version of either library, they can do so by setting `MONGOCRYPT_MONGOC_DIR` to a path to a local copy of the mongo-c-driver. If they wish to link their final application against the system's libbson (static or shared), this will still "just work" even if we `add_subdirectory()` a copy of mongo-c-driver.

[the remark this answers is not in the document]
eramongodb: > I see no benefit to provide end-users (not package-maintainers) the option to import the system's libbson and libmongoc as a supported configuration option.

I was not distinguishing between "end-users" and "package-maintainers", as these configuration variables are all public (regardless of documentation). However, it seems you have effectively managed to (partially) implemented the feature I was requesting in https://github.com/mongodb/libmongocrypt/pull/365/commits/b91cfd77713849dc1da9957352f857636283fe81 via the `MONGOCRYPT_MONGOC_DIR` variable instead. This obviates the need to use the `<library>_DIR` variables as I was suggesting.

The intent of these suggestions can now be interpreted as extending the behavior of `MONGOCRYPT_MONGOC_DIR` to include the libmongoc library being linked against by libmongocrypt online tests for consistency. Given latest state of PR (as of 5c6c431d), configuring libmongocrypt with `-DMONGOCRYPT_MONGOC_DIR=USE-SYSTEM -DUSE_SHARED_LIBBSON=ON -DENABLE_ONLINE_TESTS=ON` produces the following configuration error(s):
```
CMake Error at cmake/ImportBSON.cmake:170 (add_library):
  add_library cannot create ALIAS target "_mongocrypt::mongoc" because target
  "mongoc_static" does not already exist.
  CMakeLists.txt:24 (include)
CMake Error at cmake/ImportBSON.cmake:172 (target_include_directories):
  Cannot specify include directories for target "mongoc_static" which is not
  built by this project.
  CMakeLists.txt:24 (include)
CMake Error at cmake/ImportBSON.cmake:191 (install):
  install FILES given no DESTINATION!
  CMakeLists.txt:24 (include)
```

Given these configuration variables, libmongocrypt's online tests should ideally link with the same libmongoc library being provided by `MONGOCRYPT_MONGOC_DIR` as already being done for libbson, rather than failing configuration. This includes `-DMONGOCRYPT_MONGOC_DIR=USE-SYSTEM -DUSE_SHARED_LIBBSON=OFF` (if we wish to support statically linking with system libbson/libmongoc libraries), which currently fails during configuration.

CMakeLists.txt
```diff
@@ -432,35 +402,31 @@ endforeach ()
...
+   if (ENABLE_ONLINE_TESTS)
       message ("compiling utilities")
       add_executable (csfle test/util/csfle.c test/util/util.c)
       target_link_libraries (csfle PRIVATE mongocrypt_static)
       target_include_directories (csfle PRIVATE ${CMAKE_BINARY_DIR}/src)
       target_include_directories (csfle PRIVATE ./src)
       target_include_directories (csfle PRIVATE ./kms-message/src)
-      target_link_libraries (csfle PRIVATE mongo::mongoc_shared)
+      target_link_libraries (csfle PRIVATE _mongocrypt::mongoc)
    endif ()
```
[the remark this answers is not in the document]
eramongodb: No longer applicable; see https://github.com/mongodb/libmongocrypt/pull/365#discussion_r897275456.

kms-message/CMakeLists.txt
```diff
@@ -264,21 +264,18 @@ if (NOT DISABLE_NATIVE_CRYPTO)
...
+   target_link_libraries(test_kms_azure_online _mongocrypt::mongoc)
    target_link_libraries(test_kms_azure_online kms_message_static)
 
    add_executable(test_kms_gcp_online test/test_kms_gcp_online.c test/test_kms_util.c test/test_kms_online_util.c)
    target_include_directories(test_kms_gcp_online PRIVATE ${PROJECT_SOURCE_DIR}/src)
    target_compile_definitions(test_kms_gcp_online PRIVATE ${KMS_MESSAGE_DEFINITIONS})
-   target_link_libraries(test_kms_gcp_online mongo::mongoc_shared)
+   target_link_libraries(test_kms_gcp_online _mongocrypt::mongoc)
    target_link_libraries(test_kms_gcp_online kms_message_static)
-endif ()
```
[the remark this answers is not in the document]
eramongodb: No longer applicable; see https://github.com/mongodb/libmongocrypt/pull/365#discussion_r897275456.

rcsanchez97 (approved):

cmake/ImportBSON.cmake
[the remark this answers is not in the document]
vector-of-bool: My personal vote it to leave this particular configuration (wherein we use online-tests as well as `"USE-SYSTEM"`) unsupported, possibly issuing a warning to that effect, or possibly another guard around `"USE-SYSTEM"` to check that the online tests are disabled (as the only intended use case thereof is for package maintainers that want to dynamic link with the distribution's libraries).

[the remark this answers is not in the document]
rcsanchez97: I can support that.  In a package maintenance context, I think that "fully offline" is the expected default.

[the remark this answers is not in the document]
eramongodb: >  issuing a warning to that effect, or possibly another guard around "USE-SYSTEM" to check that the online tests are disabled
So long as it is made clear that the given configurations are not supported, SGTM.

cmake/FetchMongoC.cmake
```diff
@@ -0,0 +1,20 @@
+include (FetchContent)
+
+# Set the tag that we will fetch.
+set (MONGOC_FETCH_TAG_FOR_LIBBSON "1.17.0" CACHE STRING "The Git tag of mongo-c-driver that will be fetched to obtain libbson")
+
+# Fetch the source archive for the requested tag from GitHub
+FetchContent_Declare (
+    embedded_mcd
+    URL "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/${MONGOC_FETCH_TAG_FOR_LIBBSON}.tar.gz"
```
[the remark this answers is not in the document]
eramongodb: Poke. If deemed unnecessary/undesirable, feel free to reject this suggestion and resolve the conversation.

eramongodb (approved):
Approving, as functional behavior LGTM.

However, I still have concerns regarding CMake configuration/generation error handling and corresponding messages (motivated by my own confusion when playing around with the new configuration variables). Some suggestions are given, but I do not consider them blockers, so feel free to resolve them as deemed (un)necessary.

cmake/FetchMongoC.cmake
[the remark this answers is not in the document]
vector-of-bool: Sorry I missed this. Indeed the `VERSION_CURRENT` logic is somewhat hairy and prevents this being workable without using git version tags. It would be nice to be able to fetch arbitrary branches and commits, though. One will likely want to do so eventually, but I'll kick that can down the road for now.

cmake/ImportBSON.cmake
```diff
@@ -0,0 +1,194 @@
...
+      message (STATUS "NOTE: Using system-wide libbson library. This is intended only for package maintainers.")
+      find_library (_MONGOCRYPT_SYSTEM_LIBBSON_SHARED "${CMAKE_SHARED_LIBRARY_PREFIX}bson-1.0${CMAKE_SHARED_LIBRARY_SUFFIX}")
+      find_library (_MONGOCRYPT_SYSTEM_LIBBSON_STATIC "${CMAKE_STATIC_LIBRARY_PREFIX}bson-static-1.0${CMAKE_STATIC_LIBRARY_SUFFIX}")
+      find_path (_MONGOCRYPT_SYSTEM_LIBBSON_INCLUDE_DIR bson/bson.h PATH_SUFFIXES libbson-1.0)
+      add_library (bson_shared SHARED IMPORTED)
+      add_library (bson_static STATIC IMPORTED)
+      set_target_properties (bson_shared bson_static PROPERTIES
+         IMPORTED_CONFIGURATIONS "Release"
+         INTERFACE_INCLUDE_DIRECTORIES "${_MONGOCRYPT_SYSTEM_LIBBSON_INCLUDE_DIR}"
+         )
```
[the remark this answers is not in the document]
vector-of-bool: The main reason that I left this error handling out (as well as the checks in other comments (as well as requiring three variables to independently have the correct value to activate this code)) is to discourage casual readers from using these options. i.e. to make this codepath inconspicuous. We have at least one CI job that will execute this codepath correctly for the deb package build.

cmake/ImportBSON.cmake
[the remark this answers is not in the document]
kevinAlbs: > Do you have a libbson-static-1.0.a library installed?

I do not. 

> Alternatively, in the case of mongo::mongocrypt_static, we could forego attempting to solve this entirely and have a documentation-only requirement that the user bring-their-own libbson symbols for their runtime binaries. This would generate link errors unless the user is linking to libbson in some way. This would be the simplest solution, actually, and now I'm leaning towards it.

I am also in favor of the proposed documentation-only solution.

kevinAlbs (approved):
LGTM

I prefer excluding this from the 1.5.0 stable release since 1.5.0 release candidates are out.

I think there is small risk this may break a consumer not tested in the libmongocrypt project. For example, the Go driver builds libmongocrypt [using compile.sh](https://github.com/mongodb/mongo-go-driver/blob/95de0fb36ca077bbe9a92b3fbf66ef0d28c6eeae/.evergreen/config.yml#L129). The default build now builds the full C driver by default, rather than only libbson.

Does that seem OK? I think it can be merged to master, and I can create an r1.5 branch from the commit before.

vector-of-bool:
@kevinAlbs that will work alright.

In subsequent revisions I'd like to refine it further to ensure consumers only need to build the minimum that is necessary, as quickly as possible.

PR #367, opened 2022-06-08
Make mongocrypt.h a non-generated header

vector-of-bool:
The only reason we generate `mongocrypt.h` at configure-time is to generate the `MONGOCRYPT_VERSION` macro string. Since we are already doing configure-time generation of `mongocrypt-config.h`, it is easier to just move the version macro into that file, which is already `#include`d by `mongocrypt.h`.

### DIFF
--- a/bindings/python/pymongocrypt/binding.py
+++ b/bindings/python/pymongocrypt/binding.py
@@ -57,11 +57,8 @@ ffi.cdef("""
  * See all public API documentation in: @ref mongocrypt.h
  */
 
-/**
- * @def MONGOCRYPT_VERSION
- * The version string describing libmongocrypt.
- * Has the form x.y.z-<pre>+<date>+git<sha>.
- */
+/* clang-format off */
+/* clang-format on */
 
 /**
  * Returns the version string for libmongocrypt.
--- a/src/mongocrypt.h
+++ b/src/mongocrypt.h
@@ -26,22 +26,15 @@
 #include "mongocrypt-export.h"
 #include "mongocrypt-compat.h"
 
-/**
- * @macro MONGOCRYPT_VERSION
- * @brief The version string describing libmongocrypt.
- * Has the form x.y.z-<pre>+<date>+git<sha>.
- */
-
-// clang-format off
+/* clang-format off */
 #ifndef __has_include
    #include "mongocrypt-config.h"
 #elif __has_include("mongocrypt-config.h")
    #include "mongocrypt-config.h"
 #else
-   #error "No 'mongocrypt-config.h' header is available. " \
-          "This must be generated in order to use libmongocrypt"
+   #error "No 'mongocrypt-config.h' header is available. This must be generated in order to use libmongocrypt"
 #endif
-// clang-format on
+/* clang-format on */
 
 /**
  * Returns the version string for libmongocrypt.
--- a/src/mongocrypt.h
+++ b/src/mongocrypt.h
@@ -33,7 +33,8 @@
    #if __has_include("mongocrypt-config.h")
       #include "mongocrypt-config.h"
    #else
-      #error "No 'mongocrypt-config.h' header is available. This must be generated in order to use libmongocrypt"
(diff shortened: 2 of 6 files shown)


bindings/python/pymongocrypt/binding.py
```diff
@@ -57,11 +57,8 @@ def _parse_version(version):
  * See all public API documentation in: @ref mongocrypt.h
  */
 
-/**
- * @def MONGOCRYPT_VERSION
- * The version string describing libmongocrypt.
- * Has the form x.y.z-<pre>+<date>+git<sha>.
- */
+/* clang-format off */
+/* clang-format on */
```
jmikola, line 61: What's the point of turning `clang-format` off and on in succession?

vector-of-bool: This string is generated by a script that strips certain content from `mongocrypt.h`. In this case, the lines removed were all C preprocessor directives, but these two surrounding comments remain.

rcsanchez97 (commented):
The `bindings/python/strip_header.py` gives an example of invocation on `mongocrypt.h.in`.  You might consider changing that.  Also, if`strip_header.py` has no use after the header rename, it might make sense to remove it.  There is also a reference and a link to `mongocrypt.h.in` in `README.md`.  It probably just needs to be updated to refer to `mongocrypt.h` rather than `mongocrypt.h.in`.

src/mongocrypt.h
```diff
@@ -25,14 +25,18 @@
...
- */
-#define MONGOCRYPT_VERSION "@MONGOCRYPT_BUILD_VERSION@"
+/* clang-format off */
+#ifndef __has_include
+   #include "mongocrypt-config.h"
+#else
+   #if __has_include("mongocrypt-config.h")
+      #include "mongocrypt-config.h"
+   #else
+      #error "No 'mongocrypt-config.h' header is available. This must be generated in order to use libmongocrypt"
```
eramongodb, line 36: Confirming, is the purpose of detecting and using `__has_include` solely to provide a better error message given ``"mongocrypt-config.h"` cannot be found?

src/mongocrypt.h
```diff
@@ -25,14 +25,18 @@
...
 #include "mongocrypt-compat.h"
-#include "mongocrypt-config.h"
 
-/**
- * @def MONGOCRYPT_VERSION
- * The version string describing libmongocrypt.
- * Has the form x.y.z-<pre>+<date>+git<sha>.
- */
-#define MONGOCRYPT_VERSION "@MONGOCRYPT_BUILD_VERSION@"
+/* clang-format off */
```
eramongodb, line 29: Suggest removing the disabling of `clang-format` within this code block.

Side note: is the clang-format option `IndentPPDirectives` worth considering?

`IndentPPDirectives: BeforeHash`:
```c
#ifndef __has_include
   #include "mongocrypt-config.h"
#else
   #if __has_include("mongocrypt-config.h")
      #include "mongocrypt-config.h"
   #else
      #error \
         "No 'mongocrypt-config.h' header is available. This must be generated in order to use libmongocrypt"
   #endif
#endif
```

`IndentPPDirectives: AfterHash` (my preference):
```c
#ifndef __has_include
#   include "mongocrypt-config.h"
#else
#   if __has_include("mongocrypt-config.h")
#      include "mongocrypt-config.h"
#   else
#      error \
         "No 'mongocrypt-config.h' header is available. This must be generated in order to use libmongocrypt"
#   endif
#endif
```

eramongodb (approved):

src/mongocrypt.h
vector-of-bool: I'd like to enable that option as it really helps with readability, but I'd reserve that for a change where we re-apply formatting across the project(s), since that would have effects in many files.

jmikola (approved):
Tested locally with PHPC. This will just require removing `mongocrypt.h` from our `AC_CONFIG_FILES` line in `config.m4` (and related change in `config.w32` for Windows).

PR #369, opened 2022-06-14
MONGOCRYPT-440 clarify error for unsupported CTR mode.

kevinAlbs:
# Summary
- Add message to errors for unsupported CTR mode.
- Add string of CCCryptorStatus to errors.

# Background & Motivation

Please see MONGOCRYPT-440 for Background & Motivation.

The list of error enums is taken from [CommonCryptoError.h on macOS 10.14](https://github.com/apple-oss-distributions/CommonCrypto/blob/878937e469969b56fa2d356a7af6a1420a555d5d/include/CommonCryptoError.h).

### DIFF
--- a/src/crypto/commoncrypto.c
+++ b/src/crypto/commoncrypto.c
@@ -23,6 +23,40 @@
 #include <CommonCrypto/CommonHMAC.h>
 #include <CommonCrypto/CommonRandom.h>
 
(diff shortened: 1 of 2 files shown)


vector-of-bool (approved):
Better error messages are always good. LGTM

PR #371, opened 2022-06-16
CSHARP-4004: Key Management API.

DmitryLukyanov:


bindings/cs/MongoDB.Libmongocrypt/CryptClient.cs
```diff
@@ -102,30 +102,6 @@ public CryptContext StartEncryptionContext(string db, byte[] command)
             return new CryptContext(handle);
         }
 
-        /// <summary>
```
DmitryLukyanov, line 105: these methods no longer needed, so it's better to remove them

BorisDog (approved):
LGTM

PR #376, opened 2022-06-23
Disable ENABLE_AUTOMATIC_INIT_AND_CLEANUP by default

eramongodb:
## Description

Currently, building the imported libmongoc library generates the following error message:

```
warning: Configure the driver with ENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF. Automatic cleanup is deprecated and will be removed in version 2.0. [-W#pragma-messages]
 ```

This PR disables `ENABLE_AUTOMATIC_INIT_AND_CLEANUP` by default when building the imported libmongoc library to silence this warning. (Note: we may also want to consider changing this default value in the C driver CMake configuration from `ON` to `OFF` if able, given how many users are having to go out of their way to explicit disable it.)


vector-of-bool (approved):
Good catch on this. LGTM

kevinAlbs (approved):

PR #379, opened 2022-06-25
CSHARP-4204: EncryptOpts queryType should accept a string consistent with queryType in encryptedFields.

DmitryLukyanov:


BorisDog (approved):
I guess `StartExplicitEncryptionContext` query type changes are in previous PR.
LGTM if so.

PR #382, opened 2022-06-28
Remove unnecessary fields in result of rewrapManyDataKey

eramongodb:
## Description

Long during initial implementation of `rewrapManyDataKey()`, the arbitrary decision was made to preserve as many fields of the key document(s) being rewrapped as possible. This turns out to be unnecessary: the bulk write operation(s) constructed by Drivers to `updateOne` the rewrapped key in the key vault collection only require:

- `_id`: the ID of the key to update.
- `keyMaterial`: the (possibly) new and rewrapped data key material.
- `masterKey`: the (possibly) new master key.

The `updateDate` field is updated by the `updateOne` command's `$currentDate` operator. All other fields are unnecessary and may give the incorrect impression that a `replaceOne` with the resulting document(s) may suffice to update the key vault collection.


kevinAlbs (approved):
Nice, this seems like a reasonable improvement to prevent drivers from misinterpreting the result of rewrapping.

PR #383, opened 2022-06-28
fix(NODE-4260): mongocrypterror has name getter

durran:
Switch the `name` property of `MongoCryptError` to a getter in order to avoid the property setting errors caused by bulk write errors in the Node driver.

### DIFF
--- a/bindings/node/lib/common.js
+++ b/bindings/node/lib/common.js
@@ -38,8 +38,11 @@ function collectionNamespace(ns) {
 class MongoCryptError extends Error {
   constructor(message) {
     super(message);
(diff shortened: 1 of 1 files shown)


addaleax (approved):

nbbeeken (approved):
Lgtm!

dariakp (changes requested):
Is there any sort of regression test we can add against the actual reported issue?

durran:
> Is there any sort of regression test we can add against the actual reported issue?

https://github.com/mongodb/node-mongodb-native/pull/3305 has the test since the error doesn't come from this library but rather the behaviour in our driver errors.

PR #384, opened 2022-06-28
Update Linux distro versions before branching

kevinAlbs:
# Summary
- Update Linux distro versions before branching

# Background & Motivation
The README should be updated on the master branch.


rcsanchez97 (approved):
LGTM

PR #387, opened 2022-06-28
fix(NODE-4339): remove indexKeyId

durran:
This PR removes the usage of indexKeyId in the Node bindings.

From DRIVERS-2348:

Queryable Encryption has two keys per field: the UserKey and IndexKey.

    The UserKey encrypts user data.
    The IndexKey is used to derive tokens for indexing.

libmongocrypt assumes IndexKeyId == UserKeyId when only the UserKeyId is specified.

IndexKeyId is must must match the "keyId" in encryptedFields for correct query results.

After discussing with Mark Benvenuto, Seny Kamara, Kenneth White, Cynthia Braund, the decision is not to add the IndexKeyId to EncryptOpts for these reasons:

    We do not need to support the use-case where IndexKeyId != UserKeyId.
    Adding IndexKeyId increases risk that a user sets an incorrect IndexKeyId.

### DIFF
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -365,9 +365,6 @@ module.exports = function (modules) {
       if (options.keyId) {
         contextOptions.keyId = options.keyId.buffer;
       }
-      if (options.indexKeyId) {
-        contextOptions.indexKeyId = options.indexKeyId.buffer;
-      }
       if (options.keyAltName) {
         const keyAltName = options.keyAltName;
         if (options.keyId) {
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -337,9 +337,6 @@ export interface ClientEncryptionEncryptOptions {
    */
   keyAltName?: string;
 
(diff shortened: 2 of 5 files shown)


bindings/node/lib/clientEncryption.js
```diff
@@ -365,9 +365,6 @@ module.exports = function (modules) {
       if (options.keyId) {
         contextOptions.keyId = options.keyId.buffer;
       }
-      if (options.indexKeyId) {
-        contextOptions.indexKeyId = options.indexKeyId.buffer;
-      }
```
addaleax, line 370: I think you might want to remove this from the C++ source as well, since `contextOptinos` extends `options` and so could still contain an `indexKeyId` option

dariakp (changes requested):
Just Anna's comment, unless we want to do that in a separate PR

bindings/node/lib/clientEncryption.js
durran: This is removed from the C++ source now.

addaleax (approved):

dariakp (approved):

PR #389, opened 2022-06-29
fix(NODE-4356): update rewrapManyDataKey to use UpdateOne bulk ops

addaleax:
##### fix(NODE-4356): update rewrapManyDataKey to use UpdateOne bulk ops

See the ticket description. Since edb12c6b54602f68abe has already
been merged, this is required in order for rewrapManyDatakey
(as it would otherwise inadvertently destroy key information when run).

##### fix(NODE-4332): leave out bulk write in rewrapManyDataKey if unnecessary

Do not perform a bulk write if no keys are being updated.

### DIFF
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -282,6 +282,10 @@ module.exports = function (modules) {
             return;
           }
 
+          if (dataKey.v.length === 0) {
+            return process.nextTick(cb, null, {});
+          }
+
           const dbName = databaseNamespace(this._keyVaultNamespace);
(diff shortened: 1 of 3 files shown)


bindings/node/lib/clientEncryption.js
```diff
@@ -282,12 +282,24 @@ module.exports = function (modules) {
             return;
           }
 
+          if (dataKey.v.length === 0) {
+            return process.nextTick(cb, null, {});
```
baileympearson, line 286: Is there a technical reason that process.nextTick is necessary here?

baileympearson (approved):

nbbeeken (approved):

PR #391, opened 2022-06-29
MONGOCRYPT-315 upload tagged releases to a URL containing the tag name

rcsanchez97:
The main goal with this change, as the PR title suggests, is to upload to a URL based on the tag name for release tags.  Note that this is in addition to the normal uploads based on the commit hash and that existing uploads and upload locations were deliberately left undisturbed.  A separate ticket (MONGOCRYPT-437) will deal with necessary adjustments to the existing upload locations.

These changes cannot be effectively observed in a patch build, so you will have to consult the `upload-all` task in the last two builds on the waterfall of the special Evergreen project:

https://evergreen.mongodb.com/waterfall/libmongocrypt-feature

The only difference between the aggregate changes on the feature branch used for the special Evergreen project and this PR is that the line

```
head_tag=$(git tag -l --points-at HEAD 'MONGOCRYPT-315/[0-9].*' | tail -1 | sed 's/MONGOCRYPT-315\///')
```

is replaced with

```
head_tag=$(git tag -l --points-at HEAD '[0-9].*' | tail -1)
```

as the `MONGOCRYPT-315/` tag prefix was used only for development purposes and real release tags will have the form `1.x.y`.

Once this PR is approved by everyone, I will do the following:

- delete the `MONGOCRYPT-315/*` tags and `MONGOCRYPT-315` branch in the base repo
- announce to the team that the remote can be pruned to remove the cruft
- merge this PR

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -21,8 +21,23 @@ functions:
       params:
         script: |-
           set -o errexit
-          chmod u+x libmongocrypt/.evergreen/*.sh
-          ./libmongocrypt/.evergreen/print-env-info.sh
+          cd libmongocrypt
+          chmod u+x .evergreen/*.sh
+          ./.evergreen/print-env-info.sh
+
+          if [ "${is_patch}" != "true" ]; then
+            # determine if we have a release tag present on HEAD
+            head_tag=$(git tag -l --points-at HEAD '[0-9].*' | tail -1)
+          fi
+          echo "tag_upload_location: $head_tag"
+
+          cat <<EOT > ../tag_expansion.yml
+          tag_upload_location: "$head_tag"
+          EOT
+    - command: expansions.update
+      params:
+        ignore_missing_file: true
+        file: tag_expansion.yml
 
   "tar and upload libmongocrypt libraries":
     - command: archive.targz_pack
@@ -469,9 +484,13 @@ tasks:
     - variant: macos
       name: build-and-test-and-upload
   commands:
+    - func: "fetch source"
     - command: shell.exec
       params:
-        script: mkdir all
+        script: |-
+          # this directory was created in "fetch source", but it isn't needed here
+          rm -rf libmongocrypt
+          mkdir all
     - func: "download tarball"
       vars: { variant_name: ubuntu1604 }
     - func: "download tarball"
@@ -521,6 +540,15 @@ tasks:
         target: libmongocrypt-all.tar.gz
         source_dir: all
         include: [./**]
+    - command: shell.exec
+      params:
+        script: |-
+          set -o errexit
+          if [ -n "${tag_upload_location}" ]; then
+            # the "fetch source" step detected a release tag on HEAD, so we
+            # prepare a local file for upload to a location based on the tag
+            cp -a libmongocrypt-all.tar.gz libmongocrypt-all-${tag_upload_location}.tar.gz
+          fi
     - command: s3.put
       params:
         aws_key: '${aws_key}'
@@ -538,6 +566,17 @@ tasks:
         bucket: mciuploads
         permissions: public-read
         local_file: 'libmongocrypt-all.tar.gz'
+        content_type: '${content_type|application/x-gzip}'
+    - command: s3.put
+      params:
+        aws_key: '${aws_key}'
+        aws_secret: '${aws_secret}'
+        remote_file: 'libmongocrypt-feature/all/${tag_upload_location}/libmongocrypt-all.tar.gz'
+        bucket: mciuploads
+        permissions: public-read
+        optional: true
+        display_name: 'libmongocrypt-all-${tag_upload_location}.tar.gz'
+        local_file: 'libmongocrypt-all-${tag_upload_location}.tar.gz'
         content_type: '${content_type|application/x-gzip}'
 
 - name: publish-packages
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -21,8 +21,23 @@ functions:
...
           set -o errexit
-          chmod u+x libmongocrypt/.evergreen/*.sh
-          ./libmongocrypt/.evergreen/print-env-info.sh
+          cd libmongocrypt
+          chmod u+x .evergreen/*.sh
+          ./.evergreen/print-env-info.sh
+
+          if [ "${is_patch}" != "true" ]; then
+            # determine if we have a release tag present on HEAD
+            head_tag=$(git tag -l --points-at HEAD '[0-9].*' | tail -1)
```
eramongodb, line 30:
```suggestion
            head_tag=$(git tag --points-at HEAD '[0-9].*' | tail -1)
```
Per [Git documentation](https://git-scm.com/docs/git-tag), `--points-at <object>` implies `--list`, so `-l` is redundant.

.evergreen/config.yml
```diff
@@ -21,8 +21,23 @@ functions:
...
+          fi
+          echo "tag_upload_location: $head_tag"
+
+          cat <<EOT > ../tag_expansion.yml
+          tag_upload_location: "$head_tag"
+          EOT
```
eramongodb, line 36: TIL `EOT`. I am not sure if I prefer it to:
```suggestion
          echo 'tag_upload_location: "$head_tag"' >> ../tag_expansion.yml
```
Feel free to disregard this suggestion and resolve this comment as desired.

.evergreen/config.yml
```diff
@@ -469,9 +484,13 @@ tasks:
     - variant: macos
       name: build-and-test-and-upload
   commands:
+    - func: "fetch source"
     - command: shell.exec
       params:
-        script: mkdir all
+        script: |-
+          # this directory was created in "fetch source", but it isn't needed here
+          rm -rf libmongocrypt
```
eramongodb, line 492: Is removal of this directory necessary? Otherwise I do not see the harm in leaving it as-is.

.evergreen/config.yml
eramongodb, line 30: Given `HEAD` points to a commit with the following tags:
- `1.2.3`
- `1.2.3-pre`
- `some-tag`
The command `git tag --points-at HEAD '[0-9].*' | tail -1` yields the prerelease version `1.2.3-pre` on my local machine. I _think_ the intent of `tail -1` is to select `1.2.3` instead? If so, this command may need to be updated with a `sort` accordingly for deterministic behavior.

rcsanchez97: However, on older versions of Git, it is required.  For example, [here is a failure](https://evergreen.mongodb.com/task/libmongocrypt_feature_ubuntu1604_arm64_build_and_test_and_upload_a4593971dca2134d1715fcdaaa58158680c093ab_22_06_24_00_49_31) on Ubuntu 16.04 without `-l`:

```
[2022/06/24 00:53:17.598] fatal: --points-at option is only allowed with -l.
```

.evergreen/config.yml
rcsanchez97: Good point.  This bit started as copy/pasta from another place where multiple lines are written to a file.  It turned out that I needed to handle this one its own in a different location and I didn't simplify the pattern.

rcsanchez97: However, I think that we actually need this:

```
echo "tag_upload_location: \"$head_tag\"" >> ../tag_expansion.yml
```

Otherwise with `'` we inhibit the expansion of `$head_tag`, which is a shell variable, not an Evergreen expansion.  I'll run a quick test in the other project to confirm the syntax.

eramongodb: It looks like you are correct. I was not aware `'` prohibits variable expansions.
Note, my suggestion above also uses `>>`; I do not know if that is appropriate either.

.evergreen/config.yml
rcsanchez97: The issue with `--sort` is that it is not present on some older version of Git.  In principle, you are correct.  However, in practice the release workflow requires a separate commit for every release.  That is to say, we will never release by going to the commit already tagged `x.y.z-rcW` and then re-tagging it `x.y.z` and pushing only the tag.

For a related example, have a look at the history of `build/calc_release_version.py` in the C driver and also at CDRIVER-2831.  In that instance, because we were working in Python it was easily possible to implement the sorting behavior in the application code.  Here, since we are working in shell, we'd have to resort to `sort` (or perhaps `sort -r`).

After typing all that I see that you might have been referring to the `sort` command all along and not to the `--sort` option to Git.  It doesn't seem strictly necessary here, but it certainly won't hurt and it would make things more robust.

.evergreen/config.yml
rcsanchez97: I don't think that it's necessary, per se.  However, since the task has no other reason to have the source tree, it seemed like a good thing to keep everything tidy.

rcsanchez97: OK.  I see what you are saying.  I don't think that a race condition like you are describing could occur based on how Evergreen works.  However, I don't have a strong conviction that the directory needs to be removed, so I'll drop the removal.

kevinAlbs (approved):
Nicely done. This LGTM with one path fix.

I suggest cherry-picking to the `r1.5` branch. If there is a 1.5.1 release, that may be a good opportunity to verify the upload works.

eramongodb (approved):

vector-of-bool (approved):
Sorry for the delay on this. LGTM!

PR #393, opened 2022-06-30
Download and install a newer CMake version for Linux

vector-of-bool:

PR #395, opened 2022-07-02
Node bindings should re-use top-level find-cmake.sh

vector-of-bool:
`find-cmake.sh` was updated in #393, which significantly reduces CI times. The Node bindings still had their own copy of `find_cmake.sh` that hadn't been updated.


durran (approved):

durran:
@vector-of-bool does anything else need to go in here? All the Node tests are passing so I assumed this was good to go.

vector-of-bool:
@durran nothing else is needed. I left it in a draft so as not to ping the reviewers until I had ironed out the build errors, then the weekend happened.

PR #396, opened 2022-07-02
Set MAKEFLAGS to provide compilation parallelism

vector-of-bool:
Make respects a `MAKEFLAGS` environment variable to specify additional command-line flags. It's possible to specify flags to the build executor to `cmake --build`, but we don't have a guarantee that `cmake --build` will actually be running Make. Setting `MAKEFLAGS` will give arguments to Make if it happens to be the build executor.

This change shaves off several minutes for most Linux and macOS tasks.


kevinAlbs (approved):
Fantastic, LGTM

PR #397, opened 2022-07-03
test(NODE-4343): contention factor required with indexed

durran:
`contentionFactor` must now be provided when encrypting as indexed. This updates the prose tests and extra unit test to provide it. See https://github.com/mongodb/specifications/commit/41b335af92c12258a4cbe7e93ad4f34ef31641a7

### DIFF
--- a/bindings/node/test/clientEncryption.test.js
+++ b/bindings/node/test/clientEncryption.test.js
@@ -283,7 +283,8 @@ describe('ClientEncryption', function () {
         .then(dataKey => {
           const encryptOptions = {
             keyId: dataKey,
-            algorithm: 'Indexed'
+            algorithm: 'Indexed',
+            contentionFactor: 0
           };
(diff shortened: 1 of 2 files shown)


bindings/node/test/clientEncryption.test.js
```diff
@@ -787,7 +790,8 @@ describe('ClientEncryption', function () {
     it('Case 4: can roundtrip encrypted indexed', async function () {
       const payload = await clientEncryption.encrypt('encrypted indexed value', {
         keyId: KEY1_ID,
-        algorithm: 'Indexed'
+        algorithm: 'Indexed',
+        contentionFactor: 10
```
addaleax, line 794: I don’t think it really matters since this test is never inserting data, but the spec says `0` here as well?

addaleax (approved):

PR #401, opened 2022-07-06
Revert "Node bindings should re-use top-level find-cmake.sh (#395)"

baileympearson:
commit ec2403e3e1cb3e362b178acc352626d8af28ed40 updated the Node bindings CI scripts to use the shared find-cmake script, but this change unintentionally broke the Node driver's CSFLE tests.

Updating the cmake path didn't resolve all the issues, we now have build issues with our bindings (see example patch https://spruce.mongodb.com/version/62c5df0c0305b96e4b828a82/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).  Rather than investigate this issue right now, I want to revert this change until it can be made and properly tested before merging.

The Node driver also has CSFLE work in-flight that is impacted by a broken CI, so we should either fix it quickly or just revert for now.

### DIFF
--- a/.evergreen/find-cmake.sh
+++ b/.evergreen/find-cmake.sh
@@ -3,14 +3,8 @@ set -o errexit  # Exit the script with error if any of the commands fail
 
 find_cmake ()
 {
-  # Check if on macOS with arm64. Use system cmake. See BUILD-14565.
-  local OS_NAME="$(uname -s | tr '[:upper:]' '[:lower:]')"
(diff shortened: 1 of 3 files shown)


nbbeeken (approved):

baileympearson:
evergreen patch https://spruce.mongodb.com/version/62c5e9ee1e2d171afa350c08/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #402, opened 2022-07-07
Suppress leak warning in OpenSSL with one shared file

vector-of-bool:
This fixes the failing test execution on SLES 15. The suppression filepath option was set to a relative path. This meant that the suppression used were sensitive to the directory in which an executable runs.

This change sets the suppression filepath to point to an absolute path in the root of the repository.


.lsan-suppressions
```diff
@@ -1,2 +1,3 @@
 leak:ccrng_cryptographic_init_once
-leak:ccrng_cryptographic_generate
\ No newline at end of file
+leak:ccrng_cryptographic_generate
+leak:CRYPTO_zalloc
```
kevinAlbs, line 3: Do you know why `CRYPTO_zalloc` reports a 56 byte leak on SLES 15? My guess is this is a bug in the version of OpenSSL on SLES 15. Either way, I think there is little to no risk this points to a bug in libmongocrypt. I do not think it is necessary to determine the root cause.

kevinAlbs (approved):
Nice, LGTM

PR #407, opened 2022-07-07
fix Java bindings version

rcsanchez97:
This fixes the Java bindings version on the `r1.5` branch in anticipation of a fix to the way that build versions are generated for the project.


jyemin (approved):
LGTM

PR #409, opened 2022-07-07
MONGOCRYPT-455 improve build version number generation

rcsanchez97:
Full Evergreen patch build: https://spruce.mongodb.com/version/62c72663d6d80a340d89de46/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

@jyemin I requested your review specifically because I modified the steps associated with the Java bindings versioning.

Note that after merging this into `master` then I will cherry-pick commit `80c6585` plus the commit from the merge of this PR onto `r1.5`.

PR #411, opened 2022-07-07
fix distro package repository version

rcsanchez97:
This update is required after the change proposed in #409. Once that PR is merged then I will merge this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ Create the file `/etc/yum.repos.d/libmongocrypt.repo` with contents:
 ```
 [libmongocrypt]
 name=libmongocrypt repository
-baseurl=https://libmongocrypt.s3.amazonaws.com/yum/redhat/$releasever/libmongocrypt/1.5/x86_64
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -179,7 +179,7 @@ Create the file `/etc/yum.repos.d/libmongocrypt.repo` with contents:
 ```
 [libmongocrypt]
 name=libmongocrypt repository
-baseurl=https://libmongocrypt.s3.amazonaws.com/yum/redhat/$releasever/libmongocrypt/1.5/x86_64
+baseurl=https://libmongocrypt.s3.amazonaws.com/yum/redhat/$releasever/libmongocrypt/1.6/x86_64
```
kevinAlbs, line 182: Why refer to 1.6? The most latest release is 1.5.0.

Is using 1.6 required to install the "Unstable Development Distribution Packages"? If so, I suggest noting to replace "1.5" with "1.6" in the "Unstable Development Distribution Packages" section.

rcsanchez97: Oops.  You are absolutely right.  I got a touch carried away.  I'll submit a different PR after I merge #409

PR #412, opened 2022-07-08
fix distro package repository version

rcsanchez97:
Follow-up to #411, leaving `1.5` (as that is correct) and addressing some occurrences of `1.3` (which are outdated).


rcsanchez97:
I will also cherry-pick this onto `r1.5` after merge the PR.

kevinAlbs (approved):

PR #413, opened 2022-07-09
Invoke CTest directly as part of the build script.

vector-of-bool:
This removes the separate test.sh step and instead defines CTest tests
for each of the programs that we wish to execute for testing. We then
run CTest after each of the main build types we create. This will
increase configuration coverage, simplify the CI process, and expedite
test failures in the CI process.

The GCP and Azure tests are now executed as well, but emit test-skip
markers if the required environment variables are not defined.


.evergreen/build_all.sh
```diff
@@ -40,6 +40,16 @@ else
...
+if test -z "${CTEST-}"; then
+    if test "$(basename "$CMAKE")" = "$CMAKE"; then
+        # We use CMake from PATH. Just use CTest from PATH as well.
+        CTEST=ctest
+    else
+        # Use the CTest that lives beside the CMake we are using.
+        CTEST="$(dirname "$CMAKE")/ctest"
+    fi
+fi
+
```
rcsanchez97, line 52: Could we do this instead?

```
if test -z "${CTEST-}"; then
    CTEST="${CMAKE%make}test"
fi
```

Or, even more briefly:

```
: ${CTEST:=${CMAKE%make}test}
```

rcsanchez97 (approved):
LGTM, with one suggestion to make the setting of the `CTEST` variable simpler/easier to follow.

.evergreen/build_all.sh
```diff
@@ -76,6 +78,7 @@ echo "Installing libmongocrypt"
 $CMAKE --build . --target install --config RelWithDebInfo
 $CMAKE --build . --target test-mongocrypt --config RelWithDebInfo
 $CMAKE --build . --target test_kms_request --config RelWithDebInfo
+"$CTEST" -C RelWithDebInfo
```
kevinAlbs, line 81: The test logs are now a [summary](https://evergreen.mongodb.com/lobster/evergreen/task/libmongocrypt_rhel_62_64_bit_build_and_test_and_upload_patch_b12cb018bbd1a4e8a7fb69f8baf241daf3f0e939_62cc5c139ccd4e4e1b0669bd_22_07_11_17_21_23/0/task#bookmarks=0%2C5735&l=1&shareLine=5382).

Adding a test failure locally does not output logs by default:
```
$ ctest .
(omitted for brevity)
The following tests FAILED:
          1 - mongocrypt (Subprocess aborted)
Errors while running CTest
Output from these tests are in: /Users/kevin.albertson/review/libmongocrypt-413/cmake-build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

Using the suggested options, the output of stderr is output on one line:
```
$ ctest . --output-on-failure --rerun-failed
(omitted for brevity)
No 'real' csfle library is available. The _test_encrypt_csfle_no_needs_markings test is a no-op.No 'real' csfle library is available. The _test_encrypt_fle2_explain_with_csfle test is a no-op.No 'real' csfle library is available. The _test_encrypt_fle1_explain_with_csfle test is a no-op.No 'real' csfle library is available. The _test_fle1_create_with_csfle test is a no-op.test error /Users/kevin.albertson/review/libmongocrypt-413/test/test-mongocrypt-ctx-encrypt.c:4281 _test_fle2_create(): !mongocrypt_ctx_finalize (ctx, out) failed with msg: (null)
(omitted for brevity)
0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.13 sec

The following tests FAILED:
          1 - mongocrypt (Subprocess aborted)
Errors while running CTest
```

If a test fails on Evergreen, I prefer it be easy to identify the failing test function and the failing assertion. Can the stderr and stdout logs be printed on failure in an easily readable form?

vector-of-bool: `--output-on-failure` should be added, yes. I forget that its not the default because I use an environment variable to turn that on locally.

The repeated `No 'real' csfle library is available. The _test_encrypt_csfle_no_needs_markings test is a no-op.` on a single line is actually just a bug (the message is missing a newline character).

.evergreen/config.yml
```diff
@@ -319,13 +318,6 @@ tasks:
     vars:
       compile_env: LIBMONGOCRYPT_EXTRA_CFLAGS="-fsanitize=address"
 
-- name: build-and-test-valgrind
-  commands:
-  - func: "fetch source"
-  - func: "build and test"
-    vars:
-      test_env: VALGRIND="valgrind --leak-check=full --error-exitcode=1"
```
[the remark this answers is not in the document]
vector-of-bool: As far as I am aware, we do not regular run Valgrind tasks, and I feel that the ASan+LSan instrumentation is sufficient to replace it. If you feel otherwise, I can revive the task, but it will take a bit more complexity since Valgrind would need to be inserted as the test launcher for each CTest test.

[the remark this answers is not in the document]
kevinAlbs: ASAN+LSan covers almost all cases that Valgrind checks except [uninitialized memory reads](https://github.com/google/sanitizers/wiki/AddressSanitizerComparisonOfMemoryTools).

I do not recall any time that the valgrind task failed without ASAN failing. I am not opposed to removing it.

kevinAlbs (approved):

PR #414, opened 2022-07-13
feat(NODE-4394): add key management API to node bindings

baileympearson:
[NODE-4330](https://jira.mongodb.org/browse/NODE-4330) (parent ticket)

[NODE-4394](https://jira.mongodb.org/browse/NODE-4394) (subtask)

Passing evergreen build of spec tests in driver: https://spruce.mongodb.com/version/62cf330c57e85a64e1f70d24/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

--- 

## What it does

This PR adds the key management API methods to the node bindings for libmongocrypt.  The corresponding tests are passing and unskipped in the driver (see https://github.com/mongodb/node-mongodb-native/pull/3322).  

Some additional notes:

- This PR also removes session support for rewrapManyDataKey (drivers ticket filed to add session support to the key management API https://jira.mongodb.org/browse/DRIVERS-2389)
- This PR regenerates the documentation for mongodb-client-encryption for the first time in a long time.  there are other differences in the readme unrelated to the key management API as a result.


bindings/node/lib/clientEncryption.js
```diff
@@ -28,6 +28,40 @@ module.exports = function (modules) {
...
+   * @property {boolean} acknowledged Indicates whether this write result was acknowledged. If not, then all other members of this result will be undefined.
+   * @property {number} deletedCount The number of documents that were deleted
+   */
+
+  /**
+   * @typedef {object} BulkWriteResult The BulkWriteResult class from the MongoDB Node driver (https://mongodb.github.io/node-mongodb-native/4.8/classes/BulkWriteResult.html)
+   */
+
+  /**
+   * @typedef {object} FindCursor The FindCursor class from the MongoDB Node driver (see https://mongodb.github.io/node-mongodb-native/4.8/classes/FindCursor.html)
```
baileympearson, line 62: jsdoc doesn't let us import types from the index.d.ts for documentation purposes (i.e., `import('../index.d.ts').FindCursor`), so I created dummy types for the purposes documentation.  mongodb and mognodb-client-encryption are peer dependencies, so no users will be using this without also having mongodb installed, I think linking to the mongodb docs is sufficient until we can convert this to TS and use TSDoc or figure out a better documentation solution.

bindings/node/lib/clientEncryption.js
```diff
@@ -135,7 +166,7 @@ module.exports = function (modules) {
      * @param {object} [options] Options for creating the data key
      * @param {AWSEncryptionKeyOptions|AzureEncryptionKeyOptions|GCPEncryptionKeyOptions} [options.masterKey] Idenfities a new KMS-specific key used to encrypt the new data key
      * @param {string[]} [options.keyAltNames] An optional list of string alternate names used to reference a key. If a key is created with alternate names, then encryption may refer to the key by the unique alternate name instead of by _id.
-     * @param {ClientEncryption~createDataKeyCallback} [callback] Optional callback to invoke when key is created
+     * @param {ClientEncryptionCreateDataKeyCallback} [callback] Optional callback to invoke when key is created
```
baileympearson, line 169: misc fixes to the existing types to fix vscode's type hinting

bindings/node/test/clientEncryption.test.js
```diff
@@ -42,7 +42,7 @@ describe('ClientEncryption', function () {
   }
 
   async function setup() {
-    client = new MongoClient('mongodb://localhost:27017/test', {
+    client = new MongoClient('mongodb://bob:pwd123@localhost:27017', {
```
baileympearson, line 45: `drop` requires authentication, so we need to set authentication on the server

bindings/node/test/clientEncryption.test.js
```diff
@@ -77,7 +77,7 @@ describe('ClientEncryption', function () {
 
     beforeEach(function () {
       if (requirements.SKIP_LIVE_TESTS) {
-        this.test.skipReason = `requirements.SKIP_LIVE_TESTS=${requirements.SKIP_LIVE_TESTS}`;
+        this.currentTest.skipReason = `requirements.SKIP_LIVE_TESTS=${requirements.SKIP_LIVE_TESTS}`;
```
baileympearson, line 80: to properly attach a skip reason, we need to attach it to the `currentTest`, not the `test` when in `before` blocks.

bindings/node/test/clientEncryption.test.js
```diff
@@ -592,222 +592,4 @@ describe('ClientEncryption', function () {
         });
     });
   });
```
baileympearson, line 594: These tests were duplicates of tests in the driver (https://github.com/mongodb/node-mongodb-native/blob/4aa9d9d1add0a469d3c3c16ad70e574830158db2/test/integration/client-side-encryption/client_side_encryption.prose.test.js#L1509).

More importantly, the spec states explicitly that these tests must not be run against a standalone server.  We don't have the infrastructure to run anything other than a replica set in libmongocrypt (at least in the node bindings).

Additionally, these tests are skipped in CI because we only run tests that do not require a live server in CI.

bindings/node/README.md
```diff
@@ -40,12 +59,57 @@ npm test
...
 </dl>
 
+## Constants
+
+<dl>
+<dt><a href="#kBuffers">kBuffers</a></dt>
+<dd></dd>
+<dt><a href="#kLength">kLength</a></dt>
+<dd></dd>
+</dl>
```
addaleax, line 69: These are tagged `@internal` now – should they be included here?

bindings/node/README.md
```diff
@@ -40,12 +59,57 @@ npm test
...
+## Functions
+
+<dl>
+<dt><a href="#decorateDecryptionResult">decorateDecryptionResult()</a></dt>
+<dd><p>Recurse through the (identically-shaped) <code>decrypted</code> and <code>original</code>
+objects and attach a <code>decryptedKeys</code> property on each sub-object that
+contained encrypted fields. Because we only call this on BSON responses,
+we do not need to worry about circular references.</p>
+</dd>
+</dl>
```
addaleax, line 80: This should be `@internal` too, I assume?

bindings/node/index.d.ts
```diff
@@ -1,8 +1,19 @@
 import type { Document, Binary } from 'bson';
-import type { MongoClient, BulkWriteResult, ClientSession } from 'mongodb';
+import type { MongoClient, BulkWriteResult, ClientSession, DeleteResult, FindCursor } from 'mongodb';
 
 export type ClientEncryptionDataKeyProvider = 'aws' | 'azure' | 'gcp' | 'local' | 'kmip';
 
+export interface DataKey {
+  _id: Binary;
```
addaleax, line 7: Just a suggestion and feel free to ignore, but for this one it could make sense to go one step further:

```suggestion
  _id: Binary & { sub_type: typeof Binary.SUBTYPE_UUID };
```

bindings/node/index.d.ts
```diff
@@ -1,8 +1,19 @@
...
+  version: number;
```
addaleax, line 8: I’m not seeing `version` being mentioned anywhere – am I missing something?

bindings/node/lib/clientEncryption.js
```diff
@@ -267,79 +327,239 @@ module.exports = function (modules) {
...
+        .db(dbName)
+        .collection(collectionName)
+        .find({}, { readConcern: { level: 'majority' } });
+    }
+
+    /**
+     * Finds a key in the keyvault with the specified key.
+     *
+     * @param {ClientEncryptionDataKeyId} id - the id of the document to delete.
+     * @returns {Promise<DataKey>} IReturns a promise that either resolves to a {@link DataKey} if a document matches the key or null if no documents
```
addaleax, line 415:
```suggestion
     * @returns {Promise<DataKey>} Returns a promise that either resolves to a {@link DataKey} if a document matches the key or null if no documents
```

bindings/node/lib/clientEncryption.js
```diff
@@ -267,79 +327,239 @@ module.exports = function (modules) {
...
+        tlsOptions: this._tlsOptions
       });
 
-      return promiseOrCallback(callback, cb => {
-        stateMachine.execute(this, context, (err, dataKey) => {
-          if (err) {
-            cb(err, null);
-            return;
-          }
+      const execute = promisify(stateMachine.execute.bind(stateMachine));
```
addaleax, line 333: :blue_heart:

bindings/node/lib/clientEncryption.js
```diff
@@ -267,79 +327,239 @@ module.exports = function (modules) {
...
+        .collection(collectionName)
+        .findOne({ _id }, { readConcern: { level: 'majority' } });
+    }
+
+    /**
+     * Finds a key in the keyvault which has the specified keyAltNames as a keyAltName.
+     *
+     * @param {string} keyAltName - a keyAltName to search for a key
+     * @returns {Promise<DataKey | null>} Returns a promise that either resolves to a {@link DataKey} if a document matches the key or null if no documents
+     * match the id.  The promise rejects with an error if an error is thrown.
```
addaleax, line 439:
```suggestion
     * match the keyAltName.  The promise rejects with an error if an error is thrown.
```

bindings/node/lib/clientEncryption.js
```diff
@@ -267,79 +327,239 @@ module.exports = function (modules) {
       const stateMachine = new StateMachine({
         bson,
         proxyOptions: this._proxyOptions,
-        tlsOptions: this._tlsOptions,
-        session: options.session
```
addaleax, line 271: I assume the idea is to add back support for this for all APIs in one go at some point?

bindings/node/index.d.ts
baileympearson: I got this schema directly from the spec - https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#key-vault-collection-schema-for-data-keys

This seems to be the first time we're returning a full data key entry, so I added the schema directly.  Do you think it would be better to only return a subset of the fields?

I did notice, on re-review of the schema in the spec, that version can be unset.  So version and keyAltNames should be optional, I'll make that change.

addaleax: Yeah, I think “Implicitly 0 if unset” is the important part – I was just surprised because I’d never seen a `version` field on a data key doc :slightly_smiling_face:

bindings/node/lib/clientEncryption.js
baileympearson: Yup, we have a drivers ticket for it - https://jira.mongodb.org/browse/DRIVERS-2389

Hopefully it gets picked up soon (maybe I'll talk to Jeff about doing it, it sounds relatively straightforward and interesting).

bindings/node/index.d.ts
baileympearson: Screen Shot 2022-07-15 at 10 27 18 AM

This seems to cause errors, I don't think our Binary type is typed well enough for this.  The `sub_type` is a runtime value on Binary and we don't have a way to relate the `number` in the Binary constructor to the proper subtype.

Separate issue - I noticed that I don't see any type errors in the `index.d.ts` in libmongocrypt.  I added `const foo: string = 3;` to the index and vscode didn't complain.  All the more reason to convert our bindings to Typescript I guess!

bindings/node/index.d.ts
baileympearson: Fixed 😄

nbbeeken (changes requested):

bindings/node/index.d.ts
```diff
@@ -411,52 +411,58 @@ export class ClientEncryption {
   ): Promise<ClientEncryptionRewrapManyDataKeyResult>;
 
   /**
+   * @experimental
```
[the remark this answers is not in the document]
addaleax: Cool, SGTM!

nbbeeken (approved):

dariakp (approved):

PR #415, opened 2022-07-14
PYTHON-3067 Add support for macOS m1 universal2 wheels

juliusgeo:


.evergreen/config.yml
```diff
@@ -774,6 +774,12 @@ buildvariants:
   - build-and-test-csharp
   - test-python
   - test-java
+- name: macos-arm64
+  display_name: "macOS (arm64) 11.0"
+  run_on: macos-1100-arm64
+  tasks:
+  - build-and-test-and-upload
+  - test-python
```
ShaneHarvey, line 782: Let’s add test-Python to the existing macos-1100-arm64 task below instead.

ShaneHarvey (changes requested):

.evergreen/config.yml
```diff
@@ -1087,13 +1087,14 @@ buildvariants:
   tasks:
   - name: debian-package-build
 - name: macos
-  display_name: macOS m1 (Apple LLVM)
+  display_name: macOS (arm64) 11.0
```
ShaneHarvey, line 1090: Let’s not change the name here

ShaneHarvey: If you feel strongly I'd suggest opening a MONGOCRYPT ticket for the rename but either way let's not do it in this PR since it affects other teams.

ShaneHarvey (approved):

PR #417, opened 2022-07-15
Java: Ensure mac universal libs are published

rozza:
MONGOCRYPT-458

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1123,8 +1123,7 @@ buildvariants:
   tasks:
   - build-and-test-and-upload
   - build-and-test-node
-  # MONGOCRYPT-459 - depends on BUILD-14976
-  # - test-java
+  - test-java
 - name: windows-vs2017-32bit
   # Test Windows 32 bit builds for PHPC. PHPC builds libmongocrypt from source. See MONGOCRYPT-391.
   display_name: "Windows VS 2017 32-bit compile"
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -398,9 +398,8 @@ tasks:
       name: build-and-test-and-upload
     - variant: macos_x86_64
       name: build-and-test-and-upload
-    # Needs: BUILD-14976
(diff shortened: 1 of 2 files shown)


rozza:
Updated to only include the universal build. 

Also added a copy of the uploaded artefacts to s3 with the tag name.

rozza:
All mac tests now pass and the [libmongocrypt-java.tar.gz](https://mciuploads.s3.amazonaws.com/libmongocrypt/java/3c4114631a1863751020684863101d0858fb1994/libmongocrypt-java.tar.gz) contains the macos artifacts.

jyemin:
Should there be a test-java task running on macOS m1 (Apple LLVM)?  I don't see one in the patch build.

.evergreen/config.yml
```diff
@@ -434,6 +441,17 @@ tasks:
         permissions: public-read
         local_file: 'libmongocrypt-java.tar.gz'
         content_type: '${content_type|application/x-gzip}'
+    - command: s3.put
```
jyemin, line 444: What does this command do if `libmongocrypt-java-${tag_upload_location}.tar.gz` does not exist (since it's created conditionally, above)?  I'm guessing some sort of error, but does it cause any sort of abort or global task failure?

rozza: This is copied from the libmongocrypt-all uploading and it should just log a message like it does with the libmongocrypt-all upload:

```
file '/data/mci/c2027b19a6d6d313601aa5939dfa35e0/libmongocrypt-all-.tar.gz' not found but skip missing true
```

rozza:
> Should there be a test-java task running on macOS m1 (Apple LLVM)? I don't see one in the patch build.

Yes it was down at the bottom with the default sorting: https://spruce.mongodb.com/version/62fe1fbba4cf4779f0513cd7/tasks?baseStatuses=all,failed,success,started&page=0&sorts=STATUS%3ADESC&variant=macos

jyemin (approved):
LGTM!

PR #419, opened 2022-07-18
MONGOCRYPT-449 Reject empty KMS providers if NEED_KMS_CREDENTIALS is not supported

kevinAlbs:
# Summary
- Reject empty KMS providers if consumer does not support `MONGOCRYPT_CTX_NEED_KMS_CREDENTIALS` state.

# Background & Motivation.

libmongocrypt supports setting KMS providers with an empty document (example: `{ "gcp": {} }`) in `mongocrypt_setopt_kms_providers`. That tells libmongocrypt to enter the `MONGOCRYPT_CTX_NEED_KMS_CREDENTIALS` before KMS credentials are needed.

Drivers must opt-in to supporting on-demand credentials by calling `mongocrypt_setopt_use_need_kms_credentials_state`. This signals the driver is prepared to handle the `MONGOCRYPT_CTX_NEED_KMS_CREDENTIALS` state.

If `mongocrypt_setopt_use_need_kms_credentials_state` is not called, there is no use-case for setting an empty document.

I do not expect this change to require driver changes. There are no unified specification tests using an empty document for a KMS provider configured after https://github.com/mongodb/specifications/pull/1267.

### DIFF
--- a/src/mongocrypt-opts.c
+++ b/src/mongocrypt-opts.c
@@ -109,7 +109,9 @@ _mongocrypt_opts_cleanup (_mongocrypt_opts_t *opts)
 
 bool
 _mongocrypt_opts_kms_providers_validate (
-   _mongocrypt_opts_kms_providers_t *kms_providers, mongocrypt_status_t *status)
+   _mongocrypt_opts_t *opts,
+   _mongocrypt_opts_kms_providers_t *kms_providers,
+   mongocrypt_status_t *status)
 {
    if (!kms_providers->configured_providers &&
        !kms_providers->need_credentials) {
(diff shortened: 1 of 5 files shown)


src/mongocrypt-opts.c
```diff
@@ -132,6 +134,12 @@ _mongocrypt_opts_kms_providers_validate (
       }
    }
 
+   if (kms_providers->need_credentials &&
+       !opts->use_need_kms_credentials_state) {
+      CLIENT_ERR ("on-demand credentials not supported");
```
eramongodb, line 139: Rather than "not supported", would "not enabled" be more accurate, given users have the option to opt into using on-demand credentials?

kevinAlbs: Yes. Done.

eramongodb (approved):

addaleax (approved):

PR #426, opened 2022-07-27
MONGOCRYPT-457 Use CRLF instead of LF newlines

kevinAlbs:
# Summary
- Use `\r\n`, not `\n`, as the end-of-line delimitter in HTTP requests.

The Go driver integration tests were run with this branch in a [patch build](https://spruce.mongodb.com/version/62e16a0c3e8e8618242ec912/) to test sending KMS requests to live services.

### DIFF
--- a/kms-message/src/kms_request.c
+++ b/kms-message/src/kms_request.c
@@ -762,6 +762,13 @@ kms_request_validate (kms_request_t *request)
    }
 }
 
+/* append_http_endofline appends an HTTP end-of-line marker: "\r\n". */
+static void
+append_http_endofline (kms_request_str_t *str)
+{
+   kms_request_str_append_chars (str, "\r\n", 2);
+}
+
 char *
 kms_request_get_signed (kms_request_t *request)
 {
@@ -795,7 +802,7 @@ kms_request_get_signed (kms_request_t *request)
    }
 
    kms_request_str_append_chars (sreq, " HTTP/1.1", -1);
-   kms_request_str_append_newline (sreq);
+   append_http_endofline (sreq);
 
    /* headers */
    lst = kms_kv_list_dup (request->header_fields);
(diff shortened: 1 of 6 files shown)


kms-message/src/kms_request.c
```diff
@@ -762,6 +762,13 @@ kms_request_validate (kms_request_t *request)
    }
 }
 
+/* append_http_endofline appends an HTTP end-of-line marker: "\r\n". */
+static void
+append_http_endofline (kms_request_str_t *str)
+{
+   kms_request_str_append_chars (str, "\r\n", 2);
+}
+
```
eramongodb, line 771: To confirm, the reason why a new `append_http_endofline` function added instead of changing `kms_request_str_append_newline` to append `\r\n` is because not all KMS request content is HTTP; is this correct?

kevinAlbs: Correct. KMIP does not use HTTP.

kms-message/test/test_kms_util.c
```diff
@@ -59,3 +60,30 @@ char *
...
+char *
+replace_all (const char *input, const char *match, const char *replacement)
+{
+   kms_request_str_t *replaced = kms_request_str_new ();
+   const char *start = input;
+   const char *iter = strstr (input, match);
+   while (iter != NULL) {
+      kms_request_str_append_chars (replaced, start, (ssize_t) (iter - start));
+      kms_request_str_append_chars (
+         replaced, replacement, strlen (replacement));
```
eramongodb, line 73:
```suggestion
      kms_request_str_append_chars (
         replaced, replacement, (ssize_t) strlen (replacement));
```

kms-message/test/test_kms_util.c
```diff
@@ -59,3 +60,30 @@ char *
 data_to_hex (const uint8_t *buf, size_t len) {
    return hexlify (buf, len);
 }
+
+char *
+replace_all (const char *input, const char *match, const char *replacement)
```
eramongodb, line 65: Suggest adding not-null assertions for all parameters.

eramongodb (approved):
Minor suggestions; otherwise LGTM.

kms-message/test/test_kms_util.c
kevinAlbs: Done.

PR #434, opened 2022-08-02
MONGOCRYPT-455 use release tag on the current commit, if there is one

rcsanchez97:


kevinAlbs (approved):
Thank you for the quick fix. LGTM

PR #439, opened 2022-08-09
feat(NODE-4517): add official support for the key management API

baileympearson:
NODE-4517 (part of NODE-4443)

This PR removes experimental tags from the key management API in preparation for the release of mongodb-client-encryption@2.2.0.

It also adds a doc comment for rewrapManyDataKey.

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -405,7 +405,7 @@ export class ClientEncryption {
   ): void;
 
   /** 
-   * Searches the keyvault for any data keys matching the provided filter.  If there are matches, the attempts to re-wrap the data keys using the provided options.
+   * Searches the keyvault for any data keys matching the provided filter.  If there are matches, rewrapManyDataKey then attempts to re-wrap the data keys using the provided options.
    *
    * If no matches are found, then no bulk write is performed.
    */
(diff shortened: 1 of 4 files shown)


dariakp (commented):
Do we have a "feat" PR for this feature in general that will go in this minor version? If not, perhaps we should use "feat" on this, to bring attention to the feature being officially supported now. In fact, perhaps we should adopt a convention where removing experimental tags is similar to deprecating/undeprecating and is always marked with a "feat".

addaleax (approved):

bindings/node/index.d.ts
```diff
@@ -404,46 +404,45 @@ export class ClientEncryption {
     callback: ClientEncryptionCreateDataKeyCallback
   ): void;
 
-  /** @experimental */
+  /** 
+   * Searches the keyvault for any data keys matching the provided filter.  If there are matches, the attempts to re-wrap the data keys using the provided options.
```
dariakp, line 408:
```suggestion
   * Searches the keyvault for any data keys matching the provided filter.  If there are matches, attempts to re-wrap the data keys using the provided options.
```

dariakp (changes requested):

bindings/node/index.d.ts
baileympearson: I didn't phrase it exactly like you suggested, but I did fix the typo.  That comment was also present in the `js` source file, so I fixed it there and regenerated the readme.

dariakp (approved):

PR #441, opened 2022-08-12
MONGOCRYPT-463 check the `collMod` command for `validator.$jsonSchema`

kevinAlbs:
# Summary
- Check the `collMod` command for `validator.$jsonSchema` and send it to query analysis.

# Background & Motivation

## Checking collMod for `validator.$jsonSchema`
Since MONGOCRYPT-429, automatic encryption has supported the `collMod` command. Prior to MONGOCRYPT-429, it was always an error to automatically encrypt a `collMod` command. A `collMod` command is not bypassed because it may include sensitive plaintext in a `validator` field:
```json
{
    "collMod": "encryptedCollection",
    "validator": {
        "encryptedField": "plaintext",
    }
}
```

A `collMod` command may also include a `validator.$jsonSchema`:
```json
{
    "collMod": "encryptedCollection",
    "validator": {
        "$jsonSchema": {
            "encryptedField": {
                "encrypt": {
                    "bsonType": "string",
                    "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic",
                    "keyId": {
                        "$binary": {
                            "base64": "AAAAAAAAAAAAAAAAAAAAAA==",
                            "subType": "04"
                        }
                    }
                }
            }
        }
    }
}
```

If a `collMod` includes a `validator.$jsonSchema`, Query Analysis (mongocryptd or the mongo_crypt shared library) expects the `jsonSchema` field added by libmongocrypt to match. See MONGOCRYPT-463 for an example of this error.

A regression specification test is tested with the Go driver [here](https://spruce.mongodb.com/version/62f67ca80305b9451042e293/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -1935,8 +1935,9 @@ _try_empty_schema_for_create (mongocrypt_ctx_t *ctx)
    return true;
 }
 
-/* _try_schema_from_create_cmd tries to find a JSON schema included in a create
- * command by checking for "validator.$jsonSchema". Example:
+/* _try_schema_from_create_or_collMod_cmd tries to find a JSON schema included
+ * in a create or collMod command by checking for "validator.$jsonSchema".
+ * Example:
  * {
  *     "create" : "coll",
  *     "validator" : {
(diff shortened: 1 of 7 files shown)


src/mongocrypt-ctx-encrypt.c
```diff
@@ -1946,19 +1947,23 @@ _try_empty_schema_for_create (mongocrypt_ctx_t *ctx)
  *     }
  * }
  * If the "create" command does not include a JSON schema, an empty JSON schema
- * is returned. This is to avoid an unnecessary 'listCollections' command for
- * create. */
+ * is later. This is to avoid an unnecessary 'listCollections' command for
```
vector-of-bool, line 1950: "is later"?

vector-of-bool (approved):
LGTM, but one odd comment

src/mongocrypt-ctx-encrypt.c
kevinAlbs: Fixed.

src/mongocrypt-ctx-encrypt.c
```diff
@@ -1947,18 +1948,22 @@ _try_empty_schema_for_create (mongocrypt_ctx_t *ctx)
  * }
  * If the "create" command does not include a JSON schema, an empty JSON schema
  * is returned. This is to avoid an unnecessary 'listCollections' command for
- * create. */
+ * create.
+ *
+ * If the "collMod" command does not include a JSON schema, a schema is later
+ * requested from other sources. This is because a "collMod" command may have
```
eramongodb, line 1954: Can "from other sources" be made more specific, e.g. mention the `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` state?

eramongodb (approved):
Minor suggestion; otherwise, LGTM.

src/mongocrypt-ctx-encrypt.c
kevinAlbs: Done.

PR #442, opened 2022-08-16
fix(NODE-4516): dont require a connected client

durran:
This removes the need to pass the bson module as an option to the auto encrypter or the state machine as it is now required directly.


dariakp (changes requested):
Per the slack conversation, let's revisit the approach to make sure we don't break users of the older versions of bson and also consider bson-ext

PR #443, opened 2022-08-17
fix(NODE-4556): attempt to use exported BSON

durran:
This updates the Node bindings to first attempt to use the exported `BSON` from the `mongodb` module before falling back to the old behaviour. This means that:

- Driver 4.9.0 and higher will use the exported `BSON` from the driver itself.
- Driver 4.x before 4.9.0 and driver 3.x will continue to use the `bson` object passed in the options or on `client.topology`

This allows driver 4.9.0 and higher to not need a connected client to be able to user csfle.

Main of the Node driver is now pinning on the latest commit in this PR and passing, running all live tests: https://spruce.mongodb.com/task/mongo_node_driver_next_ubuntu1804_custom_dependency_tests_run_custom_csfle_tests_latest_pinned_commit_40d485c292a14f8251a4e3724d1bb5f8cf11577c_22_08_17_20_53_00/tests?execution=0&sortBy=STATUS&sortDir=ASC

The patch build shows this works (since the tests pass and bson is no longer a dev dependency) but the dev dependency on mongodb will need to bump to the next release once https://github.com/mongodb/node-mongodb-native/pull/3367 is merged.

### DIFF
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -8,6 +8,7 @@ module.exports = function (modules) {
   const MongocryptdManager = require('./mongocryptdManager').MongocryptdManager;
   const MongoClient = modules.mongodb.MongoClient;
   const MongoError = modules.mongodb.MongoError;
+  const BSON = modules.mongodb.BSON;
   const cryptoCallbacks = require('./cryptoCallbacks');
 
   /**
@@ -91,7 +92,7 @@ module.exports = function (modules) {
      */
     constructor(client, options) {
       this._client = client;
-      this._bson = options.bson || client.topology.bson;
+      this._bson = options.bson || BSON || client.topology.bson;
       this._bypassEncryption = options.bypassAutoEncryption === true;
 
       this._keyVaultNamespace = options.keyVaultNamespace || 'admin.datakeys';
--- a/bindings/node/test/autoEncrypter.test.js
+++ b/bindings/node/test/autoEncrypter.test.js
@@ -2,10 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
-const BSON = require('bson');
-const EJSON = require('bson').EJSON;
 const sinon = require('sinon');
 const mongodb = require('mongodb');
+const BSON = mongodb.BSON;
+const EJSON = BSON.EJSON;
 const requirements = require('./requirements.helper');
 const MongoNetworkTimeoutError = mongodb.MongoNetworkTimeoutError || mongodb.MongoTimeoutError;
 const MongoError = mongodb.MongoError;
@@ -108,6 +108,103 @@ describe('AutoEncrypter', function () {
 
   afterEach(() => {
     sandbox.restore();
+  });
+
+  describe('#constructor', function () {
+    context('when mongodb exports BSON (driver >= 4.9.0)', function () {
+      context('when a bson option is provided', function () {
+        const bson = Object.assign({}, BSON);
+        const encrypter = new AutoEncrypter(
+          {},
+          {
+            bson: bson,
+            kmsProviders: {
+              local: { key: Buffer.alloc(96) }
+            }
+          }
+        );
+
+        it('uses the bson option', function () {
+          expect(encrypter._bson).to.equal(bson);
+        });
+      });
+
+      context('when a bson option is not provided', function () {
+        const encrypter = new AutoEncrypter(
+          {},
+          {
+            kmsProviders: {
+              local: { key: Buffer.alloc(96) }
+            }
+          }
+        );
+
+        it('uses the mongodb exported BSON', function () {
+          expect(encrypter._bson).to.equal(BSON);
+        });
+      });
+
+      it('never uses bson from the topology', function () {
+        expect(() => {
+          new AutoEncrypter(
+            {},
+            {
+              kmsProviders: {
+                local: { key: Buffer.alloc(96) }
+              }
+            }
+          );
+        }).not.to.throw();
+      });
+    });
+
+    context('when mongodb does not export BSON (driver < 4.9.0)', function () {
+      context('when a bson option is provided', function () {
+        const bson = Object.assign({}, BSON);
+        const encrypter = new AutoEncrypter(
+          {},
+          {
+            bson: bson,
+            kmsProviders: {
+              local: { key: Buffer.alloc(96) }
+            }
+          }
+        );
+
+        it('uses the bson option', function () {
+          expect(encrypter._bson).to.equal(bson);
+        });
+      });
+
+      context('when a bson option is not provided', function () {
+        const mongoNoBson = { ...mongodb, BSON: undefined };
+        const AutoEncrypterNoBson = require('../lib/autoEncrypter')({
+          mongodb: mongoNoBson,
+          stateMachine
+        }).AutoEncrypter;
+
+        context('when the client has a topology', function () {
+          const client = new MockClient();
+          const encrypter = new AutoEncrypterNoBson(client, {
+            kmsProviders: {
+              local: { key: Buffer.alloc(96) }
+            }
+          });
+
+          it('uses the bson on the topology', function () {
+            expect(encrypter._bson).to.equal(client.topology.bson);
+          });
+        });
+
+        context('when the client does not have a topology', function () {
+          it('raises an error', function () {
+            expect(() => {
+              new AutoEncrypterNoBson({}, {});
+            }).to.throw(/bson/);
+          });
+        });
+      });
+    });
   });
 
   it('should support `bypassAutoEncryption`', function (done) {
--- a/bindings/node/test/clientEncryption.test.js
+++ b/bindings/node/test/clientEncryption.test.js
@@ -2,8 +2,8 @@
 const fs = require('fs');
 const expect = require('chai').expect;
 const sinon = require('sinon');
-const BSON = require('bson');
 const mongodb = require('mongodb');
+const BSON = mongodb.BSON;
 const MongoClient = mongodb.MongoClient;
 const cryptoCallbacks = require('../lib/cryptoCallbacks');
 const stateMachine = require('../lib/stateMachine')({ mongodb });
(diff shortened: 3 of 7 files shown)


nbbeeken (approved):
LGTM

bindings/node/lib/autoEncrypter.js
```diff
@@ -91,7 +92,7 @@ module.exports = function (modules) {
      */
     constructor(client, options) {
       this._client = client;
-      this._bson = options.bson || client.topology.bson;
+      this._bson = BSON || options.bson || client.topology.bson;
```
dariakp, line 95: when would `modules.mongodb.BSON` actually be undefined? I'm a bit confused about the order of the fallbacks here, especially since in the test files below we are defining the import unconditionally as mongodb.BSON - why is it ok to always expect it to be defined in tests but not in this code?

also, can we use the one subtask per PR convention we've established to make review and merge tracking easier? it would also be good to make sure we have a subtask for bumping that dependency that's called out in the description

dariakp (changes requested):

bindings/node/test/autoEncrypter.test.js
```diff
@@ -2,10 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
-const BSON = require('bson');
-const EJSON = require('bson').EJSON;
 const sinon = require('sinon');
 const mongodb = require('mongodb');
+const BSON = mongodb.BSON;
```
baileympearson, line 7: Did you run the tests locally with `SKIP_LIVE_TESTS` set to false?  My hunch is that tests will fail.  

The mongodb dev dependency is 4.8.1, which doesn't export bson.  Some of the tests (which I believe are skipped when there is no server running, like in CI) create a mock client using the `BSON` variable and store it on the topology.

For the tests, `BSON` is undefined (because we're running with 4.8.1) so my guess is that those tests will fail and CI is currently passing because the tests are skipped.

bindings/node/lib/autoEncrypter.js
baileympearson: The tests are strange.  The behavior of the tests will (generally) depend on the version of the driver that the bindings are tested with because the fle classes are initialized with an instance of the mongodb module.  I think that the current setup is fine, because it does actually (for the most part, with some exceptions) test the behavior that we want.

We do generally only test with the version in dev dependencies in mongodb-client-encryption (now 4.8.1) and latest in the driver though.

There are other issues with the tests as well I believe, see my other comment

durran: `modules.mongodb.BSON` will always be undefined when using a driver version < 4.9.0 so we need to fallback first to the case where we passed it in the options then to the topology on the client which is the last resort. So 4.9.0 will solve the connected client requirement in the parent ticket but any version before 4.9.0 will still require it.

This ticket is now sub-task NODE-4556 and I've updated the description.

bindings/node/test/autoEncrypter.test.js
durran: The tests failing is expected and is my last comment in the description of the PR:

_The patch build shows this works (since the tests pass and bson is no longer a dev dependency) but the dev dependency on mongodb will need to bump to the next release once https://github.com/mongodb/node-mongodb-native/pull/3367 is merged._

dariakp: As long as the intent is to always run latest, I suppose this is fine. But in that case I'm a little concerned that we don't have any new tests explicitly testing the fallback order. I'm also wondering why we *don't* want to check for the passed in bson options first and only fall back to the exported definition last, since if it's defined, it should be fine, right? Or is this just to indicate that the standard moving forward will be the driver exported library and that we'll be getting rid of the ability to pass it in via options or reading it from the topology altogether?

I just wish we had a more clear documentation (via test) about the 3 different fallback cases.

dariakp: P.S. should we bump the dependency in this PR after we release 4.9.0?

durran: @dariakp Yes we should update to 4.9.0 but it won't make much difference as the live tests aren't running in this repo as @baileympearson pointed out. I updated the pinned commit in the Node driver to point at the latest commit in this PR, so when this merges the Node driver tests will cover the changes. I updated the description of this PR and here's again a link to that run passing:

https://spruce.mongodb.com/task/mongo_node_driver_next_ubuntu1804_custom_dependency_tests_run_custom_csfle_tests_latest_pinned_commit_40d485c292a14f8251a4e3724d1bb5f8cf11577c_22_08_17_20_53_00/tests?execution=0&sortBy=STATUS&sortDir=ASC

dariakp: Ok, but what about my other question above? (Regarding actually testing the three different fallback cases)

> As long as the intent is to always run latest, I suppose this is fine. But in that case I'm a little concerned that we don't have any new tests explicitly testing the fallback order. I'm also wondering why we don't want to check for the passed in bson options first and only fall back to the exported definition last, since if it's defined, it should be fine, right? Or is this just to indicate that the standard moving forward will be the driver exported library and that we'll be getting rid of the ability to pass it in via options or reading it from the topology altogether?
>
> I just wish we had a more clear documentation (via test) about the 3 different fallback cases.

baileympearson: Yeah, that's a good point.  Users who do upgrade to version 4.9.0 of the driver and mongodb-client-encryption@2.2.1 _and_ pass in bson as an option would find that suddenly mongodb-client-encryption doesn't use the bson library they provided anymore.  Technically that's a breaking change? 

As far as the ordering goes - it needs to check the topology last.  Otherwise, mongodb-client-encryption will always check the topology, even if the client isn't connected.  But I do believe the intent was that eventually we'd stop supporting the BSON option too and enforce that mongodb-client-encryption and the driver always have the same bson version (@nbbeeken @durran right?).  So I think the order should be options, BSON import, topology.

durran: I've updated to use options first, then BSON, then topology.

durran: @baileympearson @dariakp I've updated the unit tests to test all the bson permutations.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -110,6 +110,92 @@ describe('AutoEncrypter', function () {
     sandbox.restore();
   });
 
+  describe('#constructor', function () {
+    // This case is driver 4.9.0 and higher.
+    context('when mongodb exports BSON', function () {
+      context('when a bson option is provided', function () {
+        const bson = Object.assign({}, BSON);
```
durran, line 117: We clone the original `BSON` module and use the clone as the `bson` option so it does not have object equality and then matches.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -110,6 +110,92 @@ describe('AutoEncrypter', function () {
...
+          {},
+          {
+            kmsProviders: {
+              local: { key: Buffer.alloc(96) }
+            }
+          }
+        );
+
+        it('uses the mongodb exported BSON', function () {
+          expect(encrypter._bson).to.equal(BSON);
```
durran, line 144: When we provide no `bson` option we use the exported `BSON` from the driver.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -110,6 +110,92 @@ describe('AutoEncrypter', function () {
...
+    });
+
+    // This case is driver < 4.9.0.
+    context('when mongodb does not export BSON', function () {
+      context('when a bson option is provided', function () {
+        const bson = Object.assign({}, BSON);
+        const encrypter = new AutoEncrypter(
+          {},
+          {
+            bson: bson,
```
durran, line 156: We ensure here we use the `bson` option.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -110,6 +110,92 @@ describe('AutoEncrypter', function () {
...
+          }
+        );
+
+        it('uses the bson option', function () {
+          expect(encrypter._bson).to.equal(bson);
+        });
+      });
+
+      context('when a bson option is not provided', function () {
+        const mongoNoBson = { ...mongodb, BSON: undefined };
```
durran, line 169: This is to ensure `BSON` is not exported by the module we inject.

bindings/node/test/autoEncrypter.test.js
```diff
@@ -110,6 +110,92 @@ describe('AutoEncrypter', function () {
     sandbox.restore();
   });
 
+  describe('#constructor', function () {
+    // This case is driver 4.9.0 and higher.
```
baileympearson, line 114: optional - we could move the comment into the test description so it's available in test output and won't get out of sync with the tests

```typescript
context('when mongodb exports BSON (driver >= 4.9.0)', ... )
```

baileympearson: this would apply to the other context block in this file too

bindings/node/test/autoEncrypter.test.js
```diff
@@ -110,6 +110,92 @@ describe('AutoEncrypter', function () {
...
+            kmsProviders: {
+              local: { key: Buffer.alloc(96) }
+            }
+          }
+        );
+
+        it('uses the mongodb exported BSON', function () {
+          expect(encrypter._bson).to.equal(BSON);
+        });
+      });
```
baileympearson, line 146: Should we add a test confirming that it never uses the BSON defined on the topology?  I think we explicitly test this, because if we ever did accidentally start using the client defined on the topology again, it would throw since these tests pass in an empty object for the client.  But it might be nice to explicitly document it with a test, something like

```typescript
it('should never use the BSON defined on the topology', function() {
  const client = { topology: { BSON: 'bson' } };
  const encrypter = new AutoEncrypter(....)
  expect(BSON).not.to.equal('bson')
});
```

or maybe

```typescript
it('should never use the BSON defined on the topology', function() {
  expect(() => new AutoEncrypter({}, ....)).not.to.throw(<error>);
});

```

bindings/node/test/clientEncryption.test.js
```diff
@@ -62,6 +62,96 @@ describe('ClientEncryption', function () {
     return client.close();
   }
 
+  describe('#constructor', function () {
```
baileympearson, line 65: same comments as on the AuthEncrypter tests apply to this file

baileympearson (changes requested):

bindings/node/test/autoEncrypter.test.js
durran: I chose the later for both tests.

durran:
This has also been rebased/squashed.

baileympearson (approved):

baileympearson:
filed a flakey test ticket for socks5 tests and restarted the run - https://jira.mongodb.org/browse/NODE-4567

dariakp (approved):

nbbeeken (approved):

PR #446, opened 2022-08-18
MONGOCRYPT-456 migrate GetVersion.cmake to calc_release_version.py

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/62fad7fc3066155e80ff329a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I compared the version number calculated in the patch build with the version number calculated in the base commit.  There is a difference, but this difference is expected and brings the version calculated for libmongocrypt in line with the versions calculated for the C and C++ drivers.

Note that the `debian-package-build` task is failing.  This is expected when adding a new file in a patch build when that file is required early in the process.  Once this is merged and a waterfall build commences I will confirm that the task passes at that point.


.evergreen/config.yml
```diff
@@ -586,7 +584,8 @@ tasks:
           ./bin/pip install poster
           ./bin/pip install pycrypto
           cd ..
-          pkg_version=$mongocrypt_version
+          # Get the current version of libmongocrypt.
+          pkg_version="$(python etc/calc_release_version.py 2>&1)"
```
kevinAlbs, line 588:
```suggestion
          pkg_version="$(python etc/calc_release_version.py)"
```

Running with python 3.10.6 shows the following deprecation warnings:
```
python ./etc/calc_release_versionNoGit.py 
/Users/kevin.albertson/review/libmongocrypt-446/./etc/calc_release_versionNoGit.py:214: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
  version_loose = LooseVersion('0.0.0')
/Users/kevin.albertson/review/libmongocrypt-446/./etc/calc_release_versionNoGit.py:73: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
  version_loose = LooseVersion('0.0.0')
/Users/kevin.albertson/review/libmongocrypt-446/./etc/calc_release_versionNoGit.py:86: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
  new_version_loose = LooseVersion(release_tag_match.group('ver'))
1.5.0-alpha0
```

Redirecting stderr to stdout may cause deprecation warnings to be captured.

IMO this does not need to be resolved in this PR. But this may require implementing an alternative to `LooseVersion` in the future.

.evergreen/config.yml
```diff
@@ -605,7 +604,7 @@ tasks:
           set -o errexit
 
           CMAKE=/cygdrive/c/cmake/bin/cmake
-          mongocrypt_version="$($CMAKE -DRELEASE_BRANCH_REF=origin/r1.5 -P ./cmake/GetVersion.cmake 2>&1)"
+          mongocrypt_version="$(python etc/calc_release_version.py 2>&1)"
```
kevinAlbs, line 607:
```suggestion
          mongocrypt_version="$(python etc/calc_release_version.py)"
```

.evergreen/create-packages-and-repos.sh
```diff
@@ -49,6 +44,11 @@ else
   PYTHONPATH="$PYTHONPATH:$WORKDIR/src"
 fi
 
+# Get current version of libmongocrypt.
+pushd "$WORKDIR/libmongocrypt"
+  mongocrypt_version="$("$python" etc/calc_release_version.py 2>&1)"
```
kevinAlbs, line 49:
```suggestion
  mongocrypt_version="$("$python" etc/calc_release_version.py)"
```

.evergreen/debian_package_build.sh
```diff
@@ -48,7 +48,7 @@ sudo chroot ./unstable-chroot /bin/bash -c "(set -o xtrace && \
   cd /tmp/libmongocrypt && \
   git clean -fdx && \
   git reset --hard HEAD && \
-  cmake -DRELEASE_BRANCH_REF=origin/r1.5 -P ./cmake/GetVersion.cmake > VERSION_CURRENT 2>&1 && \
+  python3 etc/calc_release_version.py > VERSION_CURRENT 2>&1 && \
```
kevinAlbs, line 51:
```suggestion
  python3 etc/calc_release_version.py > VERSION_CURRENT && \
```

CMakeLists.txt
```diff
@@ -149,9 +149,32 @@ if (BUILD_VERSION STREQUAL "0.0.0")
...
+         if (NOT CALC_RELEASE_VERSION_RESULT STREQUAL 0)
+            # If python failed above, stderr would tell the user about it
+            message (FATAL_ERROR
+               "BUILD_VERSION not specified and could not be calculated\
+ (script invocation failed); specify in CMake command, -DBUILD_VERSION=<version>"
+            )
+         else ()
+            set (BUILD_VERSION ${CALC_RELEASE_VERSION})
+            message ("calculated BUILD_VERSION ${BUILD_VERSION}")
+            file (WRITE ${CMAKE_BINARY_DIR}/VERSION_CURRENT ${CALC_RELEASE_VERSION})
```
kevinAlbs, line 170:
```suggestion
            file (WRITE ${PROJECT_SOURCE_DIR}/VERSION_CURRENT ${CALC_RELEASE_VERSION})
```

To match the check on L148.

kevinAlbs (commented):
Looks good. I left suggestions to not capture stderr and to update the path to VERSION_CURRENT.

kevinAlbs:
Here are observed results from testing both scripts:

| git ref | GetVersion.cmake | calc_release_version.py |
|---|---|---|
| master | 1.6.0-20220822+gitc071d5a8d5 | 1.7.0-dev+20220822gitc071d5a8d5 |
| r1.5 | 1.5.3-20220822+git7b301eefcc | 1.5.3-dev+20220822git7b301eefcc |
| 1.5.2 | 1.5.2 | 1.5.2 |
| 1.5.0-alpha0 | 1.5.0-alpha0 | 1.5.0-alpha0| 

The difference in the prerelease marker is OK with me.

.evergreen/config.yml
rcsanchez97: @kevinAlbs I concur that we may need to think about alternatives to `LooseVersion`.  I ran into the warnings as well recently when working on syncing between the C and C++ driver.

CMakeLists.txt
rcsanchez97: @kevinAlbs So, I think I left a small gap here.  The `PROJECT_SOURCE_DIR` location was original to the C driver.  When implemented `calc_release_version.py` in the C++ driver, we allowed for `PROJECT_SOURCE_DIR`, but since it did not have a history of `VERSION_CURRENT` like the C driver, we preferred the `CMAKE_BINARY_DIR` location instead.  That way, clearing directory where the build system was generated would also eliminate the `VERSION_CURRENT` file.  I think that for libmonogocrypt we are best served preferring `CMAKE_BINARY_DIR`, so I'm proposing instead to do this:

``` set (BUILD_VERSION "0.0.0" CACHE STRING "Library version")
 if (BUILD_VERSION STREQUAL "0.0.0")
-   if (EXISTS ${PROJECT_SOURCE_DIR}/VERSION_CURRENT)
+   if (EXISTS ${CMAKE_BINARY_DIR}/VERSION_CURRENT)
+      file (STRINGS ${CMAKE_BINARY_DIR}/VERSION_CURRENT BUILD_VERSION)
+   elseif (EXISTS ${PROJECT_SOURCE_DIR}/VERSION_CURRENT)
       file (STRINGS ${PROJECT_SOURCE_DIR}/VERSION_CURRENT BUILD_VERSION)
       message (STATUS "File VERSION_CURRENT contained BUILD_VERSION ${BUILD_VERSION}")
    else ()
```

This would match the "better" approach we implemented in the C++ driver.

kevinAlbs: Using `CMAKE_BINARY_DIR` SGTM.

kevinAlbs (approved):
LGTM with the update to use `CMAKE_BINARY_DIR`.

PR #449, opened 2022-08-25
MONGOCRYPT-475 Implement Edge Generation algorithm

kevinAlbs:
# Summary

This is a translation of the Edge Generation algorithm from https://github.com/10gen/mongo/pull/6895. It will be used to generate the write payload for Queryable Encryption Range Index.

- Implement Edge Generation algorithm for `int32_t` and `int64_t`.
- Add resizable array type (`mc_array_t`).
- Move optional integer types separate header.
- Add helpers to create bit strings from integers.
- Add helpers to compute number of leading 0's in bit representation of integers.

# Background & Motivation

`mc_edges_t` is an opaque struct representing a list of edges. The motivation came from https://github.com/10gen/mongo/pull/6895#discussion_r943768671. The `mc_edges_t` could be optimized to not store the list of edges.

`mc_array_t` was copied from the C driver `mongoc_array_t`. It is used in place of the `std::vector<string>` of edges in the server implementation.


src/mc-range-edge-generation.c
```diff
@@ -0,0 +1,201 @@
...
+{
+   if (sparsity < 1) {
+      CLIENT_ERR ("sparsity must be 1 or larger");
+      return NULL;
+   }
+   mc_edges_t *edges = bson_malloc0 (sizeof (mc_edges_t));
+   edges->sparsity = sparsity;
+   _mc_array_init (&edges->edges, sizeof (char *));
+
+   for (size_t i = 1; i < strlen (leaf); i++) {
```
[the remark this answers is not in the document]
eramongodb: Suggest adding comment documenting why `i = 1` (leaf handled unconditionally below).

src/mc-range-edge-generation.c
```diff
@@ -0,0 +1,201 @@
...
+   char *root = bson_strdup ("root");
+   _mc_array_append_val (&edges->edges, root);
+
+   return edges;
+}
+
+const char *
+mc_edges_get (mc_edges_t *edges, size_t index)
+{
+   if (index > edges->edges.len - 1) {
```
eramongodb, line 63:
```suggestion
   if (index > edges->edges.len - 1u) {
```

src/mc-range-edge-generation.c
```diff
@@ -0,0 +1,201 @@
...
+   _mc_array_append_val (&edges->edges, leaf_copy);
+
+   char *root = bson_strdup ("root");
+   _mc_array_append_val (&edges->edges, root);
+
+   return edges;
+}
+
+const char *
+mc_edges_get (mc_edges_t *edges, size_t index)
```
eramongodb, line 61: Missing null check for `edges` parameter.

src/mc-range-edge-generation.c
```diff
@@ -0,0 +1,201 @@
...
+{
+   if (index > edges->edges.len - 1) {
+      return NULL;
+   }
+   return _mc_array_index (&edges->edges, char *, index);
+}
+
+size_t
+mc_edges_len (mc_edges_t *edges)
```
eramongodb, line 70: Missing null check for `edges` parameter.

eramongodb (changes requested):

src/mc-range-edge-generation-private.h
```diff
@@ -0,0 +1,85 @@
...
+#ifndef MC_RANGE_EDGE_GENERATION_PRIVATE_H
+#define MC_RANGE_EDGE_GENERATION_PRIVATE_H
+
+#include <stddef.h> // size_t
+#include <stdint.h>
+#include "mc-optional-private.h"
+#include "mongocrypt-status-private.h"
+
+// mc_edges_t represents a list of edges.
+typedef struct _mc_edges_t mc_edges_t;
```
vector-of-bool, line 26: Is it necessary that `mc_edges` have its definition hidden and implemented out-of-line?

vector-of-bool (approved):
Sorry for the delay on this one. I only have one question about the physical organization, but everything else appears good.

src/mc-range-edge-generation-private.h
kevinAlbs: Not necessary. I generally prefer limiting API in private headers unless it is an inconvenience.

PR #450, opened 2022-08-26
Support Low-Friction Test Creation Using Catch2

vector-of-bool:
Adding, removing, and defining test cases is a handful. One must add a file, then add that file to the CMake source listing, then declare a function to perform the install, call the function in the appropriate place, defined functions for each test case, then install those cases in the suite. We're also limited to primitive assertion types and "fail-immediately" behavior that halts all tests on the first failure, rather than enumerating all errors that _would occur_ in a full test execution.

While we expose a C API for consumers, we don't necessarily need to restrict ourselves within test cases, which can use the C API from C++ programs to utilize the increased ergonomics and productivity afforded by the language and supporting libraries - in this case, Catch2.

This changeset includes the following functionality:

- If `BUILD_TESTING` is enabled, and certain toolchain requirements are met, attempt to obtain and build a copy of Catch2 v3.1.0 as part of the build process.
- If we have Catch2 available, glob for all `*.test.cpp` files within the `src/` tree. This convention is chosen so as to enhance physical locality between source units and their corresponding unit tests. This globbing also alleviates the requirement that the test listing be manually updated. All recent versions of CMake support configure-depends-globbing, such that adding/removing files matching a globbing expression will re-execute the configure process. In situations in which configure-depends-globbing is not available, the globbing will still succeed but the automatic reloading will be unavailable. This will not be an issue in CI as we always run configure explicitly at the beginning of the build.
- Add all globbed `*.test.cpp` files to a single executable `test-runner`, which links to `Catch2WithMain`, which provides a `main()` definition that automatically registers every test case defined in every translation unit that was linked into the binary.
- Use `catch_discover_tests(test-runner)` to automatically define CTest test cases. For each `TEST_CASE()` invocation in all test translation units, this will define a separate CTest test that will have its output, timing, and failure condition isolated.

Additional considerations:

- Because Catch2 requires a C++ compiler with some amount of C++14 capability, Catch2 will not be available on some of our build targets and any Catch2-defined tests will not be built/run those systems. Nevertheless, the existing code and tests will still execute. Assuming that the tests added using Catch2 are not particularly environmentally sensitive, and assuming that the language and dependency libraries are Sufficiently Correct™️ on the older systems, we can then assume the passing of Catch2 tests on a modern system as sufficient assurance that the code is correct on an older system.


vector-of-bool:
This was a fun experiment, but the additional compilation time adds up fast across all tasks and variants, far more than to my liking, and ccache-hits are poor on Evergreen. May look into an alternative solution in the future.

PR #451, opened 2022-08-29
[MONGOCRYPT-473] Support "accessToken" as an Azure KMS option

vector-of-bool:
Refer: MONGOCRYPT-473

This changeset adds support for `{ azure: { accessToken: 'abc' } }` to be specified as a KMS provider. This will bypass the logic that attempts to obtain an access token from Azure and instead uses the given token directly. This is required to support Automatic KMS for Azure, as drivers need to be able to provide the automatically-obtained access token to libmongocrypt directly.

### DIFF
--- a/src/mongocrypt.c
+++ b/src/mongocrypt.c
@@ -1313,6 +1313,28 @@ _mongocrypt_parse_kms_providers (
             return false;
          }
 
+         if (!_mongocrypt_parse_optional_utf8 (
+                &as_bson,
+                "azure.accessToken",
+                &kms_providers->azure.access_token,
+                status)) {
+            return false;
+         }
+
+         if (kms_providers->azure.access_token) {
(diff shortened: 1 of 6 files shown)


src/mongocrypt.c
```diff
@@ -1313,6 +1313,28 @@ _mongocrypt_parse_kms_providers (
             return false;
          }
 
+         if (!_mongocrypt_parse_optional_utf8 (
+                &as_bson,
+                "azure.accessToken",
```
kevinAlbs, line 1318: Suggest adding a testcase to `_test_setopt_kms_providers` to validate parsing:

```c
{"{'azure': {'accessToken': 'foobar'}}", NULL, NULL, false}
```

kevinAlbs (approved):
LGTM with an optional test.

PR #452, opened 2022-08-31
PYTHON-3256 Update error message

blink1073:


bindings/python/pymongocrypt/mongocrypt.py
```diff
@@ -804,7 +804,7 @@ def _ask_for_kms_credentials(kms_providers):
         return
     if not _HAVE_AUTH_AWS:
         raise RuntimeError(
-            "MONGODB-AWS authentication requires pymongo-auth-aws: "
+            "On-demand AWS credentials require pymongo-auth-aws: "
             "install with: python -m pip install 'pymongo[aws]'"
```
ShaneHarvey, line 808: "[aws]" -> "[encryption]"

ShaneHarvey (changes requested):

bindings/python/pymongocrypt/mongocrypt.py
blink1073: I was asuuming you'd only hit this if using a version of pymongo that didn't include `pymongo_aws_auth` in the `encryption` extra.

ShaneHarvey: Oh good point.

ShaneHarvey (approved):

PR #455, opened 2022-09-01
feat(NODE-4604): add fallback to find kms credentials

durran:
Adds the ability to fallback to another callback in the Node driver if the user provided callback `onKmsProviderRefresh` does not return any credentials.

- [x] Add new option `onEmptyKmsProviders` to both `AutoEncrypter` and `ClientEncryption`
- [x] Refactor `askForKMSCredentials()` to handle the user provided callback first, then the new option.
- [x] Move duplicate `askForKMSCredentials()` into common module.
- [x] Use the common `askForKMSCredentials()` in the `StateMachine`
- [x] Add new unit tests for all possible combinations.
- [x] Mark new option as internal.

### DIFF
--- a/bindings/node/lib/clientEncryption.js
+++ b/bindings/node/lib/clientEncryption.js
@@ -113,11 +113,12 @@ module.exports = function (modules) {
       // kmsProviders will be parsed by libmongocrypt, must be provided as BSON binary data
       if (options.kmsProviders && !Buffer.isBuffer(options.kmsProviders)) {
         options.kmsProviders = this._bson.serialize(options.kmsProviders);
-      } else if (!options.onKmsProviderRefresh) {
+      } else if (!options.onKmsProviderRefresh && !options.onEmptyKmsProviders) {
         throw new TypeError('Need to specify either kmsProviders ahead of time or when requested');
       }
 
       this._onKmsProviderRefresh = options.onKmsProviderRefresh;
+      this._onEmptyKmsProviders = options.onEmptyKmsProviders;
       this._keyVaultNamespace = options.keyVaultNamespace;
       this._keyVaultClient = options.keyVaultClient || client;
       this._mongoCrypt = new mc.MongoCrypt(options);
@@ -675,17 +676,6 @@ module.exports = function (modules) {
           cb(null, result.v);
         });
       });
-    }
-
-    /**
-     * Ask the user for KMS credentials.
-     *
-     * This returns anything that looks like the kmsProviders original input
-     * option. It can be empty, and any provider specified here will override
-     * the original ones.
-     */
-    async askForKMSCredentials() {
-      return this._onKmsProviderRefresh ? this._onKmsProviderRefresh() : {};
     }
   }
 
--- a/bindings/node/lib/common.js
+++ b/bindings/node/lib/common.js
@@ -104,7 +104,9 @@ async function askForKMSCredentials(encrypter) {
   // First attempt to use the user provided callback to get the credentials.
   const creds = encrypter._onKmsProviderRefresh ? await encrypter._onKmsProviderRefresh() : {};
   // If the credentials are empty and we have an additional callback provided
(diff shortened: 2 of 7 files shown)


durran:
Reasoning behind making this a callback provided to `AutoEncrypter` and `ClientEncryption` as an option:

- Does not require that the driver export anything for internal csfle use.
- Is flexible in that the driver can refactor internals at will as long it passes the callback in the format csfle expects.
- Does not require linking PRs from both repos together in order to test.
- Is easily unit tested here in this repo.

bindings/node/lib/common.js
```diff
@@ -91,7 +91,29 @@ function promiseOrCallback(callback, fn) {
...
+ * the original ones.
+ *
+ * @param {AutoEncrypter|ClientEncryption} encrypter The encrypter.
+ */
+async function askForKMSCredentials(encrypter) {
+  // First attempt to use the user provided callback to get the credentials.
+  const creds = encrypter._onKmsProviderRefresh ? encrypter._onKmsProviderRefresh() : {};
+  // If the credentials are empty and we have an additional callback provided
+  // by the driver to attempt to fetch them, attempt to use that.
+  if (!creds.aws && encrypter._onEmptyKmsProviders) {
```
durran, line 108: This currently checks for `aws` specifically but could be made more generic in the future for other KMS providers.

bindings/node/lib/clientEncryption.js
```diff
@@ -113,11 +113,12 @@ module.exports = function (modules) {
       // kmsProviders will be parsed by libmongocrypt, must be provided as BSON binary data
       if (options.kmsProviders && !Buffer.isBuffer(options.kmsProviders)) {
         options.kmsProviders = this._bson.serialize(options.kmsProviders);
-      } else if (!options.onKmsProviderRefresh) {
+      } else if (!options.onKmsProviderRefresh && !options.onEmptyKmsProviders) {
         throw new TypeError('Need to specify either kmsProviders ahead of time or when requested');
       }
 
       this._onKmsProviderRefresh = options.onKmsProviderRefresh;
+      this._onEmptyKmsProviders = options.onEmptyKmsProviders;
```
addaleax, line 121: What’s the difference between these two?

bindings/node/lib/common.js
```diff
@@ -91,7 +91,29 @@ function promiseOrCallback(callback, fn) {
...
+ *
+ * This returns anything that looks like the kmsProviders original input
+ * option. It can be empty, and any provider specified here will override
+ * the original ones.
+ *
+ * @param {AutoEncrypter|ClientEncryption} encrypter The encrypter.
+ */
+async function askForKMSCredentials(encrypter) {
+  // First attempt to use the user provided callback to get the credentials.
+  const creds = encrypter._onKmsProviderRefresh ? encrypter._onKmsProviderRefresh() : {};
```
addaleax, line 105:
```suggestion
  const creds = encrypter._onKmsProviderRefresh ? await encrypter._onKmsProviderRefresh() : {};
```

bindings/node/lib/common.js
addaleax: I guess it might be worth adding a comment as to _why_ we single out AWS for now?

bindings/node/lib/clientEncryption.js
durran: `onKmsProviderRefresh` is the user supplied function to populate the KMS providers on demand if they want. `onEmptyKmsProviders` is what the driver will set as the last fallback to pull credentials from the environment if 1) They didn't provide them to the mongo client and 2) the user provided on demand callback was not provided or returned empty results.

bindings/node/lib/common.js
durran: Note on AWS added.

bindings/node/lib/clientEncryption.js
durran: Ah that's a really good idea - so much better than this.

PR #456, opened 2022-09-07
MONGOCRYPT-469 improved and more consistent null pointer checks

rcsanchez97:
This is mean to preview the proposed sorts of changes in a subset of the code-base.  Assuming this looks good to everyone, I will proceed with similar changes under the `src/` directory (excepting examples and tests).

Evergreen patch build: https://spruce.mongodb.com/version/6317a41030661513b86bf30a/tasks?limit=100&page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/crypto/libcrypto.c
+++ b/src/crypto/libcrypto.c
@@ -75,6 +75,9 @@ _encrypt_with_cipher (const EVP_CIPHER *cipher, aes_256_args_t args)
 
    ctx = EVP_CIPHER_CTX_new ();
 
+   BSON_ASSERT (args.key);
+   BSON_ASSERT (args.in);
+   BSON_ASSERT (args.out);
    BSON_ASSERT (ctx);
    BSON_ASSERT (cipher);
    BSON_ASSERT (NULL == args.iv ||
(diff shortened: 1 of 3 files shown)


eramongodb (approved):
I believe `BSON_ASSERT` should be used when asserting a member of a parameter is not-null prior to first usage (e.g. `BSON_ASSERT (args.in);`) instead of `BSON_ASSERT_PARAM` which is intended to be used specifically for pointer-type parameters, not members thereof. Otherwise, proposed changes LGTM.

src/crypto/libcrypto.c
```diff
@@ -139,6 +142,10 @@ _decrypt_with_cipher (const EVP_CIPHER *cipher, aes_256_args_t args)
 
    ctx = EVP_CIPHER_CTX_new ();
 
+   BSON_ASSERT (args.iv);
```
kevinAlbs, line 145:
```suggestion
   BSON_ASSERT_PARAM (cipher);
   BSON_ASSERT (args.iv);
```

src/crypto/libcrypto.c
```diff
@@ -213,6 +220,11 @@ _hmac_with_hash (const EVP_MD *hash,
                  _mongocrypt_buffer_t *out,
                  mongocrypt_status_t *status)
 {
+   BSON_ASSERT_PARAM (key);
```
kevinAlbs, line 223:
```suggestion
   BSON_ASSERT_PARAM (hash);
   BSON_ASSERT_PARAM (key);
```

kevinAlbs (approved):
LGTM with two more asserts.

rcsanchez97:
@eramongodb thanks for pointing that out.  I'll blame that on some
sloppy copy-pasta on my part.  I've made the changes you suggested.

vector-of-bool (approved):

PR #460, opened 2022-09-09
MONGOCRYPT-477 Update `FLE2InsertUpdatePayload` for range algorithm

kevinAlbs:
# Summary
- Support parsing `FLE2EncryptionPlaceholder` for range encrypted values.
- Transform `FLE2EncryptionPlaceholder` to `FLE2InsertUpdatePayload` for range encrypted values.

## Additional Changes
- Add `_mongocrypt_buffer_from_string`.
- Add `mc_bson_type_to_string`.
- Reorder edges in output of `mc_edges_new`.

# Background & Motivation

The server IDL is a good reference for the payload updates:
- [`FLE2EncryptionPlaceholder` was extended](https://github.com/mongodb/mongo/blob/d870dda33fb75983f628636ff8f849c7f1c90b09/src/mongo/crypto/fle_field_schema.idl#L133) for the range algorithm:
  - The sparsity value `"s"` is included. It is used to generate edges.
  - The value `"v"` is the document of the form `FLE2RangeSpec` (for find) or `FLE2RangeInsertSpec` (for insert).
- [`FLE2InsertUpdatePayload` was extended](https://github.com/mongodb/mongo/blob/d870dda33fb75983f628636ff8f849c7f1c90b09/src/mongo/crypto/fle_field_schema.idl#L192) for the range algorithm to include a list of edges `"g"` to insert.

Reordering edges in the output of `mc_edges_new` was done to match the order produce in the server. This permitted exact matching when comparing the FLE2InsertUpdatePayload generated by libmongocrypt with the FLE2InsertUpdatePayload generated by the server.

The lifetime of buffers in `mc_FLE2InsertUpdatePayload_serializeForRange` is hard to follow IMO. But is consistent with `mc_FLE2InsertUpdatePayload_serialize`. I am open to changing.


src/mc-fle2-encryption-placeholder.c
```diff
@@ -173,3 +191,122 @@ mc_validate_contention (int64_t contention, mongocrypt_status_t *status)
...
+   bool has_min = false, has_minIncluded = false, has_max = false,
+        has_maxIncluded = false;
+
+   if (!BSON_ITER_HOLDS_DOCUMENT (&iter)) {
+      CLIENT_ERR ("invalid FLE2RangeSpec: must be an iterator to a document");
+      return false;
+   }
+   bson_iter_recurse (&iter, &iter);
+
+   while (bson_iter_next (&iter)) {
```
vector-of-bool, line 220: BSON DSL would be very useful for these parsing loops :)

src/mongocrypt-marking.c
```diff
@@ -564,6 +566,276 @@ _mongocrypt_fle2_placeholder_to_insert_update_ciphertext (
...
+// get_edges creates and returns edges from an FLE2RangeInsertSpec. Returns NULL
+// on error.
+static mc_edges_t *
+get_edges (mc_FLE2RangeInsertSpec_t *insertSpec,
+           size_t sparsity,
+           mongocrypt_status_t *status)
+{
+   bson_type_t value_type = bson_iter_type (&insertSpec->v);
+
+#define GET_EDGES(BITS)                                                       \
```
vector-of-bool, line 578: I would rename to `GET_AND_RETURN_EDGES()` (since it returns) (and possible wrap with `if (1) { ... } else ((void)0)`)

src/mc-fle2-insert-update-payload-private.h
```diff
@@ -68,10 +70,25 @@ typedef struct {
...
 } mc_FLE2InsertUpdatePayload_t;
 
+/**
+ * EdgeTokenSet is the following BSON document:
+ * d: <binary> // EDCDerivedFromDataTokenAndCounter
+ * s: <binary> // ESCDerivedFromDataTokenAndCounter
+ * c: <binary> // ECCDerivedFromDataTokenAndCounter
+ * p: <binary> // Encrypted Tokens
+ */
+typedef struct {
```
vector-of-bool, line 85: Could this struct instance be inserted into `mc_FLE2InsertUpdatePayload_t`, since these tokens seem to travel around with each other in groups most of the time.

vector-of-bool (approved):
My only concern is the frequent utterance of `edc`, `ecc`, `esc` is becoming worrisome in terms of confusion and typos. Otherwise LGTM.

src/mc-fle2-encryption-placeholder.c
vector-of-bool: Uggh. Github for VSCode duplicated my comments over here? Ignore them.

kevinAlbs:
> My only concern is the frequent utterance of `edc`, `ecc`, `esc` is becoming worrisome in terms of confusion and typos. Otherwise LGTM.

I agree. Though I do not have an alternative proposal currently.

PR #463, opened 2022-09-14
MONGOCRYPT-484 Support case-insensitive values for "algorithm" and "queryType"

kevinAlbs:
# Summary

- Support case-insensitive values for "algorithm" and "queryType"
- Add `mstr_eq_ignore_case`

# Background & Motivation

This is a goal in [DBX Scope: Driver Support for Range Index](https://docs.google.com/document/d/1kkiqr91-HPpr0lDCm1lDVWSvRCuDKKGgdmwYgi2OA7A/edit), requested by product management.

### DIFF
--- a/src/mlib/str.h
+++ b/src/mlib/str.h
@@ -607,6 +607,7 @@ mstr_eq_ignore_case (mstr_view left, mstr_view right)
       return false;
    }
    return _mstr_strncasecmp (left.data, right.data, left.len) == 0;
+#undef _mstr_strncasecmp
 }
 
 /// Determine whether the given character is an printable ASCII codepoint
--- a/src/mlib/str.h
+++ b/src/mlib/str.h
@@ -11,6 +11,10 @@
 #include <inttypes.h>
 #include <stdbool.h>
 
+#if defined(MLIB_HAVE_STRINGS_H)
+#include <strings.h> /* For strncasecmp. */
+#endif
+
 /**
  * @brief A simple non-owning string-view type.
  *
@@ -585,6 +589,29 @@ mstr_eq (mstr_view left, mstr_view right)
       return false;
    }
    return memcmp (left.data, right.data, left.len) == 0;
+}
+
+/**
+ * @brief Determine whether two strings are equivalent ignoring case.
+ */
+static inline bool
+mstr_eq_ignore_case (mstr_view left, mstr_view right)
+{
+#ifdef _WIN32
+#define _mstr_strncasecmp _strnicmp
+#else
+#define _mstr_strncasecmp strncasecmp
+#endif
+
+   int got;
+
+   if (left.len != right.len) {
+      return false;
+   }
+   got = _mstr_strncasecmp (left.data, right.data, left.len) == 0;
+
+#undef _mstr_strncasecmp
+   return got;
 }
 
 /// Determine whether the given character is an printable ASCII codepoint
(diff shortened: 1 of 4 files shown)


src/mlib/str.h
```diff
@@ -587,6 +591,29 @@ mstr_eq (mstr_view left, mstr_view right)
...
+
+   int got;
+
+   if (left.len != right.len) {
+      return false;
+   }
+   got = _mstr_strncasecmp (left.data, right.data, left.len) == 0;
+
+#undef _mstr_strncasecmp
+   return got;
```
vector-of-bool, line 614: You can return the result directly without `got`

test/test-mongocrypt-ctx-setopt.c
```diff
@@ -68,6 +68,11 @@ static char invalid_utf8[] = {(char) 0x80, (char) 0x00};
    ASSERT_FAILS (                            \
       mongocrypt_ctx_setopt_algorithm (ctx, algo, algo_len), ctx, msg);
 
+#define QUERY_TYPE_OK(qt, qt_len) \
+   ASSERT_OK (mongocrypt_ctx_setopt_query_type (ctx, qt, qt_len), ctx);
+#define QUERY_TYPE_FAILS(qt, qt_len, msg) \
+   ASSERT_FAILS (mongocrypt_ctx_setopt_query_type (ctx, qt, qt_len), ctx, msg);
```
vector-of-bool, line 74: Prefix with `ASSERT_`

src/mlib/str.h
```diff
@@ -11,6 +11,10 @@
 #include <inttypes.h>
 #include <stdbool.h>
 
+#if defined(MLIB_HAVE_STRINGS_H)
+#include <strings.h> /* For strncasecmp. */
+#endif
```
vector-of-bool, line 16: Can we know ahead-of-time whether `<strings.h>` is available? (i.e. by checking the host platform instead of a configure-time check)

vector-of-bool (changes requested):
Just a few minor comments.

src/mlib/str.h
kevinAlbs: Done.

test/test-mongocrypt-ctx-setopt.c
kevinAlbs: Done.

src/mlib/str.h
kevinAlbs: I am not sure. Checking for `strings.h` is [consistent with libbson](https://github.com/mongodb/mongo-c-driver/blob/671a15154f0dd0e4af3c8df2ac08dfe4acf01795/src/libbson/CMakeLists.txt#L88-L91).

src/mlib/str.h
```diff
@@ -587,6 +591,24 @@ mstr_eq (mstr_view left, mstr_view right)
...
+#ifdef _WIN32
+#define _mstr_strncasecmp _strnicmp
+#else
+#define _mstr_strncasecmp strncasecmp
+#endif
+
+   if (left.len != right.len) {
+      return false;
+   }
+   return _mstr_strncasecmp (left.data, right.data, left.len) == 0;
```
vector-of-bool, line 609: The `undef` of `_mstr_stcncasecmp` is missing.

vector-of-bool (approved):
LGTM, minus one comment.

src/mlib/str.h
kevinAlbs: Oops, good eye.

PR #464, opened 2022-09-16
feat(NODE-4328): expose libmongocrypt version in bindings

addaleax:
Decided to expose this both on `AutoEncrypter` and `ClientEncryption` for now, but happy to move it to another place altogether if there’s a strong preference.

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -404,7 +404,7 @@ export class ClientEncryption {
     callback: ClientEncryptionCreateDataKeyCallback
   ): void;
 
-  /** 
+  /**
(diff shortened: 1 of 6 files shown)


baileympearson:
@addaleax mongodb-client-encryption doesn't release from a libmongocrypt version tag (for better or worse) and instead bundles libmongocrypt master when we release.  Would the version just point to the last version of libmongocrypt, even if we're commits ahead of it?

addaleax:
@baileympearson It would report the same version that libmongocrypt also reports for that given commit, which is always in the format `x.y.z-<pre>+<date>+git<sha>`. For example, for this PR:

```js
> require('./').AutoEncrypter.libmongocryptVersion
'1.6.1-20220927+git3cf3d0b166'
```

baileympearson (approved):

nbbeeken (approved):

PR #465, opened 2022-09-20
When executing git, we must be in a repository.

vector-of-bool:
One could also use the preferred "-C" argument, but we use older versions of Git that may not yet support that option.

This fixes main-project build breakages following merge of #453

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -25,12 +25,8 @@ functions:
           . libmongocrypt/.evergreen/init.sh
           bash "$EVG_DIR/print-env-info.sh"
 
-          if [ "${is_patch}" != "true" ]; then
-            # determine if we have a release tag present on HEAD
-            head_tag=$(git tag -l --points-at HEAD '[0-9].*')
-          else
-            head_tag=""
-          fi
+          # determine if we have a release tag present on HEAD
+          head_tag=$(run_chdir libmongocrypt/ git tag -l --points-at HEAD '[0-9].*' || true)
           echo "tag_upload_location: $head_tag"
 
           echo "tag_upload_location: \"$head_tag\"" > tag_expansion.yml
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -27,9 +27,12 @@ functions:
 
           # determine if we have a release tag present on HEAD
           head_tag=$(run_chdir libmongocrypt/ git tag -l --points-at HEAD '[0-9].*' || true)
-          echo "tag_upload_location: $head_tag"
-
-          echo "tag_upload_location: \"$head_tag\"" > tag_expansion.yml
+          use_tag=""
+          if test "${is_patch}" != "true"; then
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):

rcsanchez97 (commented):
I am curious why you chose to eliminate the conditional.  The original intent was essentially "ensure that when we upload to a URL based on the tag name we only do so for waterfall builds".  This has the effect of limiting uploads to a release tag-like location to a single upload for the waterfall build of the commit bearing the particular release tag.  It feels like your change will allow for uploading to URLs based on tag names that may result in collisions.  Suppose that someone submits a patch build or a PR based on commit `12c5118944295599097d5a70a11bb32a1b079282`.  That commit is tagged `1.6.0` and patch/PR builds based on it would now try to upload to a URL based on 1.6.0 tag.  It is possible that the artifacts created in this subsequent build might overwrite the original artifacts.  Even if they end up in a slightly different location (suppose because we change the upload target directory for patch/PR builds) then it will be confusing for users because we would have multiple sets of artifacts claiming to be associated with version 1.6.0 in different places.

vector-of-bool:
@rcsanchez97 do we expect to ever have tags on patch builds? I removed it because I assumed patches don't get release tags. It's simple enough to restore the patch logic, but my assumptions may have been faulty.

rcsanchez97:
The problem is that a patch build occurs within a repository that is essentially "(the state of the code at the base commit + the changes in the patch/PR)".  As a result, if the base commit has a tag associated with it, then the `git` command in this instance will see that tag as associated with `HEAD` and we will upload to it.

Here is the `upload-all` task of a patch build that I ran based on the `1.6.0-alpha0` commit: https://spruce.mongodb.com/task/libmongocrypt_publish_snapshot_upload_all_patch_aa91522f9b63cb09c73a84653aba4209056eeb2a_632a0a670305b94738808d11_22_09_20_18_45_59/files?execution=0

If you look at the files that were uploaded, it definitely created an and uploaded and artifact based on the 1.6.0-alpha0 tag.  In fact, that upload overwrote the original artifact based on the `1.6.0-alpha0` tag.  (I really hope nobody is still depending on that nearly two months after it was first built.)

All I did in the patch build is make the `head_tag` variable definition unconditional as in your change.  I also dropped most of the task dependencies on `upload-all` to speed up the process.  In any event, making the git invocation unconditional seems like it opens a dangerous possibility and I would strongly recommend keeping it conditional as it originally was.  The other part of the change, the `run_chdir` invocation and the addition of `|| true` for robustness LGTM.

rcsanchez97:
FYI, I re-ran the `upload-all` task on the base commit in the event that anyone wanted the real `1.6.0-alpha0` artifact rather than the one created for my patch build test.  If you try to download the artifact now it will have all of the necessary and expected contents.

rcsanchez97 (approved):
LGTM

PR #471, opened 2022-10-05
MONGOCRYPT-486 support BSON date for range algorithm

kevinAlbs:
# Summary

- Support BSON date for range insert and find

# Background & Summary

Required small updates. BSON date is represented as int64 internally.

Test data was obtained by printing output from server unit tests:
https://github.com/kevinAlbs/mongo/blob/SERVER-69013.range_support/kevinAlbs/range-find-output.txt


vector-of-bool (approved):

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -2390,6 +2390,17 @@ _test_encrypt_fle2_insert_range_payload_int64 (_mongocrypt_tester_t *tester)
 }
 #undef RNG_DATA
 
+#include "./data/fle2-insert-range/date/RNG_DATA.h"
+static void
+_test_encrypt_fle2_insert_range_payload_date (_mongocrypt_tester_t *tester)
+{
+   _test_rng_data_source source = {
+      .buf = {.data = (uint8_t *) RNG_DATA, .len = sizeof (RNG_DATA) - 1}};
```
eramongodb, line 2398: May be worth storing `RNG_DATA` in local array rather than casting to `uint8_t*` to avoid possibility of UB on modification of a string literal.

```suggestion
   uint8_t rng_data[] = RNG_DATA;
   _test_rng_data_source source = {
      .buf = {.data = rng_data, .len = sizeof (rng_data) - 1u}};
```

src/mongocrypt-marking.c
```diff
@@ -606,41 +606,45 @@ get_edges (mc_FLE2RangeInsertSpec_t *insertSpec,
...
    else if (value_type == BSON_TYPE_INT64) {
-      GET_AND_RETURN_EDGES (64);
+      GET_AND_RETURN_EDGES (64, int64);
+   }
+
+   else if (value_type == BSON_TYPE_DATE_TIME) {
+      GET_AND_RETURN_EDGES (64, date_time);
    }
 
 #undef GET_EDGES
```
eramongodb, line 650: This `#undef` should be fixed to:
```c
#undef GET_AND_RETURN_EDGES
```

eramongodb (approved):

PR #472, opened 2022-10-06
MONGOCRYPT-480 Support explicit encryption for Range

kevinAlbs:
# Summary

- Support explicit encryption with Range payloads.
   - Add `mongocrypt_ctx_setopt_range`.
- Test explicit decryption with Range payloads.
- Make `edgesInfo` optional on `mc_FLE2RangeFindSpec_t` and `payload` optional in `mc_FLE2FindRangePayload_t`.

## Other improvements

- Add `mc_iter_document_as_bson` to simplify iterating a `bson_iter_t` into a `bson_t`.
- Use int, not uint32_t for key length in `_mongocrypt_buffer_append`.

# Background & Motivation

It may be easier to review individual commits. The parsing may be simplified by the addition of the [bson-dsl](https://github.com/mongodb/mongo-c-driver/blob/master/src/libbson/src/bson/bson-dsl.md). That was decided out of scope of this PR.

## Driver API

The [DBX Scope: Driver Support for Range Index](https://docs.google.com/document/d/1kkiqr91-HPpr0lDCm1lDVWSvRCuDKKGgdmwYgi2OA7A/edit) describes the intended driver API:
```typescript
// RangeOpts specifies index options for a Queryable Encryption field
// supporting "range" queries.
class RangeOpts {
   // min defines the inclusive lower bound for field values.
   min: BSONValue, // must match value in encryptedFields.
   // max defines the inclusive upper bound for field values.
   max: BSONValue, // must match value in encryptedFields.
   sparsity: Int64 // must match value in encryptedFields.
}

class ClientEncryption {
   // If `opts.queryType` == "range", `value` is expected to be a BSON document
   // of one of the following forms:
   // 1. A Match Expression of this form:
   //   {$and: [{<field>: {$gt: <value1>, {<field>: {$lt: <value2> }}]}
   // 2. An Aggregate Expression of this form:
   //   {$and: [{$gt: [<fieldpath>, <value1>]}, {$lt: [<fieldpath>, <value2>]}]
   // $gt may also be $gte. $lt may also be $lte.
   encrypt(value: BsonValue, opts: EncryptOpts): Binary;
}
```

The API for range options is intended to put most error checking in libmongocrypt.
Drivers are expected to pass `value` from `ClientEncryption.encrypt` directly to `mongocrypt_ctx_explicit_encrypt_init`.
Drivers are expected to pass the `RangeOpts` directly to `mongocrypt_ctx_setopt_range`.
libmongocrypt will parse and transform the range options into the placeholder payload suitable for encryption.

## Making `edgesInfo` and `payload` optional

This matches recent changes in the the server IDL for [FLE2RangeFindSpec](https://github.com/mongodb/mongo/blob/25329194fe4343fd4ef1f3423377da742a47d5d6/src/mongo/crypto/fle_field_schema.idl#L411) and [FLE2FindRangePayload](https://github.com/mongodb/mongo/blob/25329194fe4343fd4ef1f3423377da742a47d5d6/src/mongo/crypto/fle_field_schema.idl#L324). Server code refers to [payloads without edges](https://github.com/mongodb/mongo/blob/25329194fe4343fd4ef1f3423377da742a47d5d6/src/mongo/crypto/fle_crypto.h#L1186-L1188) as stubs.


src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,629 @@
...
+      case FLE2RangeOperator_kLte:
+         if (spec->upper.set) {
+            CLIENT_ERR (
+               "unexpected duplicate bound %s: %s", op.op_type_str, str);
+            goto fail;
+         }
+         spec->upper.set = true;
+         spec->upper.value = op.value;
+         spec->upper.included = true;
+         break;
```
vector-of-bool, line 340: Suggest merging these branches:

```c
case FLE2RangeOperator_kLt:
case FLE2RangeOperator_kLte:
  [ ... ]
  spec->upper.included = op.op_type == FLE2RangeOperator_kLte;
  break;
```

(as well as with Gt+Gte)

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,629 @@
...
+                                  mongocrypt_status_t *status)
+{
+   *spec = (mc_FLE2RangeFindDriverSpec_t){0};
+   // `in` may be an Aggregate Expression with this form:
+   // {$and: [{$gt: ["$age", 5]}, {$lt:["$age", 50]}]}
+   // Or `in` may be a Match Expression with this form:
+   // {$and: [{age: {$gt: 5}}, {age: {$lt: 50}} ]}
+   bson_iter_t and, array;
+   bool ok = false;
+   char *str = bson_as_canonical_extended_json (in, NULL);
```
vector-of-bool, line 238: IIUC, `str` is only used in error messages in the `parse_` functions. Recommend passing the `bson_t` directly to these functions and have them generate the extended JSON string only when they need it in error messages.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,629 @@
...
+                             _mongocrypt_buffer_t *out,
+                             mongocrypt_status_t *status)
+{
+   BSON_ASSERT_PARAM (out);
+   BSON_ASSERT (status || true);
+
+   bool ok = false;
+   bson_t *edgesInfo = bson_new ();
+   bson_t *v = bson_new ();
+   bson_t *p = bson_new ();
```
vector-of-bool, line 392: Would it be worth it to just copy in the `bson-dsl.h` file? It should "just work" with the older libbson that is used by libmongocrypt, and may simplify the building/parsing aspects of this file.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,629 @@
...
+      payloadId = 0;
+   }
+}
+
+// mc_getNextPayloadId is thread safe.
+int32_t
+mc_getNextPayloadId (void)
+{
+   mlib_call_once (&payloadId_init_flag, payloadId_init_mutex);
+   int32_t ret;
```
vector-of-bool, line 619: Is it likely that we will overflow the int32? Would it be a problem to have duplicate IDs for a very-long-running process?

vector-of-bool (approved):
Nothing problematic sticks out to me. A few minor comments, but otherwise LGTM.

src/mc-fle2-rfds.c
kevinAlbs: Good idea. Added a `ERR_WITH_BSON` macro to defer creating JSON string

src/mc-fle2-rfds.c
kevinAlbs: Done.

src/mc-fle2-rfds.c
kevinAlbs: I think the DSL is likely to simplify this logic. Though given the timeline, I prefer doing as a separate PR to save time. The remaining work for Range Support is unlikely to require much parsing / building.

src/mc-fle2-rfds.c
kevinAlbs: No. The payloadId field is used in [server-side validation](https://docs.google.com/document/d/1JLrpDNP1wjJCj-9dkQN9wbvSMxKe9MqL0XNF8dFbMuk/edit#heading=h.hemu885zhbfu) for one command:

> Every comparison operator with an encrypted range payload and a defined secondOp must have a sibling predicate with the same predicate ID and of the opposite sign ($gt and $lte, or $gt and $lt, for example).

The requirement is payloads within one command have distinct payloadIds.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+      TRY (bson_append_iter (edgesInfo, "upperBound", -1, &args.upperBound));
+      TRY (BSON_APPEND_BOOL (edgesInfo, "ubIncluded", args.ubIncluded));
+      TRY (bson_append_iter (edgesInfo, "indexMin", -1, &args.indexMin));
+      TRY (bson_append_iter (edgesInfo, "indexMax", -1, &args.indexMax));
+      TRY (BSON_APPEND_DOCUMENT (v, "edgesInfo", edgesInfo));
+   }
+
+   // create v.
+   TRY (BSON_APPEND_INT32 (v, "payloadId", args.payloadId));
+   TRY (BSON_APPEND_INT32 (v, "firstOperator", args.firstOp));
```
eramongodb, line 400:
```suggestion
   TRY (BSON_APPEND_INT32 (v, "firstOperator", (int32_t) args.firstOp));
```
Address `warning: implicit conversion changes signedness: 'mc_FLE2RangeOperator_t' to 'int32_t' (aka 'int') [-Wsign-conversion]`.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+   if (args.secondOp) {
+      TRY (BSON_APPEND_INT32 (v, "secondOperator", args.secondOp));
```
eramongodb, line 402:
```suggestion
      TRY (BSON_APPEND_INT32 (v, "secondOperator", (int32_t) args.secondOp));
```
Address `warning: implicit conversion changes signedness: 'mc_FLE2RangeOperator_t' to 'int32_t' (aka 'int') [-Wsign-conversion]`.

src/mc-rangeopts.c
```diff
@@ -0,0 +1,161 @@
...
+   }
+
+bool
+mc_RangeOpts_parse (mc_RangeOpts_t *ro,
+                    const bson_t *in,
+                    mongocrypt_status_t *status)
+{
+   bson_iter_t iter;
+   bool has_min = false, has_max = false, has_sparsity = false;
+   const char *error_prefix = "Error parsing RangeOpts: ";
```
eramongodb, line 48:
```suggestion
   const char *const error_prefix = "Error parsing RangeOpts: ";
```

src/mc-rangeopts.c
```diff
@@ -0,0 +1,161 @@
...
+      END_IF_FIELD
+      IF_FIELD (sparsity, error_prefix)
+      if (!BSON_ITER_HOLDS_INT64 (&iter)) {
+         CLIENT_ERR ("%sExpected int64 for sparsity, got: %s",
+                     error_prefix,
+                     mc_bson_type_to_string (bson_iter_type (&iter)));
+         return false;
+      };
+      ro->sparsity = bson_iter_int64 (&iter);
+      END_IF_FIELD
```
eramongodb, line 80:
```suggestion
      IF_FIELD (min, error_prefix)
      ro->min = iter;
      END_IF_FIELD

      IF_FIELD (max, error_prefix)
      ro->max = iter;
      END_IF_FIELD

      IF_FIELD (sparsity, error_prefix)
      if (!BSON_ITER_HOLDS_INT64 (&iter)) {
         CLIENT_ERR ("%sExpected int64 for sparsity, got: %s",
                     error_prefix,
                     mc_bson_type_to_string (bson_iter_type (&iter)));
         return false;
      };
      ro->sparsity = bson_iter_int64 (&iter);
      END_IF_FIELD
```
Whitespace to better indicate `IF_FIELD` blocks.

src/mc-rangeopts.c
```diff
@@ -0,0 +1,161 @@
...
+   struct {
+      bson_iter_t iter;
+      bool set;
+   } lte;
+} mc_FLE2RangeFindDriverSpec_t;
+
+void
+mc_RangeOpts_cleanup (mc_RangeOpts_t *ro)
+{
+   bson_destroy (ro->bson);
```
eramongodb, line 160:
```suggestion
{
   if (!ro) {
      return;
   }

   bson_destroy (ro->bson);
```

src/mc-rangeopts.c
```diff
@@ -0,0 +1,161 @@
...
+   struct {
+      bson_iter_t iter;
+      bool set;
+   } lt;
+   struct {
+      bson_iter_t iter;
+      bool set;
+   } lte;
+} mc_FLE2RangeFindDriverSpec_t;
+
```
eramongodb, line 156:
```suggestion
```
Unused?

src/mc-rangeopts.c
```diff
@@ -0,0 +1,161 @@
...
+                                     mongocrypt_status_t *status)
+{
+   BSON_ASSERT_PARAM (ro);
+   BSON_ASSERT_PARAM (v);
+   BSON_ASSERT_PARAM (out);
+   BSON_ASSERT (status || true);
+
+   bson_init (out);
+
+   const char *error_prefix = "Error making FLE2RangeInsertSpec: ";
```
eramongodb, line 107:
```suggestion
   const char *const error_prefix = "Error making FLE2RangeInsertSpec: ";
```

src/mc-rangeopts.c
```diff
@@ -0,0 +1,161 @@
...
+      return false;
+   }
+   if (!bson_append_iter (&child, "max", 3, &ro->max)) {
+      CLIENT_ERR ("%sError appending to BSON", error_prefix);
+      return false;
+   }
+   if (!bson_append_document_end (out, &child)) {
+      CLIENT_ERR ("%sError appending to BSON", error_prefix);
+      return false;
+   }
```
eramongodb, line 134: Can these be replaced by assertions?

src/mc-rangeopts-private.h
```diff
@@ -0,0 +1,64 @@
...
+ * {
+ *    "v": BSON value to encrypt,
+ *    "min": BSON value,
+ *    "max": BSON value
+ * }
+ *
+ * v is expect to be a BSON document of the form:
+ * { "v": BSON value to encrypt }.
+ *
+ * out is always initialized.
```
eramongodb, line 53: Can `*out` being initialized be made a precondition instead?

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+   case FLE2RangeOperator_kLt:
+      return "$lt";
+   case FLE2RangeOperator_kLte:
+      return "$lte";
+   case FLE2RangeOperator_kNone:
+      return "none";
+   default:
+      return "Unknown";
+   }
+   return "Unknown";
```
eramongodb, line 60:
```suggestion
```

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+}
+
+// Suffix an error message with the JSON string form of bson.
+#define ERR_WITH_BSON(bson, fmt, ...)                            \
+   if (1) {                                                      \
+      char *_str = bson_as_canonical_extended_json (bson, NULL); \
+      CLIENT_ERR (fmt ": %s", __VA_ARGS__, _str);                \
+      bson_free (_str);                                          \
+   } else                                                        \
+      (void) 0
```
eramongodb, line 77:
```suggestion
#define ERR_WITH_BSON(bson, fmt, ...)                              \
   if (1) {                                                        \
      char *_str = bson_as_canonical_extended_json ((bson), NULL); \
      CLIENT_ERR (fmt ": %s", __VA_ARGS__, _str);                  \
      bson_free (_str);                                            \
   } else                                                          \
      (void) 0
```
Guard expansion of `bson`.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+   if (bson_iter_next (&array)) {
+      ERR_WITH_BSON (
+         orig, "expected 2 elements in operand %s, got > 2", op_type_str);
+      goto fail;
+   }
+
+   out->field = field;
+   out->op_type_str = op_type_str;
+   out->op_type = get_operator_type (op_type_str);
+   out->value = value;
```
eramongodb, line 171:
```suggestion
   *out = (operator_value_t){
      .field = field,
      .op_type_str = op_type_str,
      .op_type = get_operator_type (op_type_str),
      .value = value,
   };

```
Opt into `-Wmissing-field-initializers` warnings.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+
+   if (bson_iter_next (&document)) {
+      ERR_WITH_BSON (orig, "expected 1 elements in operand %s, got > 1", field);
+      goto fail;
+   }
+
+   out->field = field;
+   out->op_type_str = op_type_str;
+   out->op_type = get_operator_type (op_type_str);
+   out->value = value;
```
eramongodb, line 223:
```suggestion
   *out = (operator_value_t){
      .field = field,
      .op_type_str = op_type_str,
      .op_type = get_operator_type (op_type_str),
      .value = value,
   };
```
Opt into `-Wmissing-field-initializers` warnings.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+         if (0 != strcmp (spec->field, op.field)) {
+            ERR_WITH_BSON (
+               in,
+               "unexpected field mismatch. Expected all fields to be "
+               "%s, but got %s",
+               spec->field,
+               op.field);
+            goto fail;
+         }
+      }
```
eramongodb, line 343:
```suggestion
      if (spec->field && 0 != strcmp (spec->field, op.field)) {
         ERR_WITH_BSON (in,
                        "unexpected field mismatch. Expected all fields to be "
                        "%s, but got %s",
                        spec->field,
                        op.field);
         goto fail;
      }
```
Minor simplification of conditional expressions.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+
+      if (spec->nOps == 1) {
+         spec->firstOp = op.op_type;
+      } else if (spec->nOps == 2) {
+         spec->secondOp = op.op_type;
+      } else {
+         ERR_WITH_BSON (
+            in, "expected 1 or 2 operators, got > 2: %s", op.op_type_str);
+         goto fail;
+      }
```
eramongodb, line 355:
```suggestion
      switch (spec->nOps) {
      case 1:
         spec->firstOp = op.op_type;
         break;
      case 2:
         spec->secondOp = op.op_type;
         break;
      default:
         ERR_WITH_BSON (
            in, "expected 1 or 2 operators, got > 2: %s", op.op_type_str);
         goto fail;
      }
```

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+   BSON_ASSERT_PARAM (out);
+   BSON_ASSERT (status || true);
+
+   bool ok = false;
+   bson_t *edgesInfo = bson_new ();
+   bson_t *v = bson_new ();
+   bson_t *p = bson_new ();
+   _mongocrypt_buffer_init (out);
+
+#define TRY(stmt)                                          \
```
eramongodb, line 381: Can `TRY` be replaced with `BSON_ASSERT` instead?

Question also applies to other instances in this file.

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+   TRY (BSON_APPEND_INT32 (p, "t", MONGOCRYPT_FLE2_PLACEHOLDER_TYPE_FIND));
+   TRY (BSON_APPEND_INT32 (p, "a", MONGOCRYPT_FLE2_ALGORITHM_RANGE));
+   TRY (_mongocrypt_buffer_append (args.index_key_id, p, "ki", 2));
+   TRY (_mongocrypt_buffer_append (args.user_key_id, p, "ku", 2));
+   TRY (BSON_APPEND_DOCUMENT (p, "v", v));
+   TRY (BSON_APPEND_INT64 (p, "cm", args.maxContentionCounter));
+   TRY (BSON_APPEND_INT64 (p, "s", args.sparsity));
+#undef TRY
+
+   _mongocrypt_buffer_resize (out, p->len + 1);
```
eramongodb, line 415:
```suggestion
   _mongocrypt_buffer_resize (out, p->len + 1u);
```

src/mc-fle2-rfds.c
```diff
@@ -0,0 +1,615 @@
...
+{
+   mlib_call_once (&payloadId_init_flag, payloadId_init_mutex);
+   int32_t ret;
+   MONGOCRYPT_WITH_MUTEX (payloadId_mutex)
+   {
+      if (payloadId >= INT32_MAX) {
+         payloadId = 0;
+      }
+      ret = payloadId;
+      payloadId++;
```
eramongodb, line 612:
```suggestion
      ret = payloadId;
      payloadId = payloadId < INT32_MAX ? payloadId + 1 : 0;
```
`payloadId > INT32_MAX` is an impossible condition. The value `INT32_MAX` is skipped as-written. To include `INT32_MAX` as a possible payloadId, recommend rewriting as suggested.

eramongodb (changes requested):

src/mc-fle2-rfds.c
eramongodb: Ditto of other comment regarding input validation.

PR #478, opened 2022-10-18
MONGOCRYPT-360 returns an error when kms_ctx_feed is called with empty bytes

galon1:


test/test-mongocrypt-kms-ctx.c
```diff
@@ -506,6 +506,39 @@ _test_mongocrypt_kms_ctx_default_port (_mongocrypt_tester_t *tester)
...
+static void
+_test_mongocrypt_kms_ctx_feed_empty_bytes (_mongocrypt_tester_t *tester)
+{
+   mongocrypt_t *crypt;
+   mongocrypt_kms_ctx_t kms_ctx = {0};
+   mongocrypt_status_t *status;
+   _mongocrypt_endpoint_t *endpoint;
+   mongocrypt_binary_t *bytes = mongocrypt_binary_new();
+
+   crypt = _mongocrypt_tester_mongocrypt (TESTER_MONGOCRYPT_DEFAULT);
```
kevinAlbs, line 518: Run clang-format before merging.

I suggest having Visual Studio Code format on save. This is a snippet from my user settings:
```
    "[c]": {
        "editor.tabSize": 3,
        "editor.wordBasedSuggestions": false,
        "editor.suggest.insertMode": "replace",
        "editor.semanticHighlighting.enabled": true,
        "editor.formatOnSave": true
    },
```

kevinAlbs (approved):
Nice, LGTM with a run of clang-format.

PR #480, opened 2022-10-21
PYTHON-3256 Fix handling of AWS credentials

blink1073:


bindings/python/test-requirements.txt
```diff
@@ -7,3 +7,6 @@ cffi>=1.12.0,<2;python_version!="3.4"
 # install cryptography with PyPy with a newer OpenSSL.
 cryptography>=2,<2.9;platform_python_implementation=='PyPy'
 cryptography>=2;platform_python_implementation!='PyPy'
+mock;python_version=='2.7'
+requests_mock>=1.10
+six>=1.5
```
ShaneHarvey, line 12: Is this only needed for python 2.7 as well? If so can we add python_version=='2.7' so that we can remember to eventually remove it?

blink1073: Yep, updated

ShaneHarvey (approved):

PR #481, opened 2022-10-21
MONGOCRYPT-498 Automatically Obtain and Use a Fixed CMake Version

vector-of-bool:


.evergreen/ensure-cmake.sh
```diff
@@ -0,0 +1,161 @@
...
+: "${CMAKE_CACHE_DIR:="$CMAKE_CACHES_ROOT/$CMAKE_VERSION"}"
+
+# The executable that we want to use (can be overriden by an invoker)
+: "${CMAKE_EXE:="$CMAKE_CACHE_DIR/bin/cmake$EXE_SUFFIX"}"
+: "${CTEST_EXE:="${CMAKE_EXE%cmake*}ctest"}"
+
+# Downloads a prebuilt CMake binary:
+_download_cmake() {
+    # Directory where we will extract to (temporary)
+    declare extract_dir="$CMAKE_CACHE_DIR.tmp"
```
rcsanchez97, line 35: I would suggest using `mktemp` to give a randomized temporary directory name.  It's probably not the case that we require the ability to run multiple concurrent builds out of the same directory tree, but I still see it as beneficial here.

rcsanchez97 (approved):
I approve of this change in general.

However, I do have one thing that I think we should consider.

Last year we established a toolchain project (CDRIVER-2792) which builds minimum versions of dependencies (including CMake).  The idea is that we wanted to ensure that we could always build with our specified minimum dependencies.  At the time we only implemented a single task in the C driver Evergreen project to build with the toolchain.  Expanding to multiple build variants and other projects (CXX-2162 and MONGOCRYPT-303) were moved to the backlog.

I think that this can inform our thinking in a couple of ways.  The idea of "download and build CMake" is something which we already have implemented (albeit in a rough way) in the toolchain project.  We could potentially revisit the toolchain project, expanding it to allow us to have not just a toolchain with minimum versions, but a toolchain with specific components at specific versions and to incorporate more robust logic as in this PR.  As an alternative, we could implement some of the concepts from the toolchain project in a limited fashion just for CMake (though the mechanism ought to be shared between at least the three projects).

If we go the route of a more limited implementation then it would be worth considering whether incorporating some of what is implemented here into the toolchain project would be beneficial.  We should also probably consider what the future of the toolchain project might be.

.evergreen/init.sh
```diff
@@ -190,3 +190,44 @@ _init_sh_evg_dir="$(dirname "${_init_sh_this_file}")"
...
+
+EXE_SUFFIX=""
+if test "$OS_NAME" = "windows"; then
+    EXE_SUFFIX=".exe"
+fi
+
+if test "${USER_CACHES_DIR:=${XDG_CACHE_HOME:-}}" = ""; then
+    case "$OS_NAME" in
+    linux)
+        USER_CACHES_DIR=$HOME/.cache
```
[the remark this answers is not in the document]
vector-of-bool: Unquoted assignments are okay, as there is no word splitting semantics for variable assignments.

.evergreen/ensure-cmake.sh
vector-of-bool: Beyond trying to avoid racing on creating the cached copy, I don't see the benefit of creating a unique temporary directory. Even if this were written to a temporary location, there could still be a race on placing the final cache results. I don't know of any portable shell-only method of inter-process synchronization here.

rcsanchez97:
@vector-of-bool I'm good with things as they are.  I just wanted to raise those points for your consideration.

src/mlib/macros.h
```diff
@@ -0,0 +1,17 @@
+#ifndef MLIB_MACROS_H_INCLUDED
+#define MLIB_MACROS_H_INCLUDED
+
+#include "./user-check.h"
+
+/**
+ * @brief Cross-C/C++ compatibility for a compound initializer to be treated as
```
eramongodb, line 7: Rather than modifications to `mstr` headers (wide impact), can `csfle-markup.cpp` be made into a C source file instead (small impact)?

.evergreen/config.yml
```diff
@@ -1080,6 +1073,7 @@ buildvariants:
     has_packages: true
     packager_distro: rhel67
     packager_arch: s390x
+    compile_env: CMAKE_EXE=/opt/cmake/bin/cmake
```
eramongodb, line 1076: Why the special-case for rhel67-zseries-test? Can this be omitted so that `${compile_env|}` above can be made unnecessary?

.evergreen/init.sh
```diff
@@ -190,3 +190,44 @@ _init_sh_evg_dir="$(dirname "${_init_sh_this_file}")"
 # this one, and will therefore remain as native paths
 EVG_DIR="$(native_path "${_init_sh_evg_dir}")"
 LIBMONGOCRYPT_DIR="$(dirname "${EVG_DIR}")"
+
+# Executes CMake via the cache-managing script
+run_cmake() {
+    command bash "$EVG_DIR/cmake.sh" "$@"
+}
```
eramongodb, line 197:
```suggestion
run_cmake() (
    command bash "$EVG_DIR/cmake.sh" "$@"
)
```
Recommend using subshells `()` by default instead of scopes `{}` for new functions that do not require exporting variables to the environment.

.evergreen/init.sh
```diff
@@ -190,3 +190,44 @@ _init_sh_evg_dir="$(dirname "${_init_sh_this_file}")"
...
+fi
+
+# Ensure we are dealing with a complete path
+USER_CACHES_DIR="$(abspath "$USER_CACHES_DIR")"
+
+: "${BUILD_CACHE_BUST:=1}"
+: "${BUILD_CACHE_DIR:="$USER_CACHES_DIR/libmongocrypt/build.$BUILD_CACHE_BUST"}"
+
+# Silence shellcheck:
+: "$LIBMONGOCRYPT_DIR,$EXE_SUFFIX"
```
eramongodb, line 233:
```suggestion
export LIBMONGOCRYPT_DIR
export EXE_SUFFIX
```

Recommend explicitly exporting variables that are intended to be used by the script invoker.

src/mlib/macros.h
vector-of-bool: It could be, but these changes are incidentally useful elsewhere.

.evergreen/config.yml
vector-of-bool: CMake does not ship a CMake binary for s390x, so we would need to build one. But: building a CMake would require a newer version of C++ and CMake than ships on RHEL 6.7, so rhel67 is just the unfortunate odd-one-out in this case. (Hopefully soon to be dropped as well)

.evergreen/init.sh
vector-of-bool: This also exports to subprocesses, which isn't necessarily an issue, but something I wanted to avoid.

.evergreen/init.sh
vector-of-bool: I haven't seen this idiom before. Is there a benefit to this shape? Also, would this subshell attempt to fork? On Windows the Cygwin fork() is abysmally slow.

eramongodb: The benefit is explicitly documenting that the given block/function does not intend to modify its environment (self-contained). I was not aware that subshells in Cygwin were that slow compared to on Linux. I do not think it is severe enough to discourage use of subshells, but I may be mistaken.

src/mlib/macros.h
eramongodb: Can you elaborate? The changes seem applicable only to satisfy the `csfle-markup.cpp` executable. I am not aware of `mlib` being utilized elsewhere that demands C++-compatibility.

.evergreen/init.sh
eramongodb: > does not intend to modify its environment (self-contained)

Worth noting that this also includes setting shell options such as `errexit`, `pipefail`, `nounset`, and `xtrace`. Manually preserving and restoring shell options [can be done](https://unix.stackexchange.com/a/310963), but it is complicated, unwieldy, and difficult to ensure all exit paths are reliably covered.

kevinAlbs (approved):
Apologies for the delayed review. LGTM.

src/mlib/macros.h
vector-of-bool: Some parts of libmongocrypt are being merged into the server source tree, including this file, so it is important that it compile cleanly in a C++ compiler. The fact that it was compiling cleanly previously was due to GCC supporting compound-initializers in C++ mode, when it really shouldn't.

.evergreen/init.sh
vector-of-bool: Cygwin's fork() emulation is abysmally slow. It's on the order of dozens of milliseconds, which is "fast" to the eye, but adds up extremely quickly in a script execution. Not necessarily a problem here where a CMake subprocess will dominate the time, but I don't think the isolation of a subshell matters much here either, where the function is only running a single subcommand.

eramongodb (approved):

PR #484, opened 2022-11-02
PYTHON-3396 Support the Azure VM-assigned Managed Identity for Automatic KMS Credentials

blink1073:
Will require a corresponding integration test in mongo-python-driver

### DIFF
--- a/bindings/python/pymongocrypt/credentials.py
+++ b/bindings/python/pymongocrypt/credentials.py
@@ -102,7 +102,7 @@ def _get_azure_credentials():
     return { 'accessToken': data['access_token'] }
 
 
-def _ask_for_kms_credentials(kms_providers, owner):
+def _ask_for_kms_credentials(kms_providers):
     """Get on-demand kms credentials.
 
     This is a separate function so it can be overridden in unit tests."""
--- a/bindings/python/pymongocrypt/credentials.py
+++ b/bindings/python/pymongocrypt/credentials.py
@@ -1,4 +1,4 @@
-# Copyright 2019-present MongoDB, Inc.
+# Copyright 2022-present MongoDB, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


bindings/python/pymongocrypt/credentials.py
```diff
@@ -0,0 +1,136 @@
...
+except ImportError:
+    _HAVE_AUTH_AWS = False
+
+from pymongocrypt.errors import MongoCryptError
+
+import requests
+
+
+_azure_creds = namedtuple("_azure_creds", ["access_token", "expires_in"])
+_azure_creds_cache = None
```
ShaneHarvey, line 30: Why do we cache azure creds but not GCP creds? Is that coming later?

bindings/python/pymongocrypt/credentials.py
```diff
@@ -0,0 +1,136 @@
...
+            creds_dict["sessionToken"] = aws_creds.token
+        creds['aws'] = creds_dict
+    if on_demand_gcp:
+        creds['gcp'] = _get_gcp_credentials()
+    if on_demand_azure:
+        try:
+            creds['azure'] = _get_azure_credentials()
+        except Exception as e:
+            _azure_creds_cache = None
+            raise e
```
ShaneHarvey, line 135: I believe it's preferred to use bare raise instead so that the traceback is unadulterated.

ShaneHarvey (changes requested):

bindings/python/pymongocrypt/credentials.py
blink1073: In the scope document for this issue it was determined that caching was not needed for GCP creds.

bindings/python/pymongocrypt/credentials.py
blink1073: Done

ShaneHarvey (approved):

PR #486, opened 2022-11-03
Fix compile times of range test files.

vector-of-bool:
Declaring extremely large local variables seems to confuse the GCC optimizer to the point that it takes minutes to optimize the function. Making the large variables 'static' removes this optimization barrier.


kevinAlbs (approved):
Nice find!

PR #487, opened 2022-11-08
Use Ninja for Native Component Compilation

vector-of-bool:
This changeset builds upon #481, and will pull a Ninja executable to invoke the native build. On Unix-like systems, this is roughly the same speed as parallel Make, but provides much better build output (Ninja prevents output interleaving).

On Windows with MSVC, there is a vast improvement over using MSBuild: Cumulative build times are shaved by +2 minutes in each build-and-test task, and the build output is far cleaner than the default output of MSBuild (Ninja only prints a single status line for each file).

An explanation of changes:

- Ninja 1.8.2 is chosen to be pulled from the web. This is the last version that ships a prebuilt binary that is compatible with the glibc that is shipped on RHEL 6. Building a newer version from source is also feasible, but building those newer Ninja versions require a newer CMake version than we can provide for RHEL 6 s390x.
- To tell CMake about the Ninja executable, the CMake variable `CMAKE_MAKE_PROGRAM` should be set to the executable path of Ninja. This tells CMake how to run Ninja during `cmake --build`. (Ninja does not need to be on the PATH for CMake to use it.)
- To use MSVC in CMake without the VS generator, the VS build environment must be loaded. This environment includes the required environment variables for the CL compiler and linker to execute and build code for the appropriate target platform. In this PR, loading of the VS environment is done by the `vs-env-run.ps1` PowerShell script, which uses heuristics to find the batch scripts that set up the environment. It will then execute a subprocess with the appropriate environment loaded. It is recommended to set `CXX` and `CC` to `cl` to inform CMake of a preference to use MSVC, otherwise it may find GCC or Clang on the PATH as the "default" compiler when the Ninja generator is used.
- In this PR, selecting the VS environment is done with the `VS_VERSION` and `VS_TARGET_ARCH` environment variables. These are used by `env-run.sh`, which abstracts out the hop-in+hop-out of loading the VS environment for a subcommand within a Cygwin shell script.
- The intermediate config-specific build directories are a characteristic of the multi-conf CMake generators, and not a characteristic of Windows. Path determination logic has been adjusted accordingly.


bindings/python/pymongocrypt/credentials.py
```diff
@@ -0,0 +1,136 @@
+# Copyright 2022-present MongoDB, Inc.
```
ShaneHarvey, line 1: I'm surprised these python changes are showing up here. Could you try rebasing?

vector-of-bool: I'm also unsure why this happens. I've seen it across github several times. I'll try a rebase.

vector-of-bool: Better now. I had accidentally cherry-picked a change when I intended to merge `master`.

.evergreen/build_all.sh
```diff
@@ -14,25 +14,14 @@ _cxxflags=""
 
 : "${CONFIGURE_ONLY:=}"
 
-# Use C driver helper script to find cmake binary, stored in $CMAKE.
 if [ "$OS_NAME" = "windows" ]; then
-    : "${CMAKE:=/cygdrive/c/cmake/bin/cmake}"
     # Enable exception handling for MSVC
     _cxxflags="-EHsc"
-    if [ "${WINDOWS_32BIT-}" != "ON" ]; then
+    if test "${WINDOWS_32BIT-}" != "ON" && ! "${USE_NINJA-false}" ; then
```
rcsanchez97, line 20: I recommend changing the way the `USE_NINJA` variable is used.  Having the test evaluate the variable contents as a command (`true` or `false` in this case) is slick, but different from how all the other variables in our helper scripts are treated.  I recommend changing the way `USE_NINJA` is handled so that is consistent with other similar variables.

rcsanchez97 (changes requested):

.evergreen/build_all.sh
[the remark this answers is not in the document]
vector-of-bool: I also realized that executing an arbitrary command from an environment variable is just really suspicious altogether. But that led me to another idea:

```bash
is_true() {
    declare var="$1"
    declare val="${!var-@L}"  # Convert to lowercase, default is '' empty
    set -x
    case "$val" in
    1|true|yes|on)
        return 0;;
    0|false|no|off|"")
        return 1;;
    *)
        log "Unknown 'boolean' value for \$$var: '$val'"
        return 2;;
    esac
}

...

if is_true USE_NINJA; then 
  ...
fi
```

Which affords lenience in how the booleans are spelled, and no longer requires that each boolean be specially handled to prevent tripping `nounset`. Too clever?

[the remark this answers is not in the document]
rcsanchez97: I like that approach much better.  The only thing I would say is missing is a return trap that executes `set +x`, since we don't want xtrace to stay on after this function call.

rcsanchez97 (approved):
LGTM

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,247 @@
...
+    #             $found += $(
+    #                 Get-ChildItem "${env:ProgramFiles(x86)}\Windows Kits\8.1\Include"`
+    #                 | Where-Object { Test-Path "$($_.FullName)\um\Windows.h" }`
+    #                 | ForEach-Object { Split-Path -Leaf $_.FullName }
+    #             )
+    #         }
+    #         return $found | Where-Object { $_ -like "$wordToComplete*" }
+    #     })]
+    $WinSDKVersion,
+    # The host architecture to use. Not usually needed. Detaults to x86
```
eramongodb, line 81:
```suggestion
    # The host architecture to use. Not usually needed. Defaults to x86.
```

.evergreen/init.sh
```diff
@@ -190,3 +190,70 @@ _init_sh_evg_dir="$(dirname "${_init_sh_this_file}")"
...
 LIBMONGOCRYPT_DIR="$(dirname "${EVG_DIR}")"
+
+is_true() {
+    declare var="$1"
+    declare val="${!var-}"  # Default is '' empty
+    case "$val" in
+    1|true|TRUE|True|yes|YES|Yes|on|ON|On)
+        return 0;;
+    0|false|FALSE|False|no|NO|No|off|OFF|Off|"")
+        return 1;;
```
eramongodb, line 201: I was going to suggest simplifying these comparison by using `declare -l val=...` to convert string to lowercase before comparison, but it seems like Bash on macos-1014 doesn't support it... 😕

.evergreen/init.sh
```diff
@@ -190,3 +190,70 @@ _init_sh_evg_dir="$(dirname "${_init_sh_this_file}")"
 # this one, and will therefore remain as native paths
 EVG_DIR="$(native_path "${_init_sh_evg_dir}")"
 LIBMONGOCRYPT_DIR="$(dirname "${EVG_DIR}")"
+
+is_true() {
+    declare var="$1"
+    declare val="${!var-}"  # Default is '' empty
```
eramongodb, line 196: Why the need for `!` in `${!var-}`?

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,247 @@
...
+    [switch]
+    $UseSpectreMiticationLibraries,
+    # The app platform to load. Default is "Desktop"
+    [ValidateSet("Desktop", "UWP", IgnoreCase = $false)]
+    [string]
+    $AppPlatform = "Desktop",
+    # The directory to store ephemeral files
+    [string]
+    $ScratchDir,
+
```
eramongodb, line 98:
```suggestion
```

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,247 @@
...
+    & $Command
+    if ($LASTEXITCODE -ne 0) {
+        # If it was a process that returned non-zero, throw an error.
+        throw "Subcommand failed [$LASTEXITCODE]"
+    }
+}
+finally {
+    # Restore the prior environment
+    foreach ($key in $prior_env.Keys) {
+        [System.Environment]::SetEnvironmentVariable($key, $prior_env[$key])
```
eramongodb, line 245: Given this script is intended to be invoked as a subshell, is it really necessary to include logic to preserve and restore environment variables in this manner?

.evergreen/init.sh
vector-of-bool: Yeah, I was unable to make this "pretty"...

.evergreen/init.sh
vector-of-bool: An indirect dereference. You pass the name of the var, this will dereference the var by name.

.evergreen/vs-env-run.ps1
vector-of-bool: In the case of a subprocess, no, but if you invoke this script from an existing PS script/repl (without spawning a subprocess), it will share the environment variables, so you have to put them back before returning.

.evergreen/init.sh
eramongodb: Oh, I overlooked the intended usage of this function as verifying the value of the given _variable_ has a truthy value, not the argument itself. Disregard.

.evergreen/vs-env-run.ps1
vector-of-bool: No scripts in this PR use this script in such a manner, but I use it frequently in my own workflow both from REPL and other PS scripts.

eramongodb (approved):

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,246 @@
...
+    #             $found += $(
+    #                 Get-ChildItem "${env:ProgramFiles(x86)}\Windows Kits\8.1\Include"`
+    #                 | Where-Object { Test-Path "$($_.FullName)\um\Windows.h" }`
+    #                 | ForEach-Object { Split-Path -Leaf $_.FullName }
+    #             )
+    #         }
+    #         return $found | Where-Object { $_ -like "$wordToComplete*" }
+    #     })]
+    $WinSDKVersion,
+    # The host architecture to use. Not usually needed. Detaults to x86.
```
eramongodb, line 81:
```suggestion
    # The host architecture to use. Not usually needed. Defaults to x86.
```
Minor typo still overlooked.

.evergreen/linker-tests.sh
```diff
@@ -76,15 +92,14 @@ BUILD_DIR="$linker_tests_root/app-cmake-build"
...
   -DCMAKE_PREFIX_PATH="$PREFIX_PATH" \
   "-H$SRC_PATH" \
   "-B$BUILD_DIR"
 run_cmake --build "$BUILD_DIR" --target app --config RelWithDebInfo
 
-if [ "$OS_NAME" = "windows" ]; then
-    export PATH="$PATH:$BSON1_INSTALL_PATH/bin:$LMC_INSTALL_PATH/bin"
+export PATH="$PATH:$BSON1_INSTALL_PATH/bin:$LMC_INSTALL_PATH/bin"
+if is_true IS_MULTICONF; then
     APP_CMD="$BUILD_DIR/RelWithDebInfo/app.exe"
```
kevinAlbs, line 103:
```suggestion
    APP_CMD="$BUILD_DIR/RelWithDebInfo/app$EXE_SUFFIX"
```

Since this is no longer Windows specific.

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,246 @@
...
+    # The host architecture to use. Not usually needed. Defaults to x86.
+    [ValidateSet("x86", "amd64", IgnoreCase = $false)]
+    [string]
+    $HostArch = "x86",
+    # The Visual C++ toolset to load
+    [string]
+    $VCToolsetVersion,
+    # Prefer Visual C++ libraries with Spectre mitigations
+    [switch]
+    $UseSpectreMiticationLibraries,
```
kevinAlbs, line 90:
```suggestion
    $UseSpectreMitigationLibraries,
```

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,246 @@
...
+        call "$($vcvarsall_bat.FullName)" $TargetArch $WinSDKVersion
+        set _rc=%ERRORLEVEL%
+        set
+        exit /b %_rc%
+"@
+}
+else {
+    # Build up the argument string to load the appropriate environment
+    $argstr = "-no_logo -arch=$TargetArch -host_arch=$HostArch -app_platform=$AppPlatform"
+    if ($UseSpectreMiticationLibraries) {
```
kevinAlbs, line 189:
```suggestion
    if ($UseSpectreMitigationLibraries) {
```

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,246 @@
...
+    -UseBasicParsing `
+    -Uri "https://github.com/microsoft/vswhere/releases/download/3.0.3/vswhere.exe" `
+    -OutFile $vswhere
+
+# Ask vswhere for all the installed products:
+$vswhere_json = & $vswhere -utf8 -nologo -format json -all -legacy -prerelease -products * | Out-String
+$vs_versions = $vswhere_json | ConvertFrom-Json
+Write-Debug "Detected VS versions: $vs_versions"
+
+# Pick the produce that matches the pattern
```
kevinAlbs, line 126:
```suggestion
# Pick the product that matches the pattern
```

.evergreen/vs-env-run.ps1
```diff
@@ -0,0 +1,246 @@
...
+
+New-Item $ScratchDir -ItemType Directory -ErrorAction Ignore
+$vswhere = Join-Path $ScratchDir "vswhere.exe"
+
+$ProgressPreference = "SilentlyContinue"
+[Net.ServicePointManager]::SecurityProtocol = 'tls12, tls11'
+Invoke-WebRequest `
+    -UseBasicParsing `
+    -Uri "https://github.com/microsoft/vswhere/releases/download/3.0.3/vswhere.exe" `
+    -OutFile $vswhere
```
kevinAlbs, line 119: Should this check if `$vswhere` exists first? What if `vs-env-run.ps1` is invoked twice with the same scratch directory?

kevinAlbs (approved):
LGTM, nicely done. The faster build times are much appreciated.

.evergreen/vs-env-run.ps1
vector-of-bool: It's fine if the file already exists, but a tiny perf boost to not re-download it

PR #488, opened 2022-11-09
ci(NODE-4779): Pin to an earlier Node LTS release on platforms that are not supported by prebuilt Node 18+

vector-of-bool:
On Oct 25, 2022, Node LTS "Hydrogen" 18 went active, and this changed the major version downloaded by `nvm {use,install} --lts`. The new Node binaries have a runtime-link dependency on some newer versions of glibc symbols that are not available on older Linux distros under test. The downloaded `node` executable can not be launched on these older systems, confusing `nvm` and preventing the CI from running properly.

This changeset pins the older build variants to download the prior LTS release (16.x).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -137,6 +137,7 @@ functions:
           NODE_GITHUB_TOKEN: ${node_github_token}
           DISTRO_ID: ${distro_id}
           BUILD_VARIANT: ${build_variant}
+          NODE_NVM_USE_VERSION: ${nvm_use_version|lts}
 
   "build and test node no optional dependencies":
     - command: "subprocess.exec"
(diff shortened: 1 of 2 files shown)


durran:
Just ran into this yesterday as well, we could also move to Ubuntu 20.04 to resolve but that may have some other unwanted side effects. LGTM, just waiting on CI to finish.

dariakp (approved):
I'm not sure if those remaining EVG tasks are important to let run on this PR, it doesn't look like they ran on the base commit. @durran do you know?

vector-of-bool:
It looks like EVG is struggling to spawn the SUSE 15 hosts for the tasks.

vector-of-bool:
I verified locally that these changes fix CI on SLES 15. I'll merge without waiting on the tasks. If anything comes up I'll put in another fix.

PR #489, opened 2022-11-10
ci(NODE-4779): Missing Pin of Node 16 on SUSE15 Variant

vector-of-bool:
Now that SUSE15 hosts are spawning, I realized that I missed one variant in #488. Oops, that's embarrassing.


dariakp (approved):
It happens, thanks for catching it!

PR #497, opened 2022-11-16
MONGOCRYPT-490 support precision for `double`

kevinAlbs:
# Summary

- Support optional precision for `double`.
- Support precision in explicit encryption for `double`.
- Make `min` and `max` arguments optional for explicit encryption.

## Additional changes

- Add `mc_optional_uint32_t` and `mc_optional_uint64_t`.
- Refactor `_test_encrypt_fle2_explicit` to shorten tests.

# Background & Motivation

Precision specifies the number of significant decimal places for encrypting `double` or `decimal128`. Changes to `mc_getTypeInfoDouble` are ported from the server's [getTypeInfoDouble](https://github.com/mongodb/mongo/commit/0d67266b5759f89dbf63b72d4d1172c7a421a4a2#diff-0ef885be428257ef770745c40f6ce33e93bbcb6f266b3f8e00cadf6aeb0a1e6dR3531).

The intended driver API for `RangeOpts` is described in [DBX Scope: Driver Support for Range Index](https://docs.google.com/document/d/1kkiqr91-HPpr0lDCm1lDVWSvRCuDKKGgdmwYgi2OA7A/edit#heading=h.haavpvf17xzz):
```
// RangeOpts specifies index options for a Queryable Encryption field
// supporting "range" queries.
// For double and decimal128, min/max/precision must all be set, or all be unset.
class RangeOpts {
   // min defines the inclusive lower bound for field values.
   // min is required if precision is set.
   min: Optional<BSONValue>, // must match value in encryptedFields.
   // max defines the inclusive upper bound for field values.
   // max is required if precision is set.
   max: Optional<BSONValue>, // must match value in encryptedFields.
   sparsity: Int64 // must match value in encryptedFields.
   precision: Optional<Int32> // must match value in encryptedFields. Only applicable for double or decimal128. Prohibited for other types.
}
```

Making `min` and `max` optional was motivated by matching [server behavior](https://github.com/mongodb/mongo/commit/0d9455e8ee5ed764aee6a514d789ae1b5ed0af03#diff-21fdaf455f52e97caa37093646d05b6fc182795e1152ec7d370ae98e14a7ee1cR100) for `double`/`decimal128`. For double and decimal128, min/max/precision must all be set, or all be unset.


src/mc-fle2-rfds.c
```diff
@@ -394,13 +394,19 @@ mc_makeRangeFindPlaceholder (mc_makeRangeFindPlaceholder_args_t *args,
...
    if (!args->isStub) {
       TRY (bson_append_iter (edgesInfo, "lowerBound", -1, &args->lowerBound));
       TRY (BSON_APPEND_BOOL (edgesInfo, "lbIncluded", args->lbIncluded));
       TRY (bson_append_iter (edgesInfo, "upperBound", -1, &args->upperBound));
       TRY (BSON_APPEND_BOOL (edgesInfo, "ubIncluded", args->ubIncluded));
       TRY (bson_append_iter (edgesInfo, "indexMin", -1, &args->indexMin));
       TRY (bson_append_iter (edgesInfo, "indexMax", -1, &args->indexMax));
+      if (args->precision.set) {
+         BSON_ASSERT (args->precision.value < INT32_MAX);
+         TRY (BSON_APPEND_INT32 (
```
galon1, line 407: Is this to double check that precision's value fits within a 32 bit integer, because when the value of precision is being parsed like in `mc_FLE2RangeInsertSpec_parse` the value of precision is also checked with `BSON_ITER_HOLDS_INT32`. When we get to this part is not confirmed that ` precision.value` is within a 32 bit integer? Is this check just to see that it is not equal to `INT32_MAX`?

src/mc-rangeopts-private.h
```diff
@@ -58,6 +73,26 @@ mc_RangeOpts_to_FLE2RangeInsertSpec (const mc_RangeOpts_t *ro,
...
+bool
+mc_RangeOpts_appendMin (const mc_RangeOpts_t *ro,
+                        bson_type_t valueType,
+                        const char *fieldName,
+                        bson_t *out,
+                        mongocrypt_status_t *status);
+
+/* mc_RangeOpts_appendMin appends the minimum value of the range for a given
+ * type. If `ro->min` is unset, uses the lowest representable value of the value
+ * type. Errors if `valueType` does not match the type of `ro->min`. */
```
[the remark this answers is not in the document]
kevinAlbs: Oops. Thank you.

src/mc-fle2-rfds.c
kevinAlbs: Fixed the assertion to be `<= INT32_MAX`.

precision is expected to be `uint32` in `mc_getTypeInfoDouble`.

precision is passed in as an `int32` from `mc_FLE2RangeInsertSpec_parse`, `mc_RangeOpts_parse`, or `mc_FLE2RangeFindSpecEdgesInfo_parse`.

This assertion should always hold true. The cast back from `uint32` to `int32` still fits in an `int32`.

src/mc-rangeopts.c
```diff
@@ -86,13 +91,96 @@ mc_RangeOpts_parse (mc_RangeOpts_t *ro,
...
+         if (minType == BSON_TYPE_DOUBLE || minType == BSON_TYPE_DECIMAL128) {
+            if (!has_precision) {
+               CLIENT_ERR ("expected 'precision' to be set with 'min' for %s",
+                           mc_bson_type_to_string (minType));
+               return false;
+            }
+         }
+      }
+
+      if (ro->max.set) {
```
galon1, line 173: Should there be an additional check here that `min.value <= max.value` or will that check happen in `CHECK_BOUNDS` in `mc-range-mincover.c`?

test/data/fle2-find-range-explicit/double-precision/value-to-encrypt.json
```diff
@@ -0,0 +1,20 @@
...
+        "$and": [
+            {
+                "age": {
+                    "$gte": {
+                        "$numberDouble": "123.456"
+                    }
+                }
+            },
+            {
+                "age": {
```
galon1, line 12: This might be related more to a testing ticket, and not this ticket, but it could be helpful to create a json test that has an open range (like just a `$gte` and not both `$lte` and `$gte`)?

src/mc-rangeopts.c
kevinAlbs: Good question. `min.value <= max.value` is checked when encoding in `mc_getTypeInfo32`, `mc_getTypeInfo64`, and `mc_getTypeInfoDouble`.

[Test cases](https://github.com/mongodb/libmongocrypt/pull/497/commits/d908eefd62a5c9f94e0bbb0772ed2cc2872fb315) with `max > `min` were added to expect an error.

CMakeLists.txt
```diff
@@ -336,6 +336,21 @@ elseif (MONGOCRYPT_CRYPTO STREQUAL OpenSSL)
...
+# macOS includes libm as part of libSystem (along with libc).
+# It does not need to be explicitly linked.
+if (NOT APPLE)
+   find_library (M_LIBRARY m)
+   if (M_LIBRARY)
+      target_link_libraries (mongocrypt PRIVATE ${M_LIBRARY})
+      target_link_libraries (mongocrypt_static PRIVATE ${M_LIBRARY})
+      set (PKG_CONFIG_STATIC_LIBS "${PKG_CONFIG_STATIC_LIBS} -lm")
+   endif ()
+endif ()
```
vector-of-bool, line 352: I feel like there should be a cleaner way to do this, but I'll investigate that later.

src/mc-range-encoding.c
```diff
@@ -177,22 +177,112 @@ mc_getTypeInfo64 (mc_getTypeInfo64_args_t args,
    return true;
 }
 
+#define exp10Double(x) pow (10, x)
```
vector-of-bool, line 180: Is there a reason for this macro for such a small expression?

vector-of-bool (approved):
LGTM, only one minor comment

galon1 (approved):
LGTM!

PR #498, opened 2022-11-16
MONGOCRYPT-492 Add check if datakey provider and the initial KMS provider set match

galon1:
These changes will raise an error and not create a datakey when the datakay uses a different KMs provider than the initial KMS provider configured in `mongocrypt_setopt_kms_providers`. There are two added tests: one with a configured KMS provider and one with KMS provider that needs credential (when it is configured with an empty object).

### DIFF
--- a/src/mongocrypt-ctx.c
+++ b/src/mongocrypt-ctx.c
@@ -901,20 +901,18 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
       }
    }
 
-   /* Check that the kms provider set and the datakey kms provider match. */
+   if (opts_spec->kek == OPT_PROHIBITED && ctx->opts.kek.kms_provider) {
+      return _mongocrypt_ctx_fail_w_msg (ctx, "master key prohibited");
+   }
+
+   /* Check that the kms provider required by the datakey is configured.  */
    if (ctx->opts.kek.kms_provider) {
       if (!((ctx->crypt->opts.kms_providers.need_credentials |
              ctx->crypt->opts.kms_providers.configured_providers) &
             ctx->opts.kek.kms_provider)) {
          return _mongocrypt_ctx_fail_w_msg (
-            ctx,
-            "datakey provider does not match the provider set in "
-            "mongocrypt_setopt_kms_providers");
+            ctx, "kms provider required by datakey is not configured");
       }
-   }
-
-   if (opts_spec->kek == OPT_PROHIBITED && ctx->opts.kek.kms_provider) {
-      return _mongocrypt_ctx_fail_w_msg (ctx, "master key prohibited");
    }
 
    /* Special case. key_descriptor applies to explicit encryption. It must be
--- a/src/mongocrypt-ctx.c
+++ b/src/mongocrypt-ctx.c
@@ -904,10 +904,8 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
    /* Check that the kms provider initially set and the datakey kms provider
     * match. */
    if (ctx->opts.kek.kms_provider) {
-      if (!(ctx->crypt->opts.kms_providers.need_credentials &
-            ctx->opts.kek.kms_provider) &&
-
-          !(ctx->crypt->opts.kms_providers.configured_providers &
+      if (!((ctx->crypt->opts.kms_providers.need_credentials |
+             ctx->crypt->opts.kms_providers.configured_providers) &
             ctx->opts.kek.kms_provider)) {
          return _mongocrypt_ctx_fail_w_msg (
             ctx,
--- a/src/mongocrypt-ctx.c
+++ b/src/mongocrypt-ctx.c
@@ -901,8 +901,7 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
       }
    }
 
-   /* Check that the kms provider initially set and the datakey kms provider
-    * match. */
+   /* Check that the kms provider set and the datakey kms provider match. */
    if (ctx->opts.kek.kms_provider) {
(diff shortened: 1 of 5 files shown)


test/test-gcp-auth.c
```diff
@@ -311,4 +352,6 @@ _mongocrypt_tester_install_gcp_auth (_mongocrypt_tester_t *tester)
    INSTALL_TEST (_test_encrypt_with_credentials);
    INSTALL_TEST (_test_createdatakey_with_accesstoken);
    INSTALL_TEST (_test_encrypt_with_accesstoken);
+   INSTALL_TEST (_test_createdatakey_with_wrong_kms_provider_configured);
+   INSTALL_TEST (_test_createdatakey_with_wrong_kms_provider_empty);
```
kevinAlbs, line 356: Though I originally pointed to this file, suggest moving these two tests to `test-mongocrypt-ctx-setopt.c`. The behavior change is option validation and not specific to `gcp`.

src/mongocrypt-ctx.c
```diff
@@ -901,6 +901,18 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
       }
    }
 
+   /* Check that the kms provider set and the datakey kms provider match. */
```
kevinAlbs, line 904:
```suggestion
   /* Check that the kms providers set and the datakey kms provider match. */
```

src/mongocrypt-ctx.c
```diff
@@ -901,6 +901,18 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
...
+   if (ctx->opts.kek.kms_provider) {
+      if (!((ctx->crypt->opts.kms_providers.need_credentials |
+             ctx->crypt->opts.kms_providers.configured_providers) &
+            ctx->opts.kek.kms_provider)) {
+         return _mongocrypt_ctx_fail_w_msg (
+            ctx,
+            "datakey provider does not match the provider set in "
```
kevinAlbs, line 911:
```suggestion
            "requested datakey provider was not configured"
```

This error may be propagated up through drivers. Users likely will not find references to libmongocrypt functions helpful.

src/mongocrypt-ctx.c
```diff
@@ -901,6 +901,18 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
       }
    }
 
+   /* Check that the kms provider set and the datakey kms provider match. */
+   if (ctx->opts.kek.kms_provider) {
```
kevinAlbs, line 905: Suggest moving this if block below the `OPT_PROHIBITED` check on L916.

The `kek` option, set through `mongocrypt_ctx_setopt_key_encryption_key` is only applicable to some contexts. If the `kek` option is set on an incorrect context, it would be a bug in the driver implementation. The "master key prohibited" error may be more helpful.

kevinAlbs (commented):
Looks good with some code movement and wording changes.

Parameterizing the test for different KMS provider configurations is helpful.

src/mongocrypt-ctx.c
eramongodb, line 904:
```suggestion
   /* Check that the kms provider required by the datakey is configured. */
```

src/mongocrypt-ctx.c
```diff
@@ -901,6 +901,18 @@ _mongocrypt_ctx_init (mongocrypt_ctx_t *ctx,
...
+      if (!((ctx->crypt->opts.kms_providers.need_credentials |
+             ctx->crypt->opts.kms_providers.configured_providers) &
+            ctx->opts.kek.kms_provider)) {
+         return _mongocrypt_ctx_fail_w_msg (
+            ctx,
+            "datakey provider does not match the provider set in "
+            "mongocrypt_setopt_kms_providers");
```
eramongodb, line 912:
```suggestion
         return _mongocrypt_ctx_fail_w_msg (
            ctx, "kms provider required by datakey is not configured");
```
Even if `mongocrypt_setopt_kms_providers` is the recommended method, it is not the only means by which a KMS provider may be configured. Suggest modifying error message to avoid being prescriptive.

test/test-gcp-auth.c
```diff
@@ -230,6 +229,49 @@ _test_createdatakey_with_accesstoken (_mongocrypt_tester_t *tester)
...
+_test_createdatakey_with_wrong_kms_provider_helper (
+   _mongocrypt_tester_t *tester, mongocrypt_binary_t *kms_provider)
+{
+   mongocrypt_t *crypt;
+   mongocrypt_ctx_t *ctx;
+   const char *kek = "{"
+                     "'provider': 'azure',"
+                     "'keyName': 'foo',"
+                     "'keyVaultEndpoint': 'example.com'"
+                     "}";
```
eramongodb, line 242:
```suggestion
   const char *const kek = "{"
                           "'provider': 'azure',"
                           "'keyName': 'foo',"
                           "'keyVaultEndpoint': 'example.com'"
                           "}";
```

kevinAlbs (approved):

eramongodb (approved):

test/test-mongocrypt.c
```diff
@@ -977,6 +977,7 @@ main (int argc, char **argv)
    _mongocrypt_tester_install_range_mincover (&tester);
    _mongocrypt_tester_install_mc_RangeOpts (&tester);
    _mongocrypt_tester_install_mc_FLE2RangeFindDriverSpec (&tester);
+   _mongocrypt_tester_install_gcp_auth (&tester);
```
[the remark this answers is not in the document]
galon1: Yes, since I moved the tests out of `_mongocrypt_tester_install_gcp_auth` it is unrelated to this ticket, but this test file should be running.

PR #499, opened 2022-11-17
MONGOCRYPT-494 rename queryType and algorithm from "range" to "rangePreview"

kevinAlbs:
# Summary

- Rename queryType and algorithm from "range" to "rangePreview"

# Background & Motivation

The `encryptedFields` in the server now expects the string `rangePreview`: https://github.com/mongodb/mongo/commit/484eee157ecc039fafbbef6a4aae0286c8bbc3b5#diff-64f7154b6f37606f91956209094c8cc9005f11ee48d8f54210a65cbb8bd102b8R56

It is expected to change back to "range" once the feature is stabilized.


galon1 (approved):
LGTM!

PR #502, opened 2022-11-28
MONGOCRYPT-446 add build variant and package publications for Debian 11 and Ubuntu 22.04

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/6384eb75850e615387a38ee9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Note that the `build-and-test-csharp` task is failing on Ubuntu 22.04.  I have communicated with @DmitryLukyanov about this and we concluded that it would best for me to go ahead with this PR and let the C# team address the failing task when it makes sense for the team's priorities.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -116,7 +116,7 @@ functions:
       params:
         script: |-
           if test "$OS_NAME" != "windows"; then export USE_NINJA=ON; fi
-          env ${compile_env|} CONFIGURE_ONLY=ON CC=clang CXX=clang++ \
+          env ${compile_env|} CONFIGURE_ONLY=ON ${clang_env|CC=clang CXX=clang++} \
               bash libmongocrypt/.evergreen/build_all.sh
           ./libmongocrypt/.evergreen/clang-tidy.sh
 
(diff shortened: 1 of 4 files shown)


.evergreen/config.yml
```diff
@@ -1035,6 +1047,22 @@ buildvariants:
   - name: publish-packages
     distros:
     - rhel70-small
+- name: debian11
+  display_name: "Debian 11.0"
+  run_on: debian11-large
```
kevinAlbs, line 1052:
```suggestion
  run_on: debian11-small
```

Suggest using small since libmongocrypt builds are not resource intensive.

kevinAlbs (approved):

.evergreen/config.yml
rcsanchez97: So, when I looked in the Evergreen host listings, I found that there are 4 debian11-large hosts and 0 debian11-small.  With the other distros it looks like there are a few -large and considerably more -small hosts.  I'm not sure why there aren't any debian11-small hosts, but I think the build would just stall if we went that route.  Either that, or debian11-small is an alias to debian11-large, meaning that there is only one size of host available for debian11 builds.

kevinAlbs: Interesting. There is a task in the Go driver using debian11-small: https://spruce.mongodb.com/task/mongo_go_driver_testgcpkms_variant_testgcpkms_fail_task_0e5ae89f1baf3d6edca093467ffd1f870b441add_22_11_28_21_11_31/logs?execution=0. Maybe that is an alias.

PR #509, opened 2022-12-05
MONGOCRYPT-505 add mongocrypt_ctx_explicit_encrypt_expression_init

kevinAlbs:
# Summary

- Add `mongocrypt_ctx_explicit_encrypt_expression_init`.

# Background & Motivation

The new init function is to help support the new [encryptExpression driver API](https://docs.google.com/document/d/1kkiqr91-HPpr0lDCm1lDVWSvRCuDKKGgdmwYgi2OA7A/edit#heading=h.haavpvf17xzz):

```typescript
class ClientEncryption {
   // encryptExpression encrypts an expression to query a range index.
   // `expr` is expected to be a BSON document of one of the following forms:
   // 1. A Match Expression of this form:
   //   {$and: [{<field>: {$gt: <value1>}}, {<field>: {$lt: <value2> }}]}
   // 2. An Aggregate Expression of this form:
   //   {$and: [{$gt: [<fieldpath>, <value1>]}, {$lt: [<fieldpath>, <value2>]}]
   // $gt may also be $gte. $lt may also be $lte.
   // Only supported for queryType "rangePreview"
   encryptExpression(expr: BsonDocument, opts: EncryptOpts): BsonDocument;
}
```

`mongocrypt_ctx_explicit_encrypt_expression_init` shares the same code as `mongocrypt_ctx_explicit_encrypt_init`. The only difference is the validation of the queryType argument:
- `mongocrypt_ctx_explicit_encrypt_expression_init` requires queryType of "rangePreview".
- `mongocrypt_ctx_explicit_encrypt_init` rejects queryType of "rangePreview".

A rejected alternative is for drivers could do the validation of queryType, and reuse `mongocrypt_ctx_explicit_encrypt_init`.

### DIFF
--- a/src/mongocrypt.h
+++ b/src/mongocrypt.h
@@ -947,7 +947,7 @@ mongocrypt_ctx_explicit_encrypt_init (mongocrypt_ctx_t *ctx,
                                       mongocrypt_binary_t *msg);
 
 /**
- * Explicit helper method to encrypt a Match Expression or Aggregate Expression
+ * Explicit helper method to encrypt a Match Expression or Aggregate Expression.
  * Contexts created for explicit encryption will not go through mongocryptd.
  * Requires query_type to be "rangePreview".
  * NOTE: The RangePreview algorithm is experimental only. It is not intended for


eramongodb (commented):
For clarification, what is the correlation between the term "expression" (specifically referring to the terminology being used, not the behavior of the operation) with encryption and searches? Is it only applicable to range searches, or does it also apply to equality searches? The documentation for `mongocrypt_ctx_explicit_encrypt_expression_init` suggests it refers to a BSON document expression, but that seems like it may be a bit too generic in its description?

kevinAlbs:
> For clarification, what is the correlation between the term "expression" (specifically referring to the terminology being used, not the behavior of the operation) with encryption and searches?

The terms "Match Expression" and "Aggregate Expression" come from [Technical Design: Structured Encryption Range Index support](https://docs.google.com/document/d/1JLrpDNP1wjJCj-9dkQN9wbvSMxKe9MqL0XNF8dFbMuk/edit#heading=h.eagh7zsfto82). "expression" is meant to group the two.

Expressions are used for searching. A Match Expression is used in some contexts (e.g. a find filter or $match stage) and an Aggregate Expression is used in others (e.g. inside $expr or a $project stage).

> Is it only applicable to range searches, or does it also apply to equality searches?

Encrypting an expression currently only applies to range searches. Though it seems plausible it could be extended for future search types.

> The documentation for `mongocrypt_ctx_explicit_encrypt_expression_init` suggests it refers to a BSON document expression, but that seems like it may be a bit too generic in its description?

Agreed. Documentation has been updated to clarify only a "Match Expression" and "Aggregate Expression" are supported.

eramongodb (approved):

src/mongocrypt.h
```diff
@@ -914,9 +914,48 @@ mongocrypt_ctx_encrypt_init (mongocrypt_ctx_t *ctx,
...
+ * @returns A boolean indicating success. If false, an error status is set.
+ * Retrieve it with @ref mongocrypt_ctx_status
+ */
+MONGOCRYPT_EXPORT
+bool
+mongocrypt_ctx_explicit_encrypt_init (mongocrypt_ctx_t *ctx,
+                                      mongocrypt_binary_t *msg);
+
+/**
+ * Explicit helper method to encrypt a Match Expression or Aggregate Expression
```
eramongodb, line 950:
```suggestion
 * Explicit helper method to encrypt a Match Expression or Aggregate Expression.
```

vector-of-bool (approved):
LGTM

PR #510, opened 2022-12-06
MONGOCRYPT-483: A 128-bit integer abstraction

vector-of-bool:
This changeset adds a new trivial type `mlib_int128`, which presents a method of performing 128-bit binary arithmetic. This is a prerequesite to MONGOCRYPT-483, which requires 128-bit integers and integer arithmetic.

Some platforms, such as 64-bit GCC, provide an `__int128` abstraction, but other platforms do not. MSVC and all 32-bit targets do not have such extensions. For this reason, this PR implements 128-bit arithmetic using only standard C99.

Usage looks like this:

- `mlib_int128` a typedef of a trivial type that occupies 16 bytes of storage. There is no distinction between signed/unsigned: Operations that depend on sign (for now, just integer comparison) require specifying whether you want a signed compare or an unsigned compare.
- `MLIB_INT128(N)` a macro that creates a 128-bit integer from an integer literal. The literal should not have a suffix. Negation within the macro "just works" (but may generate a false-positive compiler warning).
- `MLIB_INT128_CAST(N)` cast an arbitrary integral value `N` to a 128-bit integer.
- `MLIB_INT128_FROM_PARTS(Low64, High64)` the actual "constructor" for `mlib_in128`: Glues together the low bits and the high bits from two 64-bit integers.
- `MLIB_INT128_SMAX`, `MLIB_INT128_SMIN`, `MLIB_INT128_UMAX`: The signed-maximum, signed-minimum, and unsigned-maximum values for 128-bit integers.

The following functions all have an `mlib_int128_` prefix:

- `ucmp(L, R) -> int` and `scmp(L, R) -> int` Compare two integers as either unsigned or signed, respectively. Returns `n < 0`, `n > 0`, or `n = 0` for less, greater, and equal, respectively.
- `eq(L, R) -> bool` - Determine whether `L` and `R` are equal.
- `add(L, R) -> int128` - Return `L + R`. Overflow wraps.
- `sub(L, R) -> int128` - Returns `L - R`. Overflow wraps.
- `negate(N) -> int128` - Treat `N` as a signed integer. Return `-N`.
- `lshift(N, c) -> int128` - Return `N << c`
- `rshift(N, c) -> int128` - Return `N >> c`
- `bitor(L, R) -> int128` - Return `L | R`
- `mul(L, R) -> int128` - Return `L * R`. Overflow wraps.
- `div(L, R) -> int128` - Treats `L` and `R` as unsigned: Return `L / R` (round toward zero).
- `mod(L, R) -> int128` - Treats `L` and `R` as unsigned: Return `L % R` (remainder after integer division).
- `divmod(L, R)-> {quotient, remainder}` - Obtain both the quotient and remainder of an integer division.
- `pow10(N) -> int128` - Obtain the `N`th power of ten.
- `pow2(N) -> int128` - Obtain the `N`th power of two. (No general `pow()` is defined.)
- `from_string(S) -> int128` - Parse `S` as a string of decimal digits.
- `to_u64(N) -> uint64_t` - Return the low 64 bits of `N`.
- `format(N)` - Render the `N` as a string of decimal digits. Returns a struct containing the result array.

The implementations of `add`, `sub`, the bitshifts, to/from string, compare, and equality are all fairly straightforward extensions of 64-bit operations (with carry/borrow).

Multiplication here is implemented using Knuth's 4.3.1M multiplication algorithm (from _The Art of Computer Programming_). Here 4.3.1M is used to multiply the two low 64-bit words to obtain the 128-bit product. The high word of the result is then adjusted by two more regular 64-bit multiplications.

Division is the most complicated operation and took a lot of testing to ensure correct. It uses Knuth's 4.3.1D algorithm for arbitrary precision unsigned division (defined in `_mlibKnuth431D`). Certain optimizations and shortcuts were learned from the existing STL `constexpr` implementation. Future optimization to use actual intrinsics and hardware are not done here, since we just need these 128-bit integers and want platform-equivalence.

Most of the code has been marked as `constexpr`, and a lot of functionality is tested via `static_asserts` for failing CI as-fast-as-possible and immediate IDE feedback.

---

The PR tests are implemented in C++ and defined a test macro `CHECK`, since I miss Catch2 🥲.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -480,10 +480,11 @@ foreach (test IN ITEMS path str)
    target_link_libraries (mlib.${test}.test PRIVATE mongo::mlib)
 endforeach ()
 
-if ("cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+if ("cxx_relaxed_constexpr" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
    add_executable(mlib.int128.test src/mlib/int128.test.cpp)
    add_Test (mlib.int128 mlib.int128.test)
    target_link_libraries (mlib.int128.test PRIVATE mongo::mlib)
+   target_compile_features (mlib.int128.test PRIVATE cxx_relaxed_constexpr)
 endif ()
 
 if ("cxx_std_20" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -481,9 +481,9 @@ foreach (test IN ITEMS path str)
 endforeach ()
 
 if ("cxx_relaxed_constexpr" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-   add_executable(mlib.int128.test src/mlib/int128.test.cpp)
-   add_Test (mlib.int128 mlib.int128.test)
-   target_link_libraries (mlib.int128.test PRIVATE mongo::mlib)
+   add_executable(mlib.int128.test src/mlib/int128.test.cpp src/mlib/int128.test.c)
+   add_test (mlib.int128 mlib.int128.test)
+   target_link_libraries (mlib.int128.test PRIVATE mongo::mlib Threads::Threads)
    target_compile_features (mlib.int128.test PRIVATE cxx_relaxed_constexpr)
 endif ()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -480,6 +480,12 @@ foreach (test IN ITEMS path str)
    target_link_libraries (mlib.${test}.test PRIVATE mongo::mlib)
 endforeach ()
 
+if ("cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+   add_executable(mlib.int128.test src/mlib/int128.test.cpp)
+   add_Test (mlib.int128 mlib.int128.test)
+   target_link_libraries (mlib.int128.test PRIVATE mongo::mlib)
+endif ()
+
 if ("cxx_std_20" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
    add_executable (csfle-markup src/csfle-markup.cpp)
    target_link_libraries (csfle-markup PRIVATE mongocrypt_static _mongocrypt::libbson_for_static mongo::mlib)
--- a/src/mlib/int128.h
+++ b/src/mlib/int128.h
@@ -284,7 +284,8 @@ _mlibKnuth431D (uint32_t *const u,
                 const int vlen,
                 uint32_t *quotient)
 {
-   // Part D1 (normalization) is done by caller, normalized in u and v (b is 32)
+   // Part D1 (normalization) is done by caller,
+   // normalized in u and v (radix b is 2^32)
    typedef uint64_t u64;
    typedef int64_t i64;
    typedef uint32_t u32;
@@ -347,7 +348,7 @@ _mlibKnuth431D (uint32_t *const u,
       }
    }
 
-   // Denomralization (D8) is done by caller.
+   // Denormalization (D8) is done by caller.
 }
 
 /// The result of 128-bit division
@@ -379,7 +380,7 @@ _mlibDivide_u128_by_u64 (const mlib_int128 numer, const uint64_t denom)
       const u64 n0 = (u32) (numer.r.lo);
       const u64 n1 = (u32) (numer.r.lo >> 32);
 
-      // We don't need to split n2 and n3. (n3,n2) will be the first parital
+      // We don't need to split n2 and n3. (n3,n2) will be the first partial
       // dividend
       const u64 n3_n2 = numer.r.hi;
 
@@ -655,7 +656,7 @@ static mlib_constexpr_fn mlib_int128_charbuf
 mlib_int128_format (mlib_int128 i)
 {
    mlib_int128_charbuf into = {0};
-   char *out = into.str + sizeof into - 1;
+   char *out = into.str + (sizeof into) - 1;
    int len = 0;
    if (mlib_int128_eq (i, MLIB_INT128 (0))) {
       *out-- = '0';
--- a/src/mlib/int128.h
+++ b/src/mlib/int128.h
@@ -282,13 +282,15 @@ _mlibKnuth431D (uint32_t *const u,
                 const int vlen,
                 uint32_t *quotient)
 {
-   // Part D1 (normalization) is done by caller, normalized in u and v (b is
-   // 1<<32)
+   // Part D1 (normalization) is done by caller, normalized in u and v (b is 32)
    typedef uint64_t u64;
    typedef int64_t i64;
    typedef uint32_t u32;
    const int m = ulen - vlen - 1;
    const int n = vlen;
+
+   // 'd' is 2^32. Shifting left and right is equivalent to mult and division by
+   // d, respectively.
 
    // D2
    int j = m;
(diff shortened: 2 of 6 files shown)


src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+/**
+ * @brief A 128-bit binary integer
+ */
+typedef union {
+   struct {
+      uint64_t lo;
+      uint64_t hi;
+   } r;
+#if defined(__SIZEOF_INT128__)
+   // These union members are only for the purpoes of debugging visualization
```
addaleax, line 22: typo:
```suggestion
   // These union members are only for the purpose of debugging visualization
```

(Aside, this makes the ABI of this union depend on whether `__int128_t` is available, I assume that’s okay?)

vector-of-bool: No, but also yes, but also no: Practically no, because the union members are all 16 bytes, but yes: it may have 16byte alignment iff the __in128s are present, and technically an ODR violation if mixing TUs that differ on `__SIZEOF_INT128__`, but no: mixing TUs that would differ at that level (i.e. mixing compilers) would be a recipe for trouble, and this isn't exposed on the public API/ABI, so we're free to change it on a whim.

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+mlib_int128_rshift (mlib_int128 val, int off)
+{
+   return mlib_int128_lshift (val, -off);
+}
+
+/**
+ * @brief Bitwise-or two 128-bit integers
+ */
+static mlib_constexpr_fn mlib_int128
+mlib_int128_bitor (mlib_int128 l, mlib_int128 r)
```
kkloberdanz, line 219: I can't seem to find a test for this one. Perhaps we could have a test for it?

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+   while (!(1ull << 63 & bits)) {
+      ++n;
+      bits <<= 1;
+   }
+   return n;
+}
+
+/// Implementation of Knuth's algorithm 4.3.1 D for unsigned integer division
+static mlib_constexpr_fn void
+_mlibKnuth431D (uint32_t *const u,
```
kkloberdanz, line 279: I love the explicit citations here! D1, D2, etc

src/mlib/macros.h
```diff
@@ -14,4 +14,33 @@
...
+#define _mlibConstexprFn constexpr inline
+#else
+#define _mlibConstexprFn inline
+#endif
+
+/**
+ * @brief Mark a function as constexpr
+ *
+ * Expands to `constexpr` in C++14 and above (and someday C26...?). "inline"
+ * otherwise.
```
eramongodb, line 42:
```suggestion
 * Expands to `constexpr inline` in C++14 and above (and someday C26...?).
 * "inline" otherwise.
```

src/mlib/int128.test.cpp
```diff
@@ -0,0 +1,305 @@
...
+#include <iostream>
+#include <random>
+#include <string>
+
+#if (defined(__GNUC__) && __GNUC__ < 7) || \
+   (defined(_MSC_VER) && _MSC_VER < 1920)
+// Old GCC and old MSVC have partially-broken constexpr that prevents us from
+// properly using static_assert with from_string()
+#define BROKEN_CONSTEXPR
+#endif
```
eramongodb, line 12: Can the condition used for `_mlibConstexprFn` be used instead for consistency? It is unclear what the overlap / relationship is between the conditions used here vs. those used for `_mlibConstexprFn`.

CMakeLists.txt
```diff
@@ -480,6 +480,13 @@ foreach (test IN ITEMS path str)
    target_link_libraries (mlib.${test}.test PRIVATE mongo::mlib)
 endforeach ()
 
+if ("cxx_relaxed_constexpr" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+   add_executable(mlib.int128.test src/mlib/int128.test.cpp)
+   add_Test (mlib.int128 mlib.int128.test)
+   target_link_libraries (mlib.int128.test PRIVATE mongo::mlib)
```
eramongodb, line 486: Suggest adding a `int128.test.c` to verify C compatibility via compilation of the header as C code:

```c
#include "./int128.h"

int
main (void)
{
   return 0;
}
```

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+   struct {
+      uint64_t lo;
+      uint64_t hi;
+   } r;
+#if defined(__SIZEOF_INT128__)
+   // These union members are only for the purpose of debugging visualization
+   // and testing, and will only appear correctly on little-endian platforms.
+   __int128_t signed_;
+   __uint128_t unsigned_;
+#endif
```
eramongodb, line 26: Accessing an inactive union data member is UB according to the C++ standard. Can these be omitted in favor of bit-cast (memcpy) to/from `__(u)int128_t` where applicable (only in `int128.test.cpp`)? e.g.:
```cpp
#ifdef __SIZEOF_INT128__
      // When we have an existing i128 impl, test against that:
      __uint128_t num_u128;
      memcpy (&num_u128, &num, sizeof (__uint128_t));

      __uint128_t denom_u128;
      memcpy (&denom_u128, &denom, sizeof (__uint128_t));

      __uint128_t q_u128 = num_u128 / denom_u128;
      __uint128_t r_u128 = num_u128 % denom_u128;

      mlib_int128 quotient;
      memcpy (&quotient, &q_u128, sizeof (__uint128_t));

      mlib_int128 remainder;
      memcpy (&remainder, &r_u128, sizeof (__uint128_t));

      CHECK (q.quotient == quotient);
      CHECK (q.remainder == remainder);
#endif
```
This would also reduce complexity/noise for `mlib_u128` (can be made a simple struct).

vector-of-bool: I think the `memcpy`/`bit_cast` is a good idea in the test cases. I prefer to leave the union members because they have been life-saving while debugging.

src/mlib/int128.test.cpp
vector-of-bool: Compilers that match this condition above also declare themselves to have relaxed-constexpr, but it's utterly broken. It would probably be best the other way around, or at least to conditional-ize the "broken" function.

For reference, I whittled it down to this before I realized that the compiler had been gaslighting me:

```c++
// ↓↓↓↓↓ Delete this line, and the static_assert fails: 
constexpr auto completely_unrelated_variable = mlib_int128_mul(NINE, TEN);
// ↑↑↑↑↑ wat ‽

static_assert(mlib_int128_from_string("99999") 
                == mlib_int128(MLIB_INT128_FROM_PARTS(99'999ull, 0)),
              "pain");
```

src/mlib/int128.test.cpp
```diff
@@ -0,0 +1,305 @@
+#include "./int128.h"
+
+#include <iostream>
+#include <random>
+#include <string>
+
+#if (defined(__GNUC__) && __GNUC__ < 7) || \
+   (defined(_MSC_VER) && _MSC_VER < 1920)
```
eramongodb, line 8:
```suggestion
#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 7) || \
   (defined(_MSC_VER) && _MSC_VER < 1920)
```
This might be unnecessarily ruling out Clang according to local tests where my Clang 10 compiler is reporting `__GNUC__ == 4`.

src/mlib/int128.test.cpp
```diff
@@ -0,0 +1,305 @@
...
+          "170141183460469231731687303715884105728");
+
+   for (int i = 0; i < 100000; ++i) {
+      std::uniform_int_distribution<std::uint64_t> dist;
+      std::random_device r;
+      auto bits = r ();
+      mlib_int128 num = MLIB_INT128_FROM_PARTS (bits & 0b0001 * dist (r),
+                                                bits & 0b0010 * dist (r));
+      mlib_int128 denom = MLIB_INT128_FROM_PARTS (bits & 0b0100 * dist (r),
+                                                  bits & 0b1000 * dist (r));
```
eramongodb, line 288: Unclear what the function of the bitmasks are in this for loop. Suggest adding clarifying comments (why the bitmask prior to multiply by random `uint64_t`? Why not a direct uniform `uint64_t`?).

Also suggest moving random device and distribution out of for loop.

src/mlib/int128.test.cpp
vector-of-bool: [pain.](https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers)

Let's just hope that GCC never decides to "pretend" to be Clang as well.

src/mlib/int128.test.cpp
```diff
@@ -0,0 +1,305 @@
...
+      if (mlib_int128_eq (denom, 0_i128)) {
+         continue;
+      }
+
+      // Attempt a division, even if we have no "reference" value
+      auto q = mlib_int128_divmod (num, denom);
+      (void) q;
+
+#ifdef __SIZEOF_INT128__
+      // When we have an existing i128 impl, test against that:
```
kevinAlbs, line 298: Neat idea. Consider fuzz testing division, and comparing the result with the native support. That may help to explore more code paths.

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+      (uint32_t) (adjusted.r.lo >> (32 - d)),
+      (uint32_t) (adjusted.r.hi << d),
+      (uint32_t) (adjusted.r.hi >> (32 - d)),
+      0,
+   };
+
+   if (d != 0) {
+      // Extra bits from overlap:
+      u[2] |= (uint32_t) (adjusted.r.lo >> (64 - d));
+      u[4] |= (uint32_t) (adjusted.r.lo >> (64 - d));
```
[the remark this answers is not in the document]
vector-of-bool: Yes, that's a problem.

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+   for (;;) {
+      // D3: Select two u32 as a u64:
+      u64 two = ((u64) (u[j + n]) << 32) | u[j + n - 1];
+      // D3: Partial quotient: q̂
+      u64 q = two / v[n - 1];
+      // D3: Partial remainder: r̂
+      u64 r = two % v[n - 1];
+
+      // D3: Compute q̂ and r̂
+      while (q >> 32 || q * (u64) v[n - 2] > (r << 32 | u[j + n - 2])) {
```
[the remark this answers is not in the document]
vector-of-bool: Either way will be fine, yes. I also tried both. The test with the `q>>32` is the one I've seen more often (probably a trivial optimization).

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+#if defined(__SIZEOF_INT128__)
+   // These union members are only for the purpose of debugging visualization
+   // and testing, and will only appear correctly on little-endian platforms.
+   __int128_t signed_;
+   __uint128_t unsigned_;
+#endif
+} mlib_int128;
+
+/// Construct an mlib_in128 from an integer literal
+#define MLIB_INT128(N) MLIB_INIT (mlib_int128) MLIB_INT128_C (N)
```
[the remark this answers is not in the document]
vector-of-bool: That's not expected, no. Interesting...

After lots of poking and prodding, I am unable to make it do "the expected thing" without also generating a compiler warning about big integer literals. I'll add a comment about the range.

src/mlib/int128.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+   std::seed_seq seed ({rd (), rd (), rd (), rd ()});
+   // Pick every numerator bit pattern from 0b00'00 to 0b11'11
+   for (auto nbits = 0u; nbits < 16u; ++nbits) {
+      // This is an extremely rudimentary thread pool to parallelize the
+      // division checks. It doesn't need to be rigorous or optimal, it only
+      // needs to "just work."
+      std::vector<std::thread> threads;
+      threads.resize (15);
+      // Pick every denominator bit pattern from 0b00'01 to 0b11'11:
+      for (auto dbits = 1u; dbits < 16u; ++dbits) {
```
eramongodb, line 384: Confirming, `[1, 16)`? Not `[0, 16)`?

eramongodb (approved):
Some minor suggestions and questions remaining; otherwise, LGTM.

src/mlib/int128.test.cpp
vector-of-bool: Correct. dbits=0 would generate a zero divisor forever.

src/mlib/int128.h
```diff
@@ -0,0 +1,683 @@
...
+      // jk: We're dividing by less than UINT32_MAX: We can do a simple short
+      // division of two base32 numbers.
+      // Treat the denominator as a single base32 digit:
+      const u32 d0 = (u32) denom;
+
+      // And the numerator as four base32 digits:
+      const u64 n0 = (u32) (numer.r.lo);
+      const u64 n1 = (u32) (numer.r.lo >> 32);
+
+      // We don't need to split n2 and n3. (n3,n2) will be the first parital
```
kevinAlbs, line 382:
```suggestion
      // We don't need to split n2 and n3. (n3,n2) will be the first partial
```

src/mlib/int128.h
```diff
@@ -0,0 +1,683 @@
...
+      }
+
+      // D7:
+      --j;
+      if (j < 0) {
+         break;
+      }
+   }
+
+   // Denomralization (D8) is done by caller.
```
kevinAlbs, line 350:
```suggestion
   // Denormalization (D8) is done by caller.
```

src/mlib/int128.h
```diff
@@ -0,0 +1,614 @@
...
+/// Divide a 128-bit number by a 64bit number.
+static mlib_constexpr_fn struct mlib_int128_divmod_result
+_mlibDivide_u128_by_u64 (const mlib_int128 numer, const uint64_t denom)
+{
+   mlib_int128 adjusted = numer;
+   adjusted.r.hi %= denom;
+   int d = _mlibCountLeadingZeros_u64 (denom >> 32) - 32;
+
+   if (d == 32) {
+      // jk: We're dividing by less than UINT32_MAX: We can take a shortcut
```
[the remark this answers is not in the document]
kevinAlbs: The comments are very helpful.

src/mlib/int128.h
```diff
@@ -0,0 +1,683 @@
...
+
+/// Implementation of Knuth's algorithm 4.3.1 D for unsigned integer division
+static mlib_constexpr_fn void
+_mlibKnuth431D (uint32_t *const u,
+                const int ulen,
+                const uint32_t *const v,
+                const int vlen,
+                uint32_t *quotient)
+{
+   // Part D1 (normalization) is done by caller, normalized in u and v (b is 32)
```
kevinAlbs, line 287:
```suggestion
   // Part D1 (normalization) is done by caller, normalized in u and v (b is 2^32)
```

b is the radix. The digits are 32 bit unsigned integers.

kevinAlbs (approved):
Great work. LGTM

PR #513, opened 2022-12-07
ci(NODE-4666): pin to node18 instead of latest

baileympearson:
### Description

This PR pins the Node CI the Node 18 (the current latest LTS node version).  It also makes two other small changes (see the notes sections below).

A requirement of the ticket is that we make sure our release will be pinned to Node16 since they build on rhel7.  This was already addressed and can be confirmed in the config.yml (see [here](https://github.com/mongodb/libmongocrypt/blob/4a520d7cccee09d3d4ce4cec207701f31e050c7e/.evergreen/config.yml#L1130)).

Note: you may see failing socks5 tests on the CI.  These tests are also failing on libmongocrypt master occasionally.

### Misc Other Changes

https://github.com/mongodb/libmongocrypt/commit/216f7854cbee5dbd9247099b6db88c38d9d2c53c added debian and ubuntu22 support but did not add `build-and-test-node-no-optional-dependencies` to the new variants.  This PR adds the missing tasks.

Switching to a pinned version instead of `lts` cause us to encounter this issue - https://github.com/coreybutler/nvm-windows/issues/706.  Windows nvm 1.1.10 has a fix for this issue so windows nvm was bumped to the latest version.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1059,6 +1059,7 @@ buildvariants:
   - build-and-test-shared-bson
   - build-and-test-asan
   - build-and-test-node
+  - build-and-test-node-no-optional-dependencies
   - test-java
   - name: publish-packages
     distros:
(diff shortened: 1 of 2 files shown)


addaleax (approved):

durran (approved):

durran:
Evg failures related to socks.

PR #514, opened 2022-12-08
MONGOCRYPT-506 Extend token API for server needs

markbenvenuto:
Server needs to be able to construct a libmongocrypt token `mc_ServerDataEncryptionLevel1Token_t` from a server token (i.e. `mongo::ServerDataEncryptionLevel1Token`). We use a raw buffer for this but let the server own the buffer rather then deep copy it in this case.

Feel free to suggest a better name.

### DIFF
--- a/test/test-mc-tokens.c
+++ b/test/test-mc-tokens.c
@@ -226,6 +226,11 @@ _test_mc_tokens_raw_buffer (_mongocrypt_tester_t *tester)
 
    ASSERT_CMPBUF (*mc_ServerDataEncryptionLevel1Token_get (token), expected);
 
+   /* Assert new_from_buffer references original buffer instead of a copy. */
+   test_input.data[0] = '0';
+   expected.data[0] = '0';
+   ASSERT_CMPBUF (*mc_ServerDataEncryptionLevel1Token_get (token), expected);
+
    _mongocrypt_buffer_cleanup (&test_input);
    _mongocrypt_buffer_cleanup (&expected);
    mc_ServerDataEncryptionLevel1Token_destroy (token);
--- a/test/test-mc-tokens.c
+++ b/test/test-mc-tokens.c
@@ -202,9 +202,38 @@ _test_mc_tokens_error (_mongocrypt_tester_t *tester)
    mongocrypt_status_destroy (status);
 }
 
+static void
+_test_mc_tokens_raw_buffer (_mongocrypt_tester_t *tester)
+{
+   mc_ServerDataEncryptionLevel1Token_t *token;
+   _mongocrypt_buffer_t test_input;
+   _mongocrypt_buffer_t expected;
+
+   _mongocrypt_buffer_copy_from_hex (
+      &test_input,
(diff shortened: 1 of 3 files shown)


eramongodb (commented):
I have reservations about continuing to expand the set of interfaces with mixed ownership semantics (conditionally owning vs. non-owning; conditional value vs. reference semantics), but that is a separate issue. LGTM as proposed, but would like to suggest the inclusion of test cases to cover this new interface.

test/test-mc-tokens.c
```diff
@@ -202,9 +202,38 @@ _test_mc_tokens_error (_mongocrypt_tester_t *tester)
...
+
+   _mongocrypt_buffer_copy_from_hex (
+      &test_input,
+      "6c6a349956c19f9c5e638e612011a71fbb71921edb540310c17cd0208b7f548b");
+
+   /* Make a token from a raw buffer */
+   token = mc_ServerDataEncryptionLevel1Token_new_from_buffer (&test_input);
+   /* Assert the new_from_buffer does not steal the buffer */
+   ASSERT (test_input.owned == true);
+   ASSERT (test_input.len == MONGOCRYPT_HMAC_SHA256_LEN);
```
eramongodb, line 220:
```suggestion

   /* Assert new_from_buffer did not steal ownership. */
   ASSERT (test_input.owned);
   ASSERT (test_input.len == MONGOCRYPT_HMAC_SHA256_LEN);
```

markbenvenuto:
Add test case.

After I wrote a POC for the actual inclusion of libmongocrypt code, it was a simple situation where I could cut down on the number of allocations.

POC Example
https://github.com/markbenvenuto/mongo/blob/74e3af94de5f706ed4faa7978535f2ac34ec845a/src/mongo/crypto/fle_crypto.cpp#L2745-L2775

Server for instance inlined all the PrfBlock tokens into the struct. The libmongocrypt version all separate heap allocations. Just different designs.

test/test-mc-tokens.c
```diff
@@ -202,9 +202,38 @@ _test_mc_tokens_error (_mongocrypt_tester_t *tester)
...
+
+   _mongocrypt_buffer_copy_from_hex (
+      &expected,
+      "6c6a349956c19f9c5e638e612011a71fbb71921edb540310c17cd0208b7f548b");
+
+   ASSERT_CMPBUF (*mc_ServerDataEncryptionLevel1Token_get (token), expected);
```
eramongodb, line 226: Suggest adding test validating non-owning state of `mc_ServerDataEncryptionLevel1Token_new_from_buffer`, e.g.:

```suggestion
   ASSERT_CMPBUF (*mc_ServerDataEncryptionLevel1Token_get (token), expected);

   /* Assert new_from_buffer references original buffer instead of a copy. */
   test_input.data[0] = '0';
   expected.data[0] = '0';
   ASSERT_CMPBUF (*mc_ServerDataEncryptionLevel1Token_get (token), expected);
```

kevinAlbs (approved):

eramongodb (approved):

PR #516, opened 2022-12-09
fix(NODE-4880): specify family of ipv4 when connecting to default mongocryptd uri

baileympearson:
Node18 changed the default dns resolution ordering from "ipv4 always first" to "returned in the order that the dns resolver returns them in".  This breaks some environments that use mongocryptd, since by default mongocryptd does not listen on ipv6 addresses.

There are workarounds

- configure mongocryptd to listen on ipv6 addresses
- specify a `mongocryptdURI` that uses `127.0.0.1` instead of `localhost`
- set the resolution globally on the node process

But none of these are ideal for customers or for driver CI.

This PR adds logic to force the driver to use ipv4 addresses when no `mongocryptdURI` is specified.

Driver CI task successfully passing with these changes - https://spruce.mongodb.com/version/639752c0c9ec447a54d40daa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC.

### DIFF
--- a/bindings/node/.eslintrc.json
+++ b/bindings/node/.eslintrc.json
@@ -10,7 +10,7 @@
     "es6": true
   },
   "parserOptions": {
-    "ecmaVersion": 2022
+    "ecmaVersion": 2019
   },
   "plugins": [
     "prettier"
--- a/bindings/node/.eslintrc.json
+++ b/bindings/node/.eslintrc.json
@@ -10,7 +10,7 @@
     "es6": true
   },
   "parserOptions": {
-    "ecmaVersion": 2019
+    "ecmaVersion": 2022
   },
   "plugins": [
     "prettier"
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -165,7 +165,7 @@ module.exports = function (modules) {
           serverSelectionTimeoutMS: 10000
         };
 
-        const wasURIProvided = typeof options.extraOptions === 'object' && !!options.extraOptions.mongocryptdURI;
+        const wasURIProvided = options.extraOptions && !!options.extraOptions.mongocryptdURI;
         if (!wasURIProvided) {
           clientOptions.family = 4;
         }
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -165,8 +165,7 @@ module.exports = function (modules) {
           serverSelectionTimeoutMS: 10000
         };
 
-        const wasURIProvided = options.extraOptions && !!options.extraOptions.mongocryptdURI;
-        if (!wasURIProvided) {
+        if (options.extraOptions == null || typeof options.extraOptions.mongocryptdURI !== 'string')
           clientOptions.family = 4;
         }
 
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -158,12 +158,22 @@ module.exports = function (modules) {
       // Only instantiate mongocryptd manager/client once we know for sure
       // that we are not using the CSFLE shared library.
       if (!this._bypassMongocryptdAndCryptShared && !this.cryptSharedLibVersionInfo) {
+        const wasUriProvided =
+          typeof options.extraOptions === 'object' && !!options.extraOptions.mongocryptdURI;
         this._mongocryptdManager = new MongocryptdManager(options.extraOptions);
-        this._mongocryptdClient = new MongoClient(this._mongocryptdManager.uri, {
+        const clientOptions = {
           useNewUrlParser: true,
           useUnifiedTopology: true,
           serverSelectionTimeoutMS: 10000
-        });
+        };
+
+        if (
+          this._mongocryptdManager.uri === MongocryptdManager.DEFAULT_MONGOCRYPTD_URI &&
+          !wasUriProvided
+        ) {
+          clientOptions.family = 4;
+        }
+        this._mongocryptdClient = new MongoClient(this._mongocryptdManager.uri, clientOptions);
       }
     }
 
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -165,7 +165,10 @@ module.exports = function (modules) {
           serverSelectionTimeoutMS: 10000
         };
 
-        if (options.extraOptions == null || typeof options.extraOptions.mongocryptdURI !== 'string')
+        if (
+          options.extraOptions == null ||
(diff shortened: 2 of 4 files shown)


bindings/node/lib/mongocryptdManager.js
```diff
@@ -7,6 +7,8 @@ const spawn = require('child_process').spawn;
  * An internal class that handles spawning a mongocryptd.
  */
 class MongocryptdManager {
+  static DEFAULT_MONGOCRYPTD_URI = 'mongodb://localhost:27020/?serverSelectionTimeoutMS=1000';
```
nbbeeken, line 10: Why aren't we changing the default to be 127.0.0.1?

bindings/node/lib/mongocryptdManager.js
```diff
@@ -22,7 +24,7 @@ class MongocryptdManager {
     } else {
       // TODO: eventually support connecting on Linux Socket for non-windows,
       // blocked by SERVER-41029
```
nbbeeken, line 26: Might as well clear this up, it was closed as a won't fix

bindings/node/lib/mongocryptdManager.js
```diff
@@ -44,7 +46,7 @@ class MongocryptdManager {
   /**
    * @ignore
    * Will check to see if a mongocryptd is up. If it is not up, it will attempt
-   * to spawn a mongocryptd in a detached process, and then wait for it to be up.
+   * to spawn a mongocryptd in a detached plocess, and then wait for it to be up.
```
nbbeeken, line 49: plocess? 😅

bindings/node/lib/autoEncrypter.js
```diff
@@ -159,11 +159,18 @@ module.exports = function (modules) {
...
         this._mongocryptdManager = new MongocryptdManager(options.extraOptions);
-        this._mongocryptdClient = new MongoClient(this._mongocryptdManager.uri, {
+        const clientOptions = {
           useNewUrlParser: true,
           useUnifiedTopology: true,
           serverSelectionTimeoutMS: 10000
-        });
+        };
+
+        const wasURIProvided = typeof options.extraOptions === 'object' && !!options.extraOptions.mongocryptdURI;
```
nbbeeken, line 168: I think you need a null check here before accessing the `mongocryptdURI` key

nbbeeken (changes requested):

bindings/node/lib/mongocryptdManager.js
baileympearson: the default uri is actually spec'd https://mongodb.slack.com/archives/C72LB5RPV/p1670607134515969?thread_ts=1670604198.815689&cid=C72LB5RPV

bindings/node/lib/autoEncrypter.js
baileympearson: That's out of the bounds of our types, both in the jsdoc types and Typescript.  Is this something you still think we should guard against?

bindings/node/lib/mongocryptdManager.js
baileympearson: I don't have context on this ticket.  Do you want the logic removed or just the comment?

bindings/node/lib/mongocryptdManager.js
```diff
@@ -7,6 +7,8 @@ const spawn = require('child_process').spawn;
  * An internal class that handles spawning a mongocryptd.
  */
 class MongocryptdManager {
+  static DEFAULT_MONGOCRYPTD_URI = 'mongodb://localhost:27020';
```
baileympearson, line 10: aligns the default URI with the spec

bindings/node/lib/autoEncrypter.js
baileympearson: Yup, but our type here is explicitly "string", so null wouldn't be allowed.  This gets to a larger question about whether we should test outside the bounds of what our types declare that we support.  

I updated the check to match the other checks in the file though

bindings/node/lib/mongocryptdManager.js
```diff
@@ -15,15 +17,7 @@ class MongocryptdManager {
...
-    // TODO: this is not actually supported by the spec, so we should clarify
-    // with the spec or get rid of this
-    if (extraOptions.mongocryptdURI) {
-      this.uri = extraOptions.mongocryptdURI;
-    } else {
-      // TODO: eventually support connecting on Linux Socket for non-windows,
-      // blocked by SERVER-41029
-      this.uri = 'mongodb://localhost:27020/?serverSelectionTimeoutMS=1000';
-    }
+    this.uri = extraOptions.mongocryptdURI ?? MongocryptdManager.DEFAULT_MONGOCRYPTD_URI;
```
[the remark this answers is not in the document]
baileympearson: Ah that's a good point.

I adjusted the logic to account for the empty string.

bindings/node/lib/autoEncrypter.js
```diff
@@ -159,11 +159,18 @@ module.exports = function (modules) {
...
-        this._mongocryptdClient = new MongoClient(this._mongocryptdManager.uri, {
+        const clientOptions = {
           useNewUrlParser: true,
           useUnifiedTopology: true,
           serverSelectionTimeoutMS: 10000
-        });
+        };
+
+        const wasURIProvided = options.extraOptions && !!options.extraOptions.mongocryptdURI;
+        if (!wasURIProvided) {
```
nbbeeken, line 169:
```suggestion
        if (options.extraOptions == null || typeof options.extraOptions.mongocryptdURI !== 'string')
```

Just a suggestion, just to read this as "there's no extra options, nor is there a URI string provided" so we set family to 4, but totally optional

bindings/node/.eslintrc.json
```diff
@@ -10,7 +10,7 @@
     "es6": true
   },
   "parserOptions": {
-    "ecmaVersion": 2019
+    "ecmaVersion": 2022
```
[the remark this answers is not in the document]
nbbeeken: Is there a setting that can help us keep compat with node 12?

nbbeeken (changes requested):

bindings/node/lib/mongocryptdManager.js
```diff
@@ -66,4 +61,8 @@ class MongocryptdManager {
   }
 }
 
+Object.defineProperty(MongocryptdManager, 'DEFAULT_MONGOCRYPTD_URI', {
```
nbbeeken, line 64: Unusual for us to use defineProperty but not a blocker

nbbeeken (approved):

addaleax (approved):

nbbeeken (approved):

PR #517, opened 2022-12-12
MONGOCRYPT-513 Drop RHEL 6.7 on S390X

kevinAlbs:
# Summary

- Remove `rhel-67-s390x` variant.
- Replace `rhel-67-s390x` with `rhel72-zseries-test` in Java tasks.

# Background & Motivation

Building on RHEL 6.7 on S390X adds difficulty for upgrading cmake. Upgrading cmake may be helpful for integrating the Intel Decimal Floating-Point Math Library library as part of MONGOCRYPT-483.

The latest documented server version to support RHEL 6 on S390X is MongoDB Server 4.2.17. Automatic encryption needs mongocryptd or the crypt_shared library on the same machine as the application. So it is already not possible to do automatic encryption with version > 4.2.17 of mongocryptd / crypt_shared.

RHEL 6 is End of life (EOL). Redhat does not support it any longer. Server does not support RHEL 6 for MongoDB 5.0 or later.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -524,7 +524,7 @@ tasks:
   depends_on:
     - variant: rhel-62-64-bit
       name: build-and-test-and-upload
-    - variant: rhel-67-s390x
+    - variant: rhel72-zseries-test
       name: build-and-test-and-upload
     - variant: rhel-71-ppc64el
       name: build-and-test-and-upload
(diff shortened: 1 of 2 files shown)


rcsanchez97 (approved):
LGTM!

rozza (approved):
LGTM

PR #518, opened 2022-12-13
Evergreen link updates

rcsanchez97:
- remove link from public README
- in release procedure documentation note that auth is required

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ cd libmongocrypt
 ./cmake-build/test-mongocrypt
 ```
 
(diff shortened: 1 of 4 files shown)


kevinAlbs (approved):

rcsanchez97:
For those who might be wondering about the rationale behind these changes, the Evergreen UI is planned to no longer be accessible by the public. See: email with subject “Update: Evergreen Auth Policy”

jyemin (approved):

ShaneHarvey (approved):

PR #526, opened 2023-01-05
MONGOCRYPT-507 Add support for full deserialization of _mc_FLE2IndexedEqualityEncryptedValueTokens

shreyaskalyan:


src/mc-fle2-payload-iev.c
```diff
@@ -83,6 +88,7 @@ mc_FLE2IndexedEncryptedValue_parse (mc_FLE2IndexedEncryptedValue_t *iev,
    /* Read original_bson_type. */
    CHECK_AND_RETURN (
       mc_reader_read_u8 (&reader, &iev->original_bson_type, status));
+// TODO - should we validate this, server did
```
kevinAlbs, line 91: Is this intended to be addressed in this PR?
```suggestion
// TODO - check iev->original_bson_type is a valid BSON type
```

src/mc-fle2-payload-iev.c
```diff
@@ -151,36 +153,60 @@ mc_FLE2IndexedEncryptedValue_add_S_Key (_mongocrypt_crypto_t *crypto,
...
+                                       bool(callback)(mc_reader_t*, void*, mongocrypt_status_t* status),
+                                        void* callback_data,
+                                        mongocrypt_status_t *status)
+{
+   BSON_ASSERT_PARAM (crypto);
+   BSON_ASSERT_PARAM (iev);
+   BSON_ASSERT_PARAM (token);
+
+   if (!iev->parsed) {
+      CLIENT_ERR ("mc_FLE2IndexedEncryptedValue_add_S_Key must be "
```
kevinAlbs, line 183:
```suggestion
      CLIENT_ERR ("mc_FLE2IndexedEncryptedValue_decrypt must be "
```

src/mc-fle2-payload-iev.c
```diff
@@ -151,36 +153,60 @@ mc_FLE2IndexedEncryptedValue_add_S_Key (_mongocrypt_crypto_t *crypto,
...
+                  "called after mc_FLE2IndexedEncryptedValue_parse");
+      return false;
+   }
+
+   if (iev->inner_decrypted) {
+      CLIENT_ERR ("mc_FLE2IndexedEncryptedValue_add_S_Key must not be "
```
kevinAlbs, line 189:
```suggestion
      CLIENT_ERR ("mc_FLE2IndexedEncryptedValue_decrypt must not be "
```

src/mc-fle2-payload-iev.c
```diff
@@ -151,36 +153,60 @@ mc_FLE2IndexedEncryptedValue_add_S_Key (_mongocrypt_crypto_t *crypto,
...
+
+   return ret;
+}
+
+static bool
+mc_FLE2IndexedEncryptedValue_decrypt (_mongocrypt_crypto_t *crypto,
+                                        mc_FLE2IndexedEncryptedValue_t *iev,
+                                        mc_ServerDataEncryptionLevel1Token_t* token,
+                                       bool(callback)(mc_reader_t*, void*, mongocrypt_status_t* status),
+                                        void* callback_data,
```
kevinAlbs, line 175:
```suggestion
                                       mc_FLE2IndexedEqualityEncryptedValueTokens* tokens /* optional */,
```

Instead of a callback, can `tokens` be an optional out parameter?

shreyaskalyan (commented):
RFAL

src/mc-fle2-payload-iev.c
kevinAlbs: I prefer it here since there is only one caller that sets the callback.

kevinAlbs (approved):

src/mc-fle2-payload-iev.c
```diff
@@ -84,6 +90,14 @@ mc_FLE2IndexedEncryptedValue_parse (mc_FLE2IndexedEncryptedValue_t *iev,
    CHECK_AND_RETURN (
       mc_reader_read_u8 (&reader, &iev->original_bson_type, status));
 
+   /* Check that original_bson_type is a valid bson type. */
+   if ((iev->original_bson_type < 0) || (iev->original_bson_type > 0xFF)) {
```
vector-of-bool, line 94: I don't believe this condition can ever be met, because `original_bson_type` only has a range of `[0, 0xff]`. This will allow all 256 byte values through, but most of those are not valid type tags.

src/mongo_crypt-v1.h
```diff
@@ -23,12 +23,12 @@
```
vector-of-bool, line 23: Are there changes to this file other than formatting changes? It was copied directly from the upstream code, and I'd prefer to keep it identical to that file so that changes here are clear in the file history. Potentially want to add a `// clang-format off` to the top of this file.

vector-of-bool (commented):
Only one blocking concern in `mc-fle2-payload-iev.c`. One nit regarding file formatting of `mongo_crypt-v1.h`. Otherwise looks good.

src/mongo_crypt-v1.h
shreyaskalyan: Reverted file and added clang-format off to the top.

vector-of-bool (approved):

PR #531, opened 2023-01-17
PYTHON-3564 Add support for rangePreview queries

ShaneHarvey:
Add support for rangePreview queries. Changes include:
- Add support for the range_options and encrypt_expression parameters to ExplicitEncrypter.encrypt(). 
- Update header bindings for libmongocrypt 1.7.0-alpha1.
- Call mongocrypt_ctx_setopt_algorithm_range to set the range_options if provided.
- Call mongocrypt_ctx_explicit_encrypt_expression_init if encrypt_expression=true is provided.
- Added a basic test. More tests will be implemented in pymongo via the spec tests.

Open questions:
- Any opinions on the range_options and encrypt_expression names? Note these will only be used internally by pymongo and not exposed to users.
- Should we bump _MIN_LIBMONGOCRYPT_VERSION (currently set to 1.5.2)? If the user tries to use rangePreview with <1.7.0 they will get this error:
```python
ERROR: test_range_query_int32 (test.test_mongocrypt.TestExplicitEncryption)
  File "/Users/shane/git/libmongocrypt/bindings/python/test/test_mongocrypt.py", line 790, in test_range_query_int32
    contention_factor=4, range_options=range_opts, encrypt_expression=True)
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/explicit_encrypter.py", line 230, in encrypt
    with self.mongocrypt.explicit_encryption_context(value, opts) as ctx:
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/mongocrypt.py", line 344, in explicit_encryption_context
    self._create_context(), self.__opts.kms_providers, value, opts)
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/mongocrypt.py", line 550, in __init__
    self._raise_from_status()
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/mongocrypt.py", line 428, in _raise_from_status
    raise exc
pymongocrypt.errors.MongoCryptError: unsupported algorithm string "rangePreview"
```

If we bump the version to 1.7.0-alpha1 they'll get this error instead:
```python
ERROR: test_range_query_int32 (test.test_mongocrypt.TestExplicitEncryption)
  File "/Users/shane/git/libmongocrypt/bindings/python/test/test_mongocrypt.py", line 782, in test_range_query_int32
    kms_reply=http_data('kms-reply.txt')), self.mongo_crypt_opts())
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/explicit_encrypter.py", line 142, in __init__
    self.mongocrypt = MongoCrypt(mongo_crypt_opts, callback)
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/mongocrypt.py", line 189, in __init__
    self.__crypt = lib.mongocrypt_new()
  File "/Users/shane/git/libmongocrypt/bindings/python/pymongocrypt/binding.py", line 1537, in __getattr__
    raise self._error
RuntimeError: Expected libmongocrypt version 1.7.0-alpha1 or greater, found 1.6.1
```


bindings/python/pymongocrypt/explicit_encrypter.py
```diff
@@ -194,19 +207,26 @@ def encrypt(self, value, algorithm, key_id=None, key_alt_name=None,
           - `query_type` (str): The query type to execute.
           - `contention_factor` (int): The contention factor to use
             when the algorithm is "Indexed".
+          - `range_options` (bytes): Options for explicit encryption
+            with the "rangePreview" algorithm encoded as a BSON document.
+          - `encrypt_expression` (boolean): True if this is an
```
blink1073, line 212: This sounds like it should be the expression itself.  Perhaps `is_encrypt_expression`?

blink1073:
I think we should bump `_MIN_LIBMONGOCRYPT_VERSION` as well for a better user experience.

bindings/python/pymongocrypt/explicit_encrypter.py
```diff
@@ -30,9 +31,15 @@ def __init__(self, algorithm, key_id=None, key_alt_name=None,
           - `query_type` (str): The query type to execute.
           - `contention_factor` (int): The contention factor to use
             when the algorithm is "Indexed".
+          - `range_options` (bytes): Options for explicit encryption
```
juliusgeo, line 34: I feel like a longer name like `query_range_options` would be better because it only works when you have `query_type="rangePreview"` according to the spec.

blink1073 (approved):
LGTM!

bindings/python/pymongocrypt/explicit_encrypter.py
ShaneHarvey: I prefer not to make this name longer. Longer names are usually worse and these apis are already very wordy. https://journal.stuffwithstuff.com/2016/06/16/long-names-are-long/ is a great blog post on the subject of naming.

In this case I'd say the "range" in "range_opts" already qualifies this as only pertaining to rangePreview queries. The C driver uses "rangeopts" but we usually add an underscore between words so I shortened the name to "range_opts".

blink1073:
Yeah, I had the same train of thought.

blink1073 (approved):

PR #533, opened 2023-01-19
feat(NODE-4694): support range indexes

baileympearson:
This PR adds range index support in explicit encryption.

Changes included

- ClientEncryption.encrypt() logic was refactored to a private helper to be shared between encrypt() and encryptExpression()
- A new option, expressionMode, is passed down from _encrypt() into the bindings.  This toggles whether or not we are encrypting an expression or a regular bson value.
- maybeCallback was copied into the bindings
- documentation was generated

### DIFF
--- a/bindings/node/README.md
+++ b/bindings/node/README.md
@@ -56,9 +56,25 @@ npm test
 </dd>
 </dl>
 
+## Functions
+
+<dl>
+<dt><a href="#loadCredentials">loadCredentials(kmsProviders)</a> ⇒ <code>Promise</code></dt>
+<dd><p>Load cloud provider credentials for the user provided kms providers.
+Credentials will only attempt to get loaded if they do not exist
+and no existing credentials will get overwritten.</p>
+</dd>
+</dl>
+
 ## Typedefs
 
 <dl>
+<dt><a href="#BSONValue">BSONValue</a> : <code>*</code></dt>
+<dd><p>any serializable BSON value</p>
+</dd>
+<dt><a href="#Long">Long</a> : <code>BSON.Long</code></dt>
+<dd><p>A 64 bit integer, represented by the js-bson Long type.</p>
+</dd>
 <dt><a href="#KMSProviders">KMSProviders</a> : <code>object</code></dt>
 <dd><p>Configuration options that are used by specific KMS providers during key generation, encryption, and decryption.</p>
 </dd>
@@ -100,6 +116,13 @@ query for the data key itself against the key vault namespace.</p>
 <dd></dd>
 <dt><a href="#ClientEncryptionEncryptCallback">ClientEncryptionEncryptCallback</a> : <code>function</code></dt>
 <dd></dd>
+<dt><a href="#RangeOptions">RangeOptions</a> : <code>object</code></dt>
+<dd><p>min, max, sparsity, and range must match the values set in the encryptedFields of the destination collection.
+For double and decimal128, min/max/precision must all be set, or all be unset.</p>
+</dd>
+<dt><a href="#EncryptOptions">EncryptOptions</a> : <code>object</code></dt>
+<dd><p>Options to provide when encrypting data.</p>
+</dd>
 </dl>
 
 <a name="AutoEncrypter"></a>
@@ -265,9 +288,13 @@ The public interface for explicit in-use encryption
 
         * [.encrypt(value, options, [callback])](#ClientEncryption+encrypt)
 
+        * [.encryptExpression(expression, options)](#ClientEncryption+encryptExpression)
+
         * [.decrypt(value, callback)](#ClientEncryption+decrypt)
 
         * [.askForKMSCredentials()](#ClientEncryption+askForKMSCredentials)
+
+        * [._encrypt(value, expressionMode, options)](#ClientEncryption+_encrypt)
 
     * _inner_
         * [~decryptCallback](#ClientEncryption..decryptCallback)
@@ -534,10 +561,7 @@ if (!oldKey) {
 | Param | Type | Description |
 | --- | --- | --- |
 | value | <code>\*</code> | The value that you wish to serialize. Must be of a type that can be serialized into BSON |
-| options | <code>object</code> |  |
-| [options.keyId] | [<code>ClientEncryptionDataKeyId</code>](#ClientEncryptionDataKeyId) | The id of the Binary dataKey to use for encryption |
-| [options.keyAltName] | <code>string</code> | A unique string name corresponding to an already existing dataKey. |
-| [options.algorithm] |  | The algorithm to use for encryption. Must be either `'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'`, `'AEAD_AES_256_CBC_HMAC_SHA_512-Random'`, `'Indexed'` or `'Unindexed'` |
+| options | [<code>EncryptOptions</code>](#EncryptOptions) |  |
 | [callback] | [<code>ClientEncryptionEncryptCallback</code>](#ClientEncryptionEncryptCallback) | Optional callback to invoke when value is encrypted |
 
 Explicitly encrypt a provided value. Note that either `options.keyId` or `options.keyAltName` must
@@ -572,6 +596,21 @@ async function encryptMyData(value) {
   return clientEncryption.encrypt(value, { keyAltName: 'mySpecialKey', algorithm: 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic' });
 }
 ```
+<a name="ClientEncryption+encryptExpression"></a>
+
+### *clientEncryption*.encryptExpression(expression, options)
+**Experimental**: The Range algorithm is experimental only. It is not intended for public use. It is subject to breaking changes.The aggregation or match expression you wish to encrypt.  The value must be in the form  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| expression | <code>object</code> | a BSON document of one of the following forms:  1. A Match Expression of this form:      `{$and: [{<field>: {$gt: <value1>}}, {<field>: {$lt: <value2> }}]}`  2. An Aggregate Expression of this form:      `{$and: [{$gt: [<fieldpath>, <value1>]}, {$lt: [<fieldpath>, <value2>]}]}`    `$gt` may also be `$gte`. `$lt` may also be `$lte`. |
+| options | [<code>EncryptOptions</code>](#EncryptOptions) |  |
+
+Encrypts a Match Expression or Aggregate Expression to query a range index.
+
+Only supported when queryType is "rangePreview" and algorithm is "RangePreview".
+
+**Returns**: <code>Promise.&lt;object&gt;</code> - Returns a Promise that either resolves with the encrypted value or rejects with an error.  
 <a name="ClientEncryption+decrypt"></a>
 
 ### *clientEncryption*.decrypt(value, callback)
@@ -607,6 +646,23 @@ This returns anything that looks like the kmsProviders original input
 option. It can be empty, and any provider specified here will override
 the original ones.
 
+<a name="ClientEncryption+_encrypt"></a>
+
+### *clientEncryption*._encrypt(value, expressionMode, options)
+**Internal**:   
+
+| Param | Type | Description |
+| --- | --- | --- |
+| value | <code>\*</code> | The value that you wish to encrypt. Must be of a type that can be serialized into BSON |
+| expressionMode | <code>boolean</code> | a boolean that indicates whether or not to encrypt the value as an expression |
+| options | [<code>EncryptOptions</code>](#EncryptOptions) |  |
+
+A helper that perform explicit encryption of values and expressions.
+Explicitly encrypt a provided value. Note that either `options.keyId` or `options.keyAltName` must
+be specified. Specifying both `options.keyId` and `options.keyAltName` is considered an error.
+
+**Returns**: the raw result of the call to stateMachine.execute().  When expressionMode is set to true, the return
+         value will be a bson document.  When false, the value will be a BSON Binary.  
 <a name="ClientEncryption..decryptCallback"></a>
 
 ### *ClientEncryption*~decryptCallback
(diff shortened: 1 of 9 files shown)


bindings/node/src/mongocrypt.cc
```diff
@@ -528,64 +528,89 @@ Value MongoCrypt::MakeExplicitEncryptionContext(const CallbackInfo& info) {
         throw TypeError::New(Env(), "Parameter `value` must be a Buffer");
     }
```
baileympearson, line 530: I recommend hiding whitespace changes for this diff

bindings/node/src/mongocrypt.cc
```diff
@@ -528,64 +528,89 @@ Value MongoCrypt::MakeExplicitEncryptionContext(const CallbackInfo& info) {
...
+            throw TypeError::New(Env(), errorStringFromStatus(context.get()));
         }
     }
 
     std::unique_ptr<mongocrypt_binary_t, MongoCryptBinaryDeleter> binaryValue(BufferToBinary(valueBuffer.As<Uint8Array>()));
-    if (!mongocrypt_ctx_explicit_encrypt_init(context.get(), binaryValue.get())) {
+
+    const bool isExpressionMode = options.Has("expressionMode") && options.Get("expressionMode").IsBoolean()
+                                ? options.Get("expressionMode")
+                                : false;
```
addaleax, line 607: This should also work and might be a bit more natural:

```suggestion
    const bool isExpressionMode = options.Get("expressionMode").ToBoolean();
```

(corresponding to `!!options.expressionMode` in JS terms)

bindings/node/src/mongocrypt.cc
```diff
@@ -528,64 +528,89 @@ Value MongoCrypt::MakeExplicitEncryptionContext(const CallbackInfo& info) {
         throw TypeError::New(Env(), "Parameter `value` must be a Buffer");
     }
 
-    if (info.Length() > 1) {
-        Object options = info[1].ToObject();
+    Object options = info.Length() > 1 ? info[1].ToObject() : Object();
 
-        if (options.Has("keyId")) {
-            Napi::Value keyId = options["keyId"];
+    if (options.Has("keyId")) {
```
addaleax, line 533: Does this _work_ when `options` is `Object()`/`nullptr`? I would be surprised if it didn’t throw tbh

bindings/node/src/mongocrypt.cc
baileympearson: the documentation I read says that Object::get() returns undefined if the key is not present. But I'm now realizing that probably means a Value containing an undefined 🤦 

I'll try this out, thanks

bindings/node/src/mongocrypt.cc
baileympearson: You're wondering specifically if `Object.has(<key>)` throws on a default object?  I assumed it did not from the documentation:

https://github.com/nodejs/node-addon-api/blob/main/doc/object.md#empty-constructor
https://github.com/nodejs/node-addon-api/blob/main/doc/object.md#has (returns false if not present)

However, I may have mis-interpreted what an "empty object" means, I'll test it and see

addaleax: Yeah, I think it would – “empty” means “null pointer” here, and the Node-API functions should return/throw an error in that case. (`Napi::Object::New()` is the function that creates an empty object in the JS sense, i.e. `{ }`)

baileympearson: You're correct - this throws an exception:
```c++
Napi::Boolean Method(const Napi::CallbackInfo& info) {
  const Object obj = Object();
  const bool isPresent = obj.Has("some key");
  Napi::Env env = info.Env();
  return Napi::Boolean::New(env, isPresent);
}
```

bindings/node/README.md
```diff
@@ -56,9 +56,25 @@ npm test
 </dd>
 </dl>
 
+## Functions
+
+<dl>
+<dt><a href="#loadCredentials">loadCredentials(kmsProviders)</a> ⇒ <code>Promise</code></dt>
+<dd><p>Load cloud provider credentials for the user provided kms providers.
```
addaleax, line 63:
```suggestion
<dd><p>Load cloud provider credentials for the user provided KMS providers.
```

bindings/node/README.md
```diff
@@ -607,6 +646,23 @@ This returns anything that looks like the kmsProviders original input
 option. It can be empty, and any provider specified here will override
 the original ones.
 
+<a name="ClientEncryption+_encrypt"></a>
+
+### *clientEncryption*._encrypt(value, expressionMode, options)
+**Internal**:
```
addaleax, line 652: Isn’t the README for documenting public methods?

bindings/node/README.md
```diff
@@ -621,6 +677,29 @@ the original ones.
 ## MongoCryptError
 An error indicating that something went wrong specifically with MongoDB Client Encryption
 
+<a name="loadCredentials"></a>
+
+## loadCredentials(kmsProviders)
```
addaleax, line 682: Similarly here, this should probably not be included in the README. If it is, it should also be tagged as internal, like above

bindings/node/lib/clientEncryption.js
```diff
@@ -589,44 +610,29 @@ module.exports = function (modules) {
...
 
-          cb(null, result.v);
-        });
-      });
+    /**
+     * Encrypts a Match Expression or Aggregate Expression to query a range index.
+     *
+     * Only supported when queryType is "rangePreview" and algorithm is "RangePreview".
+     *
+     * @experimental The Range algorithm is experimental only. It is not intended for public use. It is subject to breaking changes.The aggregation or match expression you wish to encrypt.  The value must be in the form
```
addaleax, line 621:
```suggestion
     * @experimental The Range algorithm is experimental only. It is not intended for public use. It is subject to breaking changes. The value must be in the form
```

Or should this really be part of the docs here?

bindings/node/lib/clientEncryption.js
```diff
@@ -693,6 +699,61 @@ module.exports = function (modules) {
...
+          throw new TypeError(
+            `"options.keyAltName" must be of type string, but was of type ${keyAltNameType}`
+          );
+        }
+
+        contextOptions.keyAltName = bson.serialize({ keyAltName });
+      }
+
+      const rangePreviewRegex = /^rangepreview$/i;
+      if (typeof options.algorithm === 'string' && rangePreviewRegex.test(options.algorithm)) {
```
addaleax, line 739: Absolutely non-blocking, but I don’t think you really need a regexp here:

```suggestion
      if (
        typeof options.algorithm === 'string' && 
        options.algorithm.toLowerCase() === 'rangepreview'
      ) {
```

bindings/node/src/mongocrypt.cc
```diff
@@ -528,64 +528,86 @@ Value MongoCrypt::MakeExplicitEncryptionContext(const CallbackInfo& info) {
...
         }
 
-        if (options.Has("contentionFactor")) {
-            Napi::Value contention_factor_value = options["contentionFactor"];
-            int64_t contention_factor = contention_factor_value.IsBigInt() ?
-                contention_factor_value.As<BigInt>().Int64Value(nullptr) :
-                contention_factor_value.ToNumber().Int64Value();
-            if (!mongocrypt_ctx_setopt_contention_factor(context.get(), contention_factor)) {
-                throw TypeError::New(Env(), errorStringFromStatus(context.get()));
+        if (algorithm == "rangepreview") {
```
addaleax, line 567: If you want to avoid the extra lower-case check on the JS side (and that might be a good idea, considering that now this logic is split into two separate locations), you can also use

```suggestion
        if (strcasecmp(algorithm.c_str(), "rangepreview") == 0) {
```

here for case-insentive comparison

bindings/node/lib/clientEncryption.js
baileympearson: I'll remove this in favor of checking in C++

bindings/node/index.d.ts
```diff
@@ -493,6 +508,26 @@ export class ClientEncryption {
...
+   *      `{$and: [{<field>: {$gt: <value1>}}, {<field>: {$lt: <value2> }}]}`
+   *  2. An Aggregate Expression of this form:
+   *      `{$and: [{$gt: [<fieldpath>, <value1>]}, {$lt: [<fieldpath>, <value2>]}]}`
+   *
+   *    `$gt` may also be `$gte`. `$lt` may also be `$lte`.
+   */
+  encryptExpression(
+    value: Document,
+    options: ClientEncryptionOptions
+  ): Promise<Document>
```
addaleax, line 529: Tiny stylistic nit, feel free to ignore:

```suggestion
  ): Promise<Document>;
```

bindings/node/lib/common.js
```diff
@@ -46,6 +46,19 @@ class MongoCryptError extends Error {
...
 
+function maybeCallback(promiseFn, callback) {
+  const promise = promiseFn();
+  if (callback == null) {
+    return promise;
+  }
+
+  promise.then(
+    result => callback(undefined, result),
+    error => callback(error)
```
addaleax, line 57: Note that this will make exceptions from the callback result in an unhandled rejection rather than an uncaught exception; that’s a visible behavioral change. You may want to do something like this instead:

```suggestion
    result => process.nextTick(callback, undefined, result),
    error => process.nextTick(callback, error)
```

addaleax (approved):

bindings/node/test/clientEncryption.test.js
```diff
@@ -695,6 +692,142 @@ describe('ClientEncryption', function () {
...
+          sparsity: new BSON.Long(1)
+        },
+        keyId: dataKey
+      };
+    });
+
+    afterEach(() => teardown());
+
+    it('throws if rangeOptions is not provided', async function () {
+      delete completeOptions['rangeOptions'];
```
nbbeeken, line 783:
```suggestion
expect(delete completeOptions.rangeOptions).to.be.true;
```

Along with the other deletes it could be confusing if something prevents these from being removed down the road

bindings/node/test/clientEncryption.test.js
```diff
@@ -695,6 +692,142 @@ describe('ClientEncryption', function () {
...
+      const errorOrResult = await clientEncryption
+        .encryptExpression(expression, completeOptions)
+        .catch(e => e);
+
+      expect(errorOrResult).to.be.instanceof(Error);
```
nbbeeken, line 788: Same Q from the driver prose, is there anything more specific or at least a small portion of an error message we can match on here?

bindings/node/test/clientEncryption.test.js
```diff
@@ -695,6 +692,142 @@ describe('ClientEncryption', function () {
...
+    });
+
+    it(`throws if algorithm does not equal 'rangePreview'`, async function () {
+      completeOptions['algorithm'] = 'equality';
+      const errorOrResult = await clientEncryption
+        .encryptExpression(expression, completeOptions)
+        .catch(e => e);
+
+      expect(errorOrResult, errorOrResult.message).to.be.instanceof(Error);
```
nbbeeken, line 806: wanna include the message in the other assertions as well? I assume this one is different b/c of debugging

bindings/node/test/clientEncryption.test.js
```diff
@@ -695,6 +692,142 @@ describe('ClientEncryption', function () {
...
+    let completeOptions;
+    let dataKey;
+    const expression = {
+      $and: [{ someField: { $gt: 1 } }]
+    };
+
+    beforeEach(async function () {
+      if (requirements.SKIP_LIVE_TESTS) {
+        this.currentTest.skipReason = `requirements.SKIP_LIVE_TESTS=${requirements.SKIP_LIVE_TESTS}`;
+        this.test.skip();
```
nbbeeken, line 754: above we used this.currentTest.skip, and here it's this.test.skip, I guess they both work, any insight?

nbbeeken (changes requested):

bindings/node/test/clientEncryption.test.js
baileympearson: I copied this code block from above too...lol

bindings/node/lib/common.js
baileympearson: Neal and I looked at this for a while and, while we believe you're right, we actually couldn't reproduce the issue.

```javascript
async function wait() {
  await setTimeout(1000);
}
maybeCallback(() => wait(), () => { throw new Error("crashed process") });
```

Doesn't result in an UnhandledPromiseRejection

bindings/node/test/clientEncryption.test.js
nbbeeken: go figure, just curious

baileympearson:
I linted our d.ts file, which might make the diff hard to read.  It's in its own commit, so the changes to that file can either be viewed in other commits or I can revert that commit.

nbbeeken (approved):

bindings/node/lib/common.js
addaleax: Are you _sure_? Running this code (as `node test.js` with Node.js 16 through 19 at least) results in the unhandled rejection handler being called:

```js

function maybeCallback(promiseFn, callback) {
  const promise = promiseFn();
  if (callback == null) {
    return promise;
  }

  promise.then(
    result => callback(undefined, result),
    error => callback(error)
  );
  return;
}

async function wait() {
  await setTimeout(1000);
}

process.on('unhandledRejection', (...args) => console.log('unhandledRejection', args));

maybeCallback(() => wait(), () => { throw new Error("crashed process") });
```

nbbeeken: ah, we were looking for the unhandledRejection console log message, ty! We should make this improvement here, we'll want to file a follow up to replicate it elsewhere we've wrapped callbacks

baileympearson: Yup, you're right.  For some reason, the unhandled promise isn't being printed as an unhandled promise (we should have listened for unhandled promises).  

We filed a follow up ticket for the driver and legacy driver: https://jira.mongodb.org/browse/NODE-4996

nbbeeken (approved):

bindings/node/lib/common.js
addaleax: The default mode for unhandled rejections was changed to throwing in Node.js 15, which is why I think you weren’t seeing this printed as a warning: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V15.md#throw-on-unhandled-rejections---33021

PR #534, opened 2023-01-19
Reuse the proper DFP names from bid_conf.h

vector-of-bool:
This will ensure that we get the proper function references, regardless of where DFP comes from.


rcsanchez97 (approved):
LGTM

PR #535, opened 2023-01-20
MONGOCRYPT-518 add missing header for GCC 13

remicollet:
Using GCC 13 (Fedora 38) the build is failing

```
/builddir/build/BUILD/libmongocrypt-1.6.2/src/csfle-markup.cpp: In function 'int {anonymous}::do_main(int, const char* const*)':
/builddir/build/BUILD/libmongocrypt-1.6.2/src/csfle-markup.cpp:201:24: error: 'uint32_t' in namespace 'std' does not name a type; did you mean 'wint_t'?
  201 |       static_cast<std::uint32_t> (strlen (doc_ns)),
      |                        ^~~~~~~~
      |                        wint_t

```

See https://gcc.gnu.org/gcc-13/porting_to.html

### DIFF
--- a/src/csfle-markup.cpp
+++ b/src/csfle-markup.cpp
@@ -12,6 +12,7 @@ extern "C" {
 #include <iostream>
 #include <fstream>
 #include <string>
+#include <cstdint>
 
 #include <mongo_crypt-v1.h>
 


remicollet:
FYI: Fedora 38 build of version 1.6.2

* without the fix: https://koji.fedoraproject.org/koji/taskinfo?taskID=96366766
* with the fix: https://koji.fedoraproject.org/koji/taskinfo?taskID=96408802

This PR targets 1.6 but know know what is the preferred workflow, and if a 1.6.x bugfix is planed before 1.7

addaleax (approved):

remicollet:
> Is it OK to wait?

I don't need a 1.6.3, I was able to build the Fedora RPM using 1.6.2 + this patch, so everything OK on my side
(you will see if other report exists from other users)

BTW, I haven't check if the 1.7 branch is OK (other files can be affected, e.g. src/mlib/int128.test.cpp, src/mc-dec128.test.cpp)

kevinAlbs:
> I don't need a 1.6.3, I was able to build the Fedora RPM using 1.6.2 + this patch, so everything OK on my side
(you will see if other report exists from other users)

OK. I will merge to master to be released in 1.7.0 since 1.7.0 is planned soon.

> BTW, I haven't check if the 1.7 branch is OK (other files can be affected, e.g. src/mlib/int128.test.cpp, src/mc-dec128.test.cpp)

Good point. 1.7 has not been branched yet. I reproduced the issue in Fedora 38 in docker on the master branch. After applying this patch on the master branch, there is no error.

PR #537, opened 2023-01-20
fix(NODE-4987): node csfle uses promises instead of callbacks

durran:
Node driver v5 removes callback support so we needed to update `createDataKey` and `markCommand` to use the promise API.

This included searches through the codebase for all CRUD methods - all other usage already uses async await or promises.

Patch build on the Node driver pointing at this commit: https://spruce.mongodb.com/version/63cac4515623431af90fd308/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -230,7 +230,14 @@ module.exports = function (modules) {
      */
     teardown(force, callback) {
       if (this._mongocryptdClient) {
-        this._mongocryptdClient.close(force).then(callback);
+        this._mongocryptdClient.close(force).then(
+          result => {
+            return callback(null, result);
+          },
+          error => {
+            callback(error);
+          }
+        );
       } else {
         callback();
       }
--- a/bindings/node/lib/autoEncrypter.js
+++ b/bindings/node/lib/autoEncrypter.js
@@ -202,10 +202,26 @@ module.exports = function (modules) {
       };
 
       if (this._mongocryptdManager.bypassSpawn) {
-        return this._mongocryptdClient.connect(_callback);
+        return this._mongocryptdClient.connect().then(
+          result => {
+            return _callback(null, result);
+          },
+          error => {
+            _callback(error, null);
+          }
+        );
       }
 
-      this._mongocryptdManager.spawn(() => this._mongocryptdClient.connect(_callback));
+      this._mongocryptdManager.spawn(() => {
+        this._mongocryptdClient.connect().then(
+          result => {
+            return _callback(null, result);
+          },
+          error => {
+            _callback(error, null);
+          }
+        );
+      });
     }
 
     /**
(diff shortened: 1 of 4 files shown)


addaleax (approved):

bindings/node/lib/stateMachine.js
```diff
@@ -437,13 +437,15 @@ module.exports = function (modules) {
       const dbName = databaseNamespace(ns);
       const rawCommand = bson.deserialize(command, options);
 
-      client.db(dbName).command(rawCommand, options, (err, response) => {
-        if (err) {
+      client
+        .db(dbName)
+        .command(rawCommand, options)
+        .then(response => {
```
nbbeeken, line 443:
```typescript
client
        .db(dbName)
        .command(rawCommand, options)
        .then(
          response => {
            return callback(null, bson.serialize(response, this.options));
          },
          error => {
            callback(error, null);
          }
        );
```
We should use the two arguments of `then`, otherwise if the callback throws here we'll call it twice

bindings/node/lib/clientEncryption.js
```diff
@@ -262,13 +262,12 @@ module.exports = function (modules) {
...
             .db(dbName)
             .collection(collectionName)
-            .insertOne(dataKey, { writeConcern: { w: 'majority' } }, (err, result) => {
-              if (err) {
-                cb(err, null);
-                return;
-              }
-
+            .insertOne(dataKey, { writeConcern: { w: 'majority' } })
+            .then(result => {
```
nbbeeken, line 266: two arg of then format, same as other comment

nbbeeken (changes requested):
Don't we also need to change:
- `this._mongocryptdClient.connect` in autoEncrypter.js
- `this._mongocryptdClient.close` in autoEncrypter.js
- `fetchCollectionInfo` in stateMachine.js
- `fetchKeys` in stateMachine.js

bindings/node/lib/stateMachine.js
durran: Updated.

bindings/node/lib/clientEncryption.js
durran: Updated.

durran:
@nbbeeken The connect, close, and cursor usages have also been updated and new evg patch link in description.

bindings/node/lib/autoEncrypter.js
```diff
@@ -214,7 +230,7 @@ module.exports = function (modules) {
      */
     teardown(force, callback) {
       if (this._mongocryptdClient) {
-        this._mongocryptdClient.close(force, callback);
+        this._mongocryptdClient.close(force).then(callback);
```
nbbeeken, line 233: If close fails (unlikely) then the callback will never be invoked

nbbeeken (changes requested):

nbbeeken (approved):
LGTM, pending CI

PR #540, opened 2023-01-23
test(NODE-4897): bump release test timeout

durran:
Sometimes this test times out in the before hook running npm pack. Takes around ~600ms for me locally but sometimes on CI it takes over 2s. This moves the default timeout to 5s on that test.

https://spruce.mongodb.com/version/63d284791e2d177b176053f5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/bindings/node/test/release.test.js
+++ b/bindings/node/test/release.test.js
@@ -27,7 +27,7 @@ const REQUIRED_FILES = [
   'package/src/mongocrypt.h'
 ];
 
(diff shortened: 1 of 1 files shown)


baileympearson (changes requested):
Do we have a patch on the driver showing that the driver socks5 csfle tests pass with this change on Node19?

baileympearson (approved):

PR #541, opened 2023-01-23
Add a Basic Earthly File

vector-of-bool:
I've been using a variation of this for local dev on most projects for a long time. Unlike the prior mongo-c-driver PR with the same purpose, this one is far simpler and trimmed down.

This does not modify CI or introduce any new EVG tasks. Perhaps in the future? For now, it's just useful for local dev. It only defines a target that runs `build_all.sh`, and does not export any artifacts. Adding new tasks and artifacts may be a task for the future.

### DIFF
--- a/.earthlyignore
+++ b/.earthlyignore
@@ -1,0 +1,3 @@
+camke-build/
+camke-build-deb/
+_build/
--- a/.earthlyignore
+++ b/.earthlyignore
@@ -1,3 +1,3 @@
-camke-build/
-camke-build-deb/
+cmake-build/
+cmake-build-deb/
 _build/
--- a/.evergreen/ensure-cmake.sh
+++ b/.evergreen/ensure-cmake.sh
@@ -140,6 +140,9 @@ _ensure_cmake() {
 
     declare arch
     arch="$(uname -m)"
+    if test -f /etc/alpine-release; then
+        arch="$arch-musl"
+    fi
 
     # Otherwise we need to obtain it
     log "Obtaining CMake $CMAKE_VERSION for $OS_NAME-$arch"
--- a/Earthfile
+++ b/Earthfile
@@ -1,0 +1,257 @@
+# Earthly Intro: https://docs.earthly.dev/
+# Earthfile Reference: https://docs.earthly.dev/docs/earthfile
+
+# Quick notes:
+    #   • The shell script at ".evergreen/earthly.sh" can be used to automatically
+    #     download and use a fixed version of Earthly that is compatible with
+    #     this file version. Execute the shell script as if it were the Earthly
+    #     executable itself.
+    #   • In this file, the convention is to copy the source tree into /s/libmongocrypt
+    #   • Earthly copies the "build context" (working directory) into the local buildkit
+    #     daemon before each build execution. (Similar to the Docker build context).
+    #     If you have a large amount of data in the working directory, this could be
+    #     slow. The ".earthlyignore" file specifies patterns of files and directories
+    #     to exclude from the context upload. Modify it to suite your needs, if necessary.
+    #   • Only a subset of the source tree is COPY'd into the build environment. Files
+    #     outside of this set will not be available in the build. See the COPY_SOURCE
+    #     command for the list.
+    #   • Modification at any layer will invalidate caching on all subsequent build
+    #     layers. This is important and by-design in Earthly. Push infrequently-modified
+    #     operations to earlier in the process the pipeline to make better use of
+    #     the cache.
+    #
+    # This file has a few major sections:
+    #   - Setup COMMANDs
+    #   - Utility COMMANDs
+    #   - Environment targets
+    #   - Build/test/CI targets
+    #
+    # All environment targets begin with "env.". All build targets (should) accept an "env"
+    # parameter that specifies the name of the environment to use for the task. The name
+    # of an environment is specified following the "env." prefix. For example, the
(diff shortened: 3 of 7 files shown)


kkloberdanz:
> Is the goal to eventually migrate Evergreen scripts to being executed in Earthly/Docker environments to improve test environment isolation?

I've had good experiences running CI in a docker environment that can be easily replicated locally. It really helps to debug CI failures if you can spin up a near identical environment on your own machine and start setting break points and get a fast feedback loop when making changes.

I'm not familiar with Earthly, but it seems interesting.

vector-of-bool:
@eramongodb not necessarily to replace all CI tasks, but being able to run tasks _for_ arbitrary platforms on _any_ arbitrary platform is extremely beneficial in terms of build runner utilization and reproducibility. i.e. I don't need to wait in queue for a rhel6.7 runner to spawn when I can just run it on the same machine that I just used for deb10, etc.

And indeed, a significant benefit is being able to replicate an environment locally, and make modifications to that environment without depending on an external team to schedule, approve, and apply changes to a system (which could also have knock-on effects on other teams).

kkloberdanz (approved):
I think that Docker is a great tool for reproducing specific environments locally, and I think this will help with debugging CI failures. I haven't worked with Earthly before, but I'm up for us  giving it a try.

eramongodb (approved):
Confirmed behavior on Ubuntu 20.04 in WSL (required installing [Docker Desktop with WSL2 Backend for Windows](https://docs.docker.com/desktop/windows/wsl/) per [installation instructions](https://earthly.dev/get-earthly)) and ran several builds with different environments. Very interesting tool. Looking forward to its further use.

.earthlyignore
```diff
@@ -0,0 +1,3 @@
+camke-build/
+camke-build-deb/
```
kevinAlbs, line 2:
```suggestion
cmake-build/
cmake-build-deb/
```

.evergreen/ensure-cmake.sh
```diff
@@ -140,6 +140,9 @@ _ensure_cmake() {
 
     declare arch
     arch="$(uname -m)"
+    if test -f /etc/alpine-release; then
+        arch="$arch-musl"
```
kevinAlbs, line 144: Is appending `-musl` for the benefit of logging? It does not appear to impact the cmake or ninja downloads.

Earthfile
```diff
@@ -0,0 +1,257 @@
...
+    COPY --dir \
+        .git/ \
+        cmake/ \
+        kms-message/ \
+        test/ \
+        debian/ \
+        src/ \
+        etc/ \
+        LICENSE \
+        VERSION_CURRENT \
```
kevinAlbs, line 232:
```suggestion
```

VERSION_CURRENT is generated by cmake. Running `earthly +build --env=u22` may result in `Error: File not found "/VERSION_CURRENT"` if libmongocrypt was not previously configured in the build context.

Earthfile
```diff
@@ -0,0 +1,257 @@
...
+    COPY etc/install-package.sh /usr/local/bin/__install
+    RUN chmod +x /usr/local/bin/__install
+    ENV USER_CACHES_DIR=/Cache
+
+# Environment setup commands below. Each provides the basic environment for a
+# libmongocrypt build. Additional packages and setup may be required for
+# individual tasks.
+
+DEBIAN_SETUP:
+    # Setup for a debian-like build environment. Used for both Debain and Ubuntu
```
kevinAlbs, line 82:
```suggestion
    # Setup for a debian-like build environment. Used for both Debian and Ubuntu
```

Earthfile
```diff
@@ -0,0 +1,257 @@
...
+    COMMAND
+    RUN __install build-essential g++ libssl-dev curl unzip python3 \
+                  libbson-dev pkg-config git ccache findutils ca-certificates
```
kevinAlbs, line 85: Is `libbson-dev` needed?

Earthfile
```diff
@@ -0,0 +1,257 @@
...
+    #   • rl8 - RockyLinux 8 - Stand-in for RHEL 8
+    #   • c7 - CentOS 7 - Stand-in for RHEL 7
+    #   • c6 - CentOS 6 - Stand-in for RHEL 6
+    #   • amzn1 - AmazonLinux (2018.03)
+    #   • amzn2 - AmazonLinux 2
+    #   • deb9 - Debian 9.2
+    #   • deb10 - Debian 10.0
+    #   • deb11 - Debian 11.0
+    #   • sles15 - OpenSUSE Leap 15.0
+    #   • sles15 - OpenSUSE Leap 15.0
```
kevinAlbs, line 50:
```suggestion
```

kevinAlbs (approved):
LGTM. Earthly seems really helpful for reproducing build failures quickly.

This may help adding support for platforms unavailable on Evergreen. Alpine is requested in MONGOCRYPT-416.

.evergreen/ensure-cmake.sh
vector-of-bool: It causes the `case` branch to fall through to `_build_cmake`/`_build_ninja`, which is required since the prebuilt downloads do not work with muslc.

Earthfile
vector-of-bool: Nope, since we download our own.

Earthfile
vector-of-bool: Good point. Although this does present an issue on e.g. CentOS 6, where the system Git and Python cannot generate the `VERSION_CURRENT` file correctly. A problem for another time, perhaps.

PR #546, opened 2023-01-24
fix(NODE-5000): make aws credential providers peer optional

nbbeeken:
#### What is changing?

aws credentials provider is now an optional peer dependency mirroring the driver (post v5) in order to remove the default installation of the optional package.

AWS credential provider should only be installed when needed.

- [ ] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat([NODE-1234](https://jira.mongodb.org/browse/NODE-1234))!: rewriting everything in coffeescript`


nbbeeken:
Driver patch: https://spruce.mongodb.com/version/63d15b7c3e8e860b62274c5c/tasks

bindings/node/etc/build-static.sh
```diff
@@ -34,6 +33,4 @@ run_cmake --build . --target install --config RelWithDebInfo
 popd #./
 
 # build the `mongodb-client-encryption` addon
-# note the --unsafe-perm parameter to make the build work
```
baileympearson, line 37: Why did you remove `--unsafe-perm`?  Is it no longer relevant?

bindings/node/test/tools/utils.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = Object.create(null);
```
baileympearson, line 3: We already have a file that builds a "requirements" object for tests to use (`requirements.helpers.js`).  Can we move this code there?

Also, following the pattern set in that file, it might be nice to simply export a boolean indicating whether or not the credential provider is installed (considering we'll be adding support for gcp / azure soon as well).

Maybe exporting something like this:
```javascript
{
  credentialProvidersInstalled: { 
    aws: <boolean>
  }
}
```

baileympearson (changes requested):

bindings/node/etc/build-static.sh
nbbeeken: Yep, the PR linked is closed, we also don't run as a root user in CI

baileympearson (approved):

PR #549, opened 2023-01-26
Java publishing

rozza:
Evergreen is now responsible for publishing Java artifacts.

- If the version string ends with SNAPSHOT then publishes to the Snapshots repo.
  Note: Uses the JNA libs from the current build.

- If is a release then publishes the release to maven central staging.
  A release is when the current git tag is prefixed with java (eg: java-1.7.0) AND
  the git tag version matches the version the build.gradle.kts.
  Note: Uses the JNA libs from the associated tag.
        Eg: Tag java-1.7.0 will use the JNA libs created by the 1.7.0 release tag.

MONGOCRYPT-491


bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -254,41 +259,47 @@ tasks.javadoc {
...
+        |
+        | - If is a release then publishes the release to maven central staging.
+        |   A release is when the current git tag is prefixed with java (eg: java-1.7.0)
+        |   AND the git tag version matches the version the build.gradle.kts.
+        |   Note: Uses the JNA libs from the associated tag.
+        |   Eg: Tag java-1.7.0 will use the JNA libs created by the 1.7.0 release tag.
+        |
+        | To override the JNA library downloaded use -DgitRevision=<git hash>
+    """.trimMargin()
+    val isRelease = version.toString().endsWith("-SNAPSHOT") || (isJavaTag && gitVersion == version)
```
kevinAlbs, line 277: If version ends with `-SNAPSHOT` (e.g. `1.7.0-SNAPSHOT`) and mismatches tag (e.g. `java-1.7.1`) there is no error.

Consider expanding this check, and renaming to `versionOK`:

```kotlin
    val isSnapshot = version.toString().endsWith("-SNAPSHOT")
    var versionOK = false
    if (isSnapshot && !isJavaTag) {
        // This is a snapshot commit, and the commit is untagged.
        versionOK = true
    } else if (!isSnapshot && isJavaTag && gitVersion == version) {
        // This is a release commit, and the tag matches.
        versionOK = true
    }
```

kevinAlbs:
After merging this may need to be cherry-picked to the r1.7 branch. The r1.7 branch contains the commit tagged for 1.7.0 and will contain future 1.7 patch releases. The r1.7 branch is tracked in the [libmongocrypt-release Evergreen project](https://evergreen.mongodb.com/waterfall/libmongocrypt-release).

.evergreen/config.yml
```diff
@@ -1339,13 +1340,18 @@ buildvariants:
...
   run_on: ubuntu1804-test
   tasks:
     - name: "upload-java"
-    - name: "publish-snapshot"
     - name: "upload-all"
+- name: java-release
+  display_name: Java Release
+  run_on: ubuntu1804-test
+  tasks:
+    - "publish-java-snapshots"
```
jyemin, line 1353: Why do we need to publish the snapshot when we publish a release? I would think that the waterfall would have done that already.

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -43,7 +43,7 @@ repositories {
 }
 
 group = "org.mongodb"
-version = "1.8.0-SNAPSHOT"
+version = "1.7.0-SNAPSHOT"
```
jyemin, line 46: Should this stay as 1.8 on the `master` branch and only switch to 1.7 when this is cherry-picked to the 1.7 branch?

bindings/java/mongocrypt/.evergreen/publish.sh
```diff
@@ -21,5 +21,7 @@ export ORG_GRADLE_PROJECT_signing_secretKeyRingFile=${PROJECT_DIRECTORY}/secring
 echo "Publishing snapshot with jdk8"
 export JAVA_HOME="/opt/java/jdk8"
 
+SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true -Dorg.gradle.internal.http.connectionTimeout=120000 -Dorg.gradle.internal.http.socketTimeout=120000"
```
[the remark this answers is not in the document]
rozza: Yes, its added because we sign the packages using md5 and sha1.

.evergreen/config.yml
rozza: There are two tasks: 

- publish-java-snapshots
  which happens on non tagged pushes: `allow_for_git_tag: false`
  depends on upload_java - the latest java build
- publish-java-release
  Only happens on tag releases: `git_tag_only: true`
  Doesn't depend on upload_java - because it will use the canonical tagged assets

They both call the same function underneath so could be removed. But it does indicate in the evergreen build what task is run.

Its the gradle file that determines the s3 location for the assets to be packaged with the release - regardless of it being a snapshot or tagged release.

bindings/java/mongocrypt/build.gradle.kts
rozza: Ack

bindings/java/mongocrypt/build.gradle.kts
rozza: > If version ends with -SNAPSHOT (e.g. 1.7.0-SNAPSHOT) and mismatches tag (e.g. java-1.7.1) there is no error.

Like that now made it an error in that scenario.

I've kept it as not an error if its not a java tag though. As this task is run on all tags.

.evergreen/config.yml
rozza: As they both call the same underlying function - there probably isnt a benefit for the added complexity.  Updated to use a single task.

jyemin (approved):
LGTM

bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -66,16 +66,20 @@ dependencies {
 /*
  * Git version information
  */
-val gitVersion: String by lazy {
+
+// Returns a Pair<Boolean, String> representing if the current tag is a `java-<version>` tag and the version
```
kevinAlbs, line 70:
```suggestion
// Returns a String representing the output of `git describe`
```

kevinAlbs (approved):

PR #550, opened 2023-01-26
MONGOCRYPT-517 vendor Intel DFP code and also allow using system DFP

rcsanchez97:
Full Evergreen patch build: https://evergreen.mongodb.com/version/63d2a6e31e2d1703aede1fd8?redirect_spruce_users=true


CMakeLists.txt
```diff
@@ -602,6 +609,10 @@ if (ENABLE_BUILD_FOR_PPA)
    endif ()
    set (PKG_CONFIG_LIBS "${PKG_CONFIG_LIBS} -pthread")
 endif ()
+if (MONGOCRYPT_DFP_DIR STREQUAL "USE-SYSTEM")
+   get_property (SYSTEM_DFP_LOC TARGET intel_dfp PROPERTY IMPORTED_LOCATION)
+   set (PKG_CONFIG_LIBS "${PKG_CONFIG_LIBS} ${SYSTEM_DFP_LOC}")
```
vector-of-bool, line 614: Is there not a pkg-config for DFP that one can depend on transitively?

vector-of-bool (approved):
LGTM. One comment.

CMakeLists.txt
rcsanchez97: Sadly, there is not one.  This is basically "here's a box of Lego, see what you can make from it"

kevinAlbs (approved):

PR #551, opened 2023-01-26
[MONGOCRYPT-521] LTO Build Fixes

vector-of-bool:
Being too clever in `mc-dec128.h` is caught by the LTO phase as ABI mismatches.

This PR also adds an `LTO.cmake` module which can be used to build and test with LTO enabled.


cmake/LTO.cmake
```diff
@@ -0,0 +1,57 @@
+
+set (MONGO_LTO "OFF"
```
rcsanchez97, line 2: I feel like the variable should actually be named `MONGOCRYPT_LTO`.  This brings it more into line with `MONGOCRYPT_CRYPTO`, which seems to establish the pattern of using `MONGOCRYPT_` as the prefix.  Additionally, the name `MONGO_LTO` (at least to me) implies the server project or a more generic standard variable that would apply to multiple MongoDB projects.

remicollet:
While this seems to fix LTO build, build is failing again with only shared libson  (static not built, per RPM guidelines)

```
gmake[2]: Entering directory '/dev/shm/extras/BUILD/libmongocrypt-1.7.0/redhat-linux-build'
gmake[2]: *** No rule to make target '_MONGOCRYPT_SYSTEM_LIBBSON_STATIC-NOTFOUND', needed by 'mc-dec128.test'.  Stop.

```

So again https://jira.mongodb.org/browse/MONGOCRYPT-451

cmake/LTO.cmake
```diff
@@ -0,0 +1,57 @@
...
+    return ()
+endif ()
+
+# We need try_compile(), because we need more than one source file to accurately
+# check for LTO support
+try_compile (
+    MONGO_HAVE_LTO_${MONGO_LTO}
+    "${CMAKE_CURRENT_BINARY_DIR}/_mongo-lto-check/${MONGO_LTO}"
+    SOURCES "${CMAKE_CURRENT_LIST_DIR}/ltocheck-lib.c"
+            "${CMAKE_CURRENT_LIST_DIR}/ltocheck-main.c"
```
kevinAlbs, line 45: Do the source files `ltocheck-lib.c` and `ltocheck-main.c` need to be included?

kevinAlbs (approved):
LGTM with one comment.

cmake/LTO.cmake
vector-of-bool: Oh yeah, those would be important :)

cmake/LTO.cmake
vector-of-bool: This is actually chosen deliberately to match what is used in `Sanitizers.cmake` etc (currently in mongo-c-driver, but I plan on bringing that one over soon as well)

PR #552, opened 2023-01-26
feat(NODE-5013): automatically create Queryable Encryption keys

nbbeeken:
### Description

https://github.com/mongodb/node-mongodb-native/pull/3544

#### What is changing?

Adds a new API to `ClientEncryption` called `createEncryptedCollection`.
- Searches the `fields` array of the `encryptedFields` `CreateCollectionOptions` option
- Any `keyId`s that are null will be generated by calling `createDataKey`
- User input is not modified 

It is convenient to combine creating dataKeys and the collection that will store the encrypted fields in one step.

- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`


bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +829,115 @@ describe('ClientEncryption', function () {
...
+
+    afterEach(async () => {
+      sinon.restore();
+      await db.dropDatabase().catch(() => null);
+    });
+
+    it('throws if options are omitted', async () => {
+      const error = await clientEncryption
+        .createEncryptedCollection(db, collectionName)
+        .catch(error => error);
```
baileympearson, line 858: We're missing a test for the case when `encryptedFieldsMap` is not an array (we guard against it, we should test it).  Also, should we throw in that case instead of just not entering the if-statement?

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +829,115 @@ describe('ClientEncryption', function () {
...
+    });
+
+    it('throws if options.encryptedFields are omitted', async () => {
+      const error = await clientEncryption
+        .createEncryptedCollection(db, collectionName, {})
+        .catch(error => error);
+      expect(error).to.be.instanceOf(MongoCryptError);
+    });
+
+    it('throws MongoCryptCreateEncryptedCollectionError if createDataKey rejects', async () => {
```
baileympearson, line 869: The logic in the method is incorrect.  From the spec:
```
Let EF' be a copy of EF. Update EF' in the following manner:
Let Fields be the "fields" element within EF'.
If Fields is present and is an array value, then for each element F of Fields:
If F is not a document element, skip it.
Otherwise, if F has a "keyId" named element K and K is a null value:
Let D be the result of CE.createDataKey(kmsProvider, dkOpts).
If generating D resulted in an error E, the entire CreateEncryptedCollection must now fail with error E. Return the partially-formed EF' with the error so that the caller may know what datakeys have already been created by the helper.
```

So first we deep copy the whole EF object, then we update elements as we encounter them.  Consequently, if an error is encountered, we return the whole copied EF document (which includes any fields that have not yet been processed).  Your logic pushes new fields into the array, so when an error is encountered, only processed keys are returned in the returned `encryptedFieldsMap`.

So test cases we need to ensure we test here (I would have specific tests for each test case, perhaps all wrapped in a context block)

when there is an error thrown

- the root cause of the error is assigned to the mongo crypt error
- the encrypted fields map is present on the error
- the encrypted fields map is the same length as the original and contains documents in the same order
- the encrypted fields map only has data keys filled out for successful data key creations

bindings/node/lib/common.js
```diff
@@ -36,16 +36,22 @@ function collectionNamespace(ns) {
...
   }
 
   get name() {
     return 'MongoCryptError';
   }
 }
 
+class MongoCryptCreateEncryptedCollectionError extends MongoCryptError {
+  constructor(message, { encryptedFields, cause }) {
+    super(message, { cause });
```
baileympearson, line 50: We can't use cause, it's only supported in Node16+.

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +829,115 @@ describe('ClientEncryption', function () {
...
+
+      expect(result).to.have.property('namespace', 'createEncryptedCollectionDb.secure');
+      expect(encryptedFields, 'original encryptedFields should be unmodified').nested.property(
+        'fields[0].keyId',
+        false
+      );
+      expect(
+        result.options.encryptedFields,
+        'encryptedFields created by helper should have replaced nullish keyId'
+      ).nested.property('fields[1].keyId', keyId);
```
baileympearson, line 940: my two cents - mocking a return value and then asserting on properties of your own mocked method is circular.

i think instead we could spy on `db.createCollection` (and return enough just to satisfy the code path, like a Promise.resolve() or something) and then assert that `createCollection` was called with the expected parameters.

Or we could consider running these tests against a live server and not mocking the `Db` object at all.

baileympearson (changes requested):

bindings/node/test/clientEncryption.test.js
nbbeeken: `encryptedFields.fields`? The spec has only an if stmt for this, I think we shouldn't add more validation than is needed for the helper. Added a test for this, and for when an item is not a document

bindings/node/index.d.ts
```diff
@@ -27,6 +30,13 @@ export interface DataKey {
  */
 export class MongoCryptError extends Error {}
 
+/**
+ * An error indicating that `ClientEncryption.createEncryptedCollection()` failed to create a collection
+ */
+export class MongoCryptCreateEncryptedCollectionError extends MongoCryptError {
```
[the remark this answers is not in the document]
nbbeeken: documented

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +829,115 @@ describe('ClientEncryption', function () {
...
+      db = client.db('createEncryptedCollectionDb');
+      await db.dropDatabase().catch(() => null);
+    });
+
+    afterEach(async () => {
+      sinon.restore();
+      await db.dropDatabase().catch(() => null);
+    });
+
+    it('throws if options are omitted', async () => {
```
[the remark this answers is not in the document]
nbbeeken: > destructuring errors are opaque

It actually states the name of the missing property
```txt
> createEncryptedCollection(db, '')
TypeError: Cannot destructure property 'provider' of 'undefined' as it is undefined.
```

> throwing if it is not provided ... are we testing this only because we're destructing options from it? it does feel odd to test for the presence of only 1/3 required arguments

Yea so we generally don't try to call our functions with missing arguments unless there is overloads that permit it. For example, `insertOne` crashes with `Cannot read properties of undefined` if you don't pass it an object to insert. I added this test because it is a new way that we just let the runtime handle crashing, but I think its equivalent to that approach we've taken so far, but with a _slightly_ enhanced error message.

> missing a test for options.provider

this will be validated by `createDataKey` in the c++

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,75 @@ module.exports = function (modules) {
...
+     *
+     * @template {TSchema} - Schema for the collection being created
+     * @param {Db} db - A database to create the collection in
+     * @param {string} name - The name of the collection to be created
+     * @param {object} options - Options for createDataKey and for createCollection.
+     * @param {string} options.provider - provider name
+     * @param {ClientEncryptionCreateDataKeyProviderOptions} options.createDataKeyOptions - options to pass to createDataKey
+     * @param {CreateCollectionOptions} options.createCollectionOptions - options to pass to createCollection, must include `encryptedFields`
+     * @returns {Promise<{ collection: Collection<TSchema>, encryptedFields: Document }>} - created collection and generated encryptedFields
+     * @throws {MongoCryptCreateEncryptedCollectionError} - If part way through the process createDataKey fails, an error will be rejected that has the `encryptedFields` that were created.
```
addaleax, line 568: As I understand the code, it’s not just the `encrypedFields` that were created, it’s also including the fields that were already present before (potentially with a non-null `keyId`)?

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,75 @@ module.exports = function (modules) {
...
+        const createDataKeyPromises = [];
+        for (const [index, field] of encryptedFields.fields.entries()) {
+          if (typeof field !== 'object' || field.keyId != null) {
+            continue;
+          }
+
+          createDataKeyPromises.push(
+            (async () => {
+              const keyId = await this.createDataKey(provider, createDataKeyOptions);
+              return { nullKeyIdOffset: index, keyId };
```
addaleax, line 592: Fwiw, I think this might be a bit more readable if this was also just `index` rather than `nullKeyIdOffset` :)

bindings/node/index.d.ts
```diff
@@ -27,6 +32,14 @@ export interface DataKey {
...
+  /** @experimental The partial `encryptedFields` that was completed while attempting createEncryptedCollection */
+  encryptedFields: NonNullable<CreateCollectionOptions['encryptedFields']>;
```
addaleax, line 40: Also document `errors: Error[]`?

bindings/node/test/clientEncryption.test.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 const fs = require('fs');
-const expect = require('chai').expect;
+const chai = require('chai');
+chai.config.truncateThreshold = 0;
```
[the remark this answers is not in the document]
addaleax: TIL! Now I want this in the mongosh CI as well :)

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,75 @@ module.exports = function (modules) {
...
+          encryptedFields: { ...encryptedFields },
+          ...createCollectionOptions
+        }
+      } = options;
+
+      if (Array.isArray(encryptedFields.fields)) {
+        encryptedFields.fields = [...encryptedFields.fields];
+
+        const createDataKeyPromises = [];
+        for (const [index, field] of encryptedFields.fields.entries()) {
```
baileympearson, line 584: I don't think this logic is correct.  The algorithm outlined in the spec says

- for each entry in the encrypted fields map
- if the keyId is null, create it.  if there's an error, return early with the partially filled encryptedFields.

We can't parallelize this, it should be sequential

baileympearson (changes requested):

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +829,115 @@ describe('ClientEncryption', function () {
...
+      expect(error).to.be.instanceOf(MongoCryptCreateEncryptedCollectionError);
+      expect(error.cause).to.equal(customError);
+      expect(error.encryptedFields).property('fields').that.is.an('array');
+      expect(error.encryptedFields.fields).to.have.lengthOf(3);
+      expect(error.encryptedFields.fields).to.have.nested.property('[0].keyId', keyId);
+      expect(error.encryptedFields.fields).to.not.have.nested.property('[1].keyId');
+      expect(error.encryptedFields.fields).to.have.nested.property('[2].keyId', 'cool id!');
+    });
+
+    it('creates keyIds where they are nullish without editing input', async () => {
```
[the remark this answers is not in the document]
baileympearson: yup, sorry for confusion

bindings/node/lib/clientEncryption.js
addaleax: I had the same thought but came to the conclusion that this is probably okay – practically speaking, callers are still going to have to handle the same set of circumstances, right?

bindings/node/test/clientEncryption.test.js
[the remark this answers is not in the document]
baileympearson: A destructing error tells you what key was undefined, but not on what object it was accessed:

```
function bar({ name })  {
	console.log(name);
}

bar();

// TypeError: Cannot destructure property 'name' of 'undefined' as it is undefined.
```

An error that says something like "option `<optionName>` is required but was not provided` is much clearer and more actionable.

bindings/node/lib/clientEncryption.js
baileympearson: I guess that's true.

Personally, I see it as an unnecessary deviation from the spec.  It's not a high-traffic code path, so I don't see parallelization being important.  Also

- the code is more complicated
- users need to contend with multiple errors and mapping errors to objects

If we do decide to go forward with this approach - should we consider storing more actionable information on the error than just an array of errors?  The current implementation doesn't provide an easy mechanism for users to map which error came from which null keyId.

bindings/node/test/types/index.test-d.ts
```diff
@@ -0,0 +1,30 @@
+import { expectAssignable, expectError } from 'tsd';
+import { ClientEncryption } from '../..';
+
+type RequiredCreateEncryptedCollectionSettings = Parameters<
+  ClientEncryption['createEncryptedCollection']
+>[2];
+
+expectError<RequiredCreateEncryptedCollectionSettings>({});
```
baileympearson, line 8: do we have precedence for type tests like this?  It seems like all that's being tested is optionality of arguments.

bindings/node/lib/clientEncryption.js
nbbeeken: After hopping on a call with Bailey we've agreed that parallelism is okay, but we shouldn't try to have all this index tracking, it adding to the complexity of it all. I've updated things to use filter, map, and some where applicable, take another look.

bindings/node/lib/clientEncryption.js
nbbeeken: That's correct, based on these bits of the spec:

- If F (a field from fields) is not a document element, skip it.
- Otherwise, if F has a "keyId" named element K and K is a null value:
  - Let D be the result of CE.createDataKey(kmsProvider, dkOpts).
  - etc...

bindings/node/test/types/index.test-d.ts
nbbeeken: I made the typescript change the type of CreateCollectionOptions to be the same except `encryptedFields` is now required, so I wanted to demonstrate that in a test so it couldn't be changed without notice down the road

bindings/node/test/clientEncryption.test.js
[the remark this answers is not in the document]
nbbeeken: Setting it here will effect things globally, but we can align with the hook we have in the driver for this.

bindings/node/test/clientEncryption.test.js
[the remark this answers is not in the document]
nbbeeken: We discussed this and I think landed on being happy with the errors that come from destructuring. These errors are only encountered when folks "break" the TS contracts we've setup.

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,63 @@ module.exports = function (modules) {
...
+        );
+
+        if (createDataKeyResolutions.some(resolution => resolution.status === 'rejected')) {
+          throw new common.MongoCryptCreateEncryptedCollectionError(
+            'Could not complete creating data keys',
+            {
+              encryptedFields,
+              errors: createDataKeyResolutions.map(resolution =>
+                resolution.status === 'rejected' ? resolution.reason : null
+              )
```
nbbeeken, line 600: @baileympearson @addaleax 

This would results in an array with null entries for the operations that did not have an error. Alternatively, we could return:

```ts
errors: createDataKeyResolutions.reduce((acc, resolution, index) =>
  if (resolution.status === 'rejected') acc[index] = resolution.reason;
  return acc;
}, {})
```

An object with indexes as keys that map to the error. 

I think regardless of the decision we make it will bespoke to this function so users will have to rely on the documentation of _how_ to use the errors prop on the `MongoCryptCreateEncryptedCollectionError`

baileympearson (changes requested):

bindings/node/index.d.ts
```diff
@@ -497,6 +533,21 @@ export class ClientEncryption {
...
+   * Will create dataKeys for any fields missing a `keyId`
+   *
+   * @param db - The database to create the collection within
+   * @param name - The name of the new collection
+   * @param options - Options for createDataKey and for createCollection. A provider and partially created encryptedFields **must** be provided.
+   * @throws {MongoCryptCreateEncryptedCollectionError} - If part way through the process createDataKey fails, an error will be rejected that has the `encryptedFields` that were created.
+   */
+  createEncryptedCollection<TSchema extends Document = Document>(db: Db, name: string, options: {
+    provider: ClientEncryptionDataKeyProvider;
+    createCollectionOptions: Omit<CreateCollectionOptions, 'encryptedFields'> & { encryptedFields: NonNullable<CreateCollectionOptions['encryptedFields']> };
```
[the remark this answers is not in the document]
nbbeeken: We don't have types in the 3.x driver so I'm not worried. But this type would fail if `CreateCollectionOptions` does not have an `encryptedFields` property so should we leave this untyped / `Document`? it's frustrating that these could fall out of sync, but maybe it's safer.

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,63 @@ module.exports = function (modules) {
...
+        provider,
+        createDataKeyOptions = {},
+        createCollectionOptions: {
+          encryptedFields: { ...encryptedFields },
+          ...createCollectionOptions
+        }
+      } = options;
+
+      if (Array.isArray(encryptedFields.fields)) {
+        const createDataKeyPromises = [...encryptedFields.fields].map(async field =>
```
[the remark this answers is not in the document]
nbbeeken: fixed locally for the prettier, good catch

bindings/node/lib/clientEncryption.js
nbbeeken: And alternatively: 
- returning an array makes all the typical helpers available, map/filter/etc depending what the user needs to do with their error handling
- And the array isn't sparse so it defines all numeric keys btwn `'0'` and length of `encryptedFields.fields`

I'm up for either, let's get one more engineer's input (notably the tests shouldn't need any changes b/c of my `.nested.property` checks, whatever we decide let's remember to assert the expected outcome type)

bindings/node/lib/clientEncryption.js
nbbeeken: Followed up https://github.com/mongodb/specifications/pull/1375 we should throw an error that still communicates the `encryptedFields` settings. We now have two distinct error classes

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,63 @@ module.exports = function (modules) {
...
+        const createDataKeyResolutions = await Promise.allSettled(createDataKeyPromises);
+
+        encryptedFields.fields = createDataKeyResolutions.map((resolution, index) =>
+          resolution.status === 'fulfilled' ? resolution.value : encryptedFields.fields[index]
+        );
+
+        if (createDataKeyResolutions.some(resolution => resolution.status === 'rejected')) {
+          throw new common.MongoCryptCreateEncryptedCollectionError(
+            'Could not complete creating data keys',
+            {
```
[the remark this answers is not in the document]
nbbeeken: done

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +554,64 @@ module.exports = function (modules) {
...
+        if (wereErrors) {
+          const errors = createDataKeyResolutions.map(resolution =>
+            resolution.status === 'rejected' ? resolution.reason : null
+          );
+          throw new common.MongoCryptCreateEncryptedCollectionError(
+            'Could not complete creating data keys',
+            { encryptedFields, errors }
+          );
+        }
+      }
```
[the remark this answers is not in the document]
nbbeeken: This would loose the error's positional information w.r.t the data key that we failed to create. So we either need to:
- build up an object with numeric properties that map to an error
- create an array that is the same length as the fields input and put null where there was no errors
- modify this suggestion above slightly to map to a tuple/struct `{index, error}`

bindings/node/lib/clientEncryption.js
nbbeeken: We could also make the array compact if we maintained the index information like so:

```js
const errors = createDataKeyResolutions
  .map((resolution, index) => ({ index, error: resolution.reason }))
  .filter(({error}) => error != null);
```

nbbeeken: From virtual office @dariakp @baileympearson we discussed taking only the first error. This is updated.

bindings/node/test/clientEncryption.test.js
baileympearson: I think we should revisit this.  The spec says to silently continue but if we do, a non-array `encryptedFields` will silently succeed (this method will not reject), the `createCollection` call succeeds but no data keys were generated.

This pseudocode will fail, probably surprisingly, and require the user to manually delete the created collections:
```typescript
const options = { encryptedFields: {} }; // non-array
const newEncryptedFields = clientEncryption.createEncryptedCollection(.... , options);

const client = await new MongoClient(..., { encryptedFieldsMap: newEncryptedFields }).connect() // throws
```

There have been multiple modifications to the spec - I think we should consider this as well.  Thoughts?  @nbbeeken for visibility

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +559,69 @@ module.exports = function (modules) {
...
+          encryptedFields: { ...encryptedFields },
+          ...createCollectionOptions
+        }
+      } = options;
+
+      if (createDataKeyOptions.keyAltNames != null) {
+        throw new MongoCryptInvalidArgumentError(
+          'Cannot pass createDataKeyOptions with keyAltNames set'
+        );
+      }
```
baileympearson, line 590: It feels inconsistent to throw a type error for missing parameters in the destructuring and a `MongoCryptInvalidArgumentError` specifically for `keyAltNames`.  Now that we have `MongoCryptInvalidArgumentError`, we should use it for all invalid arguments.

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +828,220 @@ describe('ClientEncryption', function () {
...
+
+      db = client.db('createEncryptedCollectionDb');
+    });
+
+    afterEach(async () => {
+      sinon.restore();
+    });
+
+    context('validates input', () => {
+      it('throws if options are omitted', async () => {
```
baileympearson, line 853: Let's be specific about what errors these throw.  For example, `throws TypeError if options are omitted` (although I suggest above that we throw a `MongoCryptInvalidArgumentError`).

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +828,220 @@ describe('ClientEncryption', function () {
...
+        const error = await clientEncryption
+          .createEncryptedCollection(db, collectionName)
+          .catch(error => error);
+        expect(error).to.be.instanceOf(TypeError);
+      });
+
+      it('throws if options.createCollectionOptions are omitted', async () => {
+        const error = await clientEncryption
```
baileympearson, line 861: Let's try and make this more specific, maybe looking for a keyword in the error message?

This works now because there are no other nested destructured parameters in the options object.  However, suppose the logic changes to something like this in the future:

```typescript
{
        provider,
        createDataKeyOptions = {},
		option2: { ...fields },
        createCollectionOptions: {
          encryptedFields: { ...encryptedFields },
          ...createCollectionOptions
        }
      }
```

This test would still throw, but it would be throwing because `option2` was undefined, not `createCollectionOptions`.

A more defensive approach would be to create a complete options object that each test can clone and delete any relevant properties from.  then we're sure that each test is only missing the expected property.

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +828,220 @@ describe('ClientEncryption', function () {
...
+            createCollectionOptions: { encryptedFields: { fields: [] } },
+            createDataKeyOptions: { keyAltNames: ['blah!'] }
+          })
+          .catch(error => error);
+        expect(error).to.be.instanceOf(MongoCryptInvalidArgumentError, /keyAltNames/);
+      });
+    });
+
+    context('when input is not the correct type do not validate', () => {
+      it('when options.encryptedFields.fields is not an array no key generation is performed', async () => {
```
baileympearson, line 887: the `context` block should describe the scenario, the `it` block should describe the assertion.  Both the context blocks in these tests and the it blocks contain both scenario and assertions.  

Here's an example of breaking up scenario / assertion:

```typescript
context('when options.encryptedFields.fields is not an array', () => {
  it('does not generate any encryption keys', () => { ... });
}); 
```

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +828,220 @@ describe('ClientEncryption', function () {
...
+        expect(error).to.be.instanceOf(MongoCryptCreateDataKeyError);
+        expect(error.cause).to.equal(customError);
+
+        expect(error.encryptedFields).property('fields').that.is.an('array');
+        expect(error.encryptedFields.fields).to.have.lengthOf(
+          createCollectionOptions.encryptedFields.fields.length
+        );
+        expect(error.encryptedFields.fields).to.have.nested.property('[0].keyId', keyId);
+        expect(error.encryptedFields.fields).to.not.have.nested.property('[1].keyId');
+        expect(error.encryptedFields.fields).to.have.nested.property('[2].keyId', 'cool id!');
```
baileympearson, line 945: From the test output, it's not clear that we're actually testing that the `encryptedFields` is present on the error:

```
      when subroutines throw the error is wrapped
        ✔ throws MongoCryptCreateDataKeyError if createDataKey rejects
        ✔ throws MongoCryptCreateEncryptedCollectionError if createCollection rejects
      when there are nullish keyIds in the encryptedFields.fields array
        ✔ does not mutate the input fields array when generating data keys
        ✔ generates dataKeys for all null keyIds in the fields array
```

I suggest breaking this up into context + it blocks, so the output is clear.  As an example, the following output more closely documents what we expect to happen:

when createDataKeyRejects
- a MongoCryptCreateDataKeyError is thrown with error message < whatever >
- the `MongoCryptCreateDataKeyError` contains a partially filled encrypted fields map
- the `MongoCryptCreateDataKeyError` contains the first error in the array of rejected promises

baileympearson: this test breakup would apply to all the tests that contain assertions on multiple behaviors

baileympearson (changes requested):

bindings/node/lib/clientEncryption.js
nbbeeken: MongoCryptInvalidArgumentError is removed

bindings/node/test/clientEncryption.test.js
nbbeeken: I leave that up to the spec owners, so let's circle back when there's word.

Something to consider, the more validation we pack into this function the more coordination it needs to maintain with server-side data structures, maybe there's a future where fields isn't an array and this function can continue to be invoked without a breaking a new shape.

bindings/node/test/clientEncryption.test.js
nbbeeken: added TypeError

bindings/node/test/clientEncryption.test.js
nbbeeken: tests reorganized

bindings/node/test/clientEncryption.test.js
nbbeeken: broke these up

bindings/node/test/clientEncryption.test.js
nbbeeken: added regexps to the throw assertions

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +559,69 @@ module.exports = function (modules) {
...
     }
 
+    /**
+     * Creates a collection that has encrypted document fields
+     *
+     * @template {TSchema} - Schema for the collection being created
+     * @param {Db} db - A database to create the collection in
+     * @param {string} name - The name of the collection to be created
+     * @param {object} options - Options for createDataKey and for createCollection.
+     * @param {string} options.provider - provider name
```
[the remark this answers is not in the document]
nbbeeken: fixed

bindings/node/test/clientEncryption.test.js
nbbeeken: From slack: We filed DRIVERS-2540, and we will not do that work here.

bindings/node/test/clientEncryption.test.js
```diff
@@ -900,44 +902,59 @@ describe('ClientEncryption', function () {
...
         const stub = sinon.stub(clientEncryption, 'createDataKey');
-        const keyId = new Binary(Buffer.alloc(16, 0), 4);
         stub.onCall(0).resolves(keyId);
         stub.onCall(1).rejects(customError);
         stub.onCall(2).resolves(keyId);
 
-        const createCollectionOptions = {
-          encryptedFields: { fields: [{}, {}, { keyId: 'cool id!' }] }
-        };
         const error = await clientEncryption
```
baileympearson, line 933: any reason to not inline this into each test case?

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +827,258 @@ describe('ClientEncryption', function () {
...
+
+        return error;
+      };
+
+      it('throws MongoCryptCreateDataKeyError', async () => {
+        const error = await createDataKeyRejection();
+        expect(error).to.be.instanceOf(MongoCryptCreateDataKeyError);
+      });
+
+      it('thrown error has a cause set to the error that was thrown from createDataKey', async () => {
```
baileympearson, line 951: I think this test case should be more specific - we want to test that the cause is the first error in the rejected errors array (since we decided that we're always returning the first error we encounter in the array of errors).

baileympearson (changes requested):
some minor comments, overall looks good.  now prose testing

bindings/node/test/clientEncryption.test.js
nbbeeken: Like copy this to each it block? I put it here b/c it's used in an assertion in the last it block

bindings/node/test/clientEncryption.test.js
nbbeeken: Added case for first error

bindings/node/test/clientEncryption.test.js
baileympearson: from an organization standpoint, it might make more sense to

- put all the stubbing into a beforeEach hook
- each test then just calls `await clientEncryption.createEncryptedCollection`

no need to call a helper method in the test instead of calling the method directly

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +827,260 @@ describe('ClientEncryption', function () {
...
+        });
+
+        expect(createDataKeySpy.callCount).to.equal(0);
+        const options = createCollectionSpy.getCall(0).args[1];
+        expect(options).to.deep.equal({ encryptedFields: { fields: 'not an array' } });
+      });
+    });
+
+    context('when options.encryptedFields.fields elements are', () => {
+      it('not objects they are passed along to createCollection', async () => {
```
baileympearson, line 889:
```suggestion
    context('when options.encryptedFields.fields elements are not objects', () => {
      it('they are passed along to createCollection', async () => {
```

bindings/node/test/clientEncryption.test.js
```diff
@@ -815,4 +827,260 @@ describe('ClientEncryption', function () {
...
+
+        return error;
+      };
+
+      it('throws MongoCryptCreateEncryptedCollectionError', async () => {
+        const error = await createCollectionRejection();
+        expect(error).to.be.instanceOf(MongoCryptCreateEncryptedCollectionError);
+      });
+
+      it('thrown error has a cause set to the error that was thrown from createDataKey', async () => {
```
baileympearson, line 1002:
```suggestion
      it('thrown error has a cause set to the error that was thrown from createCollection', async () => {
```

baileympearson (changes requested):
test stuff, otherwise lgtm

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +558,65 @@ module.exports = function (modules) {
...
+        masterKey,
+        createCollectionOptions: {
+          encryptedFields: { ...encryptedFields },
+          ...createCollectionOptions
+        }
+      } = options;
+
+      if (Array.isArray(encryptedFields.fields)) {
+        const createDataKeyPromises = encryptedFields.fields.map(async field =>
+          typeof field !== 'object' || field.keyId != null
```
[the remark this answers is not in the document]
baileympearson: believe it or not - no 😄 

From the [algorithm](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#create-encrypted-collection-helper):

> If F is not a document element, skip it.

bindings/node/test/clientEncryption.test.js
nbbeeken: done

bindings/node/lib/errors.js
```diff
@@ -0,0 +1,48 @@
...
+
+  get name() {
+    return 'MongoCryptCreateDataKeyError';
+  }
+}
+
+/** @class */
+class MongoCryptCreateEncryptedCollectionError extends MongoCryptError {
+  constructor({ encryptedFields, cause }) {
+    super('Unable to create collection', { cause });
```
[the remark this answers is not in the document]
nbbeeken: Good idea, updated, with tests!

bindings/node/index.d.ts
```diff
@@ -497,6 +524,22 @@ export class ClientEncryption {
...
+   * @param db - The database to create the collection within
+   * @param name - The name of the new collection
+   * @param options - Options for createDataKey and for createCollection. A provider and partially created encryptedFields **must** be provided.
+   * @throws {MongoCryptCreateDataKeyForEncryptedCollectionError} - If part way through the process a createDataKey invocation fails, an error will be rejected that has the partial `encryptedFields` that were created.
+   * @throws {MongoCryptCreateEncryptedCollectionError} - If creating the collection fails, an error will be rejected that has the entire `encryptedFields` that were created.
+   */
+  createEncryptedCollection<TSchema extends Document = Document>(db: Db, name: string, options: {
+    provider: ClientEncryptionDataKeyProvider;
+    createCollectionOptions: Omit<CreateCollectionOptions, 'encryptedFields'> & { encryptedFields: Document };
+    createDataKeyOptions?: ClientEncryptionCreateDataKeyProviderOptions;
```
baileympearson, line 540: This should be updated to only take `masterKey`

bindings/node/lib/clientEncryption.js
```diff
@@ -166,10 +177,8 @@ module.exports = function (modules) {
     /**
      * Creates a data key used for explicit encryption and inserts it into the key vault namespace
      *
-     * @param {string} provider The KMS provider used for this data key. Must be `'aws'`, `'azure'`, `'gcp'`, or `'local'`
-     * @param {object} [options] Options for creating the data key
-     * @param {AWSEncryptionKeyOptions|AzureEncryptionKeyOptions|GCPEncryptionKeyOptions} [options.masterKey] Idenfities a new KMS-specific key used to encrypt the new data key
-     * @param {string[]} [options.keyAltNames] An optional list of string alternate names used to reference a key. If a key is created with alternate names, then encryption may refer to the key by the unique alternate name instead of by _id.
+     * @param {string} provider The KMS provider used for this data key.
```
baileympearson, line 180: let's revert the changes to `ClientEncryptionCreateDataKeyProviderOptions`, since don't need them anymore

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +558,65 @@ module.exports = function (modules) {
       return value;
     }
 
+    /**
+     * A convenience method for creating an encrypted collection.
+     * This method will create data keys for any encryptedFields that do not have a `keyId` defined
+     * and then create a new collection with the full set of encryptedFields.
+     *
+     * @template {TSchema} - Schema for the collection being created
+     * @param {Db} db - A database to create the collection in
```
baileympearson, line 567:
```suggestion
     * @param {Db} db - A Node driver Db object with which to create the collection
```

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +558,65 @@ module.exports = function (modules) {
...
+     * @param {string} name - The name of the collection to be created
+     * @param {object} options - Options for createDataKey and for createCollection.
+     * @param {string} options.provider - KMS provider name
+     * @param {AWSEncryptionKeyOptions | AzureEncryptionKeyOptions | GCPEncryptionKeyOptions} [options.masterKey] - masterKey to pass to createDataKey
```
baileympearson, line 571: this type needs to reflect that we only accept the master key, right?

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +558,65 @@ module.exports = function (modules) {
...
 
+    /**
+     * A convenience method for creating an encrypted collection.
+     * This method will create data keys for any encryptedFields that do not have a `keyId` defined
+     * and then create a new collection with the full set of encryptedFields.
+     *
+     * @template {TSchema} - Schema for the collection being created
+     * @param {Db} db - A database to create the collection in
+     * @param {string} name - The name of the collection to be created
+     * @param {object} options - Options for createDataKey and for createCollection.
```
baileympearson, line 569:
```suggestion
     * @param {object} options - Options for createDataKey and for createCollection
```

let's be consistent with either always ending with '.'s or never ending with them.

bindings/node/lib/clientEncryption.js
```diff
@@ -549,6 +558,65 @@ module.exports = function (modules) {
...
+      } = options;
+
+      if (Array.isArray(encryptedFields.fields)) {
+        const createDataKeyPromises = encryptedFields.fields.map(async field =>
+          typeof field !== 'object' || field.keyId != null
+            ? field
+            : {
+                ...field,
+                keyId: await this.createDataKey(provider, { masterKey })
+              }
```
baileympearson, line 594:
```suggestion
  if (typeof field !== 'object' || field.keyId != null) {
    return field;
  }
  return {
                ...field,
                keyId: await this.createDataKey(provider, { masterKey })
              }
```

excuse the formatting here.  ternaries where we're returning objects and cloning are far less readable than an if-statement + return.

baileympearson (changes requested):

bindings/node/lib/clientEncryption.js
nbbeeken: it does, no?

- `options.createCollectionOptions`
- `options.provider`
- `[options.masterKey]`

bindings/node/lib/clientEncryption.js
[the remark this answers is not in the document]
addaleax: Right – my point is, `field` could still be `null` and in that case this line would throw (just as because as a general rule all `typeof x === 'object'` checks need an `x !== null` check).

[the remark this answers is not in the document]
nbbeeken: ah yea, good point, adding

bindings/node/lib/clientEncryption.js
baileympearson: Ah you're correct.

baileympearson (approved):
LGTM, pending CI

baileympearson (approved):

PR #553, opened 2023-01-30
MONGOCRYPT-525 do not assert on base64 decode error

kevinAlbs:
# Summary

- do not assert on base64 decode error

# Background & Motivation

Without the fix, `./cmake-build/test-mongocrypt _test_kms_responses` results in the assert:

```
- Encryption response with invalid base64
/Users/kevin.albertson/code/libmongocrypt-M525/src/mongocrypt-kms-ctx.c:590 _ctx_done_aws(): precondition failed: result_len >= 0
```


addaleax (approved):

rcsanchez97 (approved):
LGTM

PR #554, opened 2023-01-30
[MONGOCRYPT-524] Fix ILP32-target builds

vector-of-bool:
This fixes a confusion about the data model available on certain platforms.

This changeset also defines an additional `debian-package-build-i386` task that will validate this platform going forward. We also now have an Earthly target for the deb-build, and the changes were validated there (using `earthly --platform=i386 +deb-build`).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -855,14 +855,36 @@ tasks:
         script: |-
           set -o errexit
           set -o xtrace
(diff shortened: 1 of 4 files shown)


rcsanchez97 (approved):
LGTM

kevinAlbs (approved):
LGTM

PR #556, opened 2023-02-02
Define a new EVG task that runs the RPM build

vector-of-bool:
This PR adds a new EVG task to the "Linux Distro Packaging" group that builds an RPM using Fedora Rawhide, based on the same RPM spec that is used in the upstream. This will be the first Earthly-based task running in EVG.

This changeset also caught an accidental new built-time dependency on Git that wasn't listed in the upstream `libmongocrypt.spec`. I hope that this will allow us to preempt RPM-related build issues in the future, before merging errant changesets.

Tweaks:
  - Don't BuildRequires git, and just use 'patch' on the IntelDFP sources if possible.
  - Only install exactly what is requried by BuildRequires for creating the RPM, to detect accidental new build dependencies.

(I note that the Docker availability and stability in the EVG distros isn't entirely stable, but I have narrowed it down to a generally reliable set.)

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -917,7 +917,7 @@ tasks:
     - func: fetch source
     - func: earthly
       vars:
-        args: +rpm-build
+        args: +rpm-runtime-test
 
 pre:
   # Update the evergreen expansion to dynamically set the ${libmongocrypt_s3_suffix} and ${libmongocrypt_s3_suffix_copy} expansions.
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -352,6 +352,12 @@ functions:
         permissions: public-read
         content_type: ${content_type|application/gzip}
         display_name: Release Python files all
+  earthly:
+    - command: shell.exec
+      params:
+        shell: bash
+        working_dir: ${working_dir|libmongocrypt}
+        script: bash ${workdir}/libmongocrypt/.evergreen/earthly.sh ${args}
 
 tasks:
 - name: build-and-test-and-upload
@@ -845,6 +851,10 @@ tasks:
         content_type: '${content_type|application/x-gzip}'
 
 - name: debian-package-build
+  run_on: &deb-package-build-run_on
+    - ubuntu2004
+    - ubuntu2004-small
+  tags: [packaging]
   commands:
     - func: "fetch source"
     - command: shell.exec
@@ -868,6 +878,8 @@ tasks:
         display_name: "deb.tar.gz"
 
 - name: debian-package-build-i386
+  run_on: *deb-package-build-run_on
+  tags: [packaging]
   commands:
     - func: "fetch source"
     - command: shell.exec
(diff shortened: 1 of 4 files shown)


Earthfile
```diff
@@ -240,14 +240,18 @@ COPY_SOURCE:
 
 rpm-build:
     FROM +init --base fedora:rawhide
```
rcsanchez97, line 242: I'm assuming that this creates a clean environment based on Fedora rawhide and so we don't need to run `mock`.  Is that right?

Earthfile
```diff
@@ -240,14 +240,18 @@ COPY_SOURCE:
...
-    RUN cp -r /s/libmongocrypt/. .
-    RUN awk -f etc/rpm/tweak.awk < libmongocrypt.spec > libmongocrypt.2.spec
-    RUN rpmbuild -ba libmongocrypt.2.spec -v -D "_sourcedir $PWD"
+    RUN cp -r /s/libmongocrypt/. /R
+    RUN awk -f /R/etc/rpm/tweak.awk < /R/libmongocrypt.spec > /R/libmongocrypt.2.spec
+    RUN rpmbuild -ba /R/libmongocrypt.2.spec \
+        -D "_topdir /X" \
+        -D "_sourcedir /R"
+    SAVE ARTIFACT /X/RPMS /rpm/
+    SAVE ARTIFACT /X/SRPMS /rpm/
```
rcsanchez97, line 254: I don't recognize the `SAVE` directive, so I'm assuming that it's related to Earthly rather than directly to Docker.  That said, I can't seem to find what, if anything, is done with these RPMs after they are built.  Note that in the C driver, which uses a script to configure a `mock` environment and perform the build within that, the last thing which we do is build a sample program to ensure that the system's toolchain are able to properly reference the installed resources:

```sudo mock -r ${config} --use-bootstrap-image --isolation=simple --cwd "/tmp/${build_dir}" --chroot -- /bin/sh -c "(
  rpm -Uvh ../mock-result/*.rpm &&
  gcc -I/usr/include/libmongoc-1.0 -I/usr/include/libbson-1.0 -o example-client src/libmongoc/examples/example-client.c -lmongoc-1.0 -lbson-1.0
  )"
```

Ideally, this task would install the resulting RPMs in the rawhide environment and then perhaps compile `./test/example-state-machine.c`.

I should have made a comment along these same in the PR (#554) where you added the Debian task to this file.

rcsanchez97 (changes requested):

rcsanchez97:
@vector-of-bool BTW, it would be good to cherry-pick this into r1.7 after you finalize it and merge.  That way Remi can have a clean working build for his packaging efforts.

Earthfile
vector-of-bool: That is correct, as an additional `chroot` may be somewhat redundant. Other Mock features might be useful here, but not at the moment, as far as I am aware.

Earthfile
vector-of-bool: Yes, `SAVE` is an Earthly thing that tells it was files to "export" from a build (by default they are thrown into the void).

I'll define a task to build a simple project.

rcsanchez97 (approved):
I love it!  LGTM!

.evergreen/config.yml
```diff
@@ -352,6 +352,12 @@ functions:
         permissions: public-read
         content_type: ${content_type|application/gzip}
         display_name: Release Python files all
+  earthly:
+    - command: shell.exec
+      params:
+        shell: bash
+        working_dir: ${working_dir|libmongocrypt}
+        script: bash ${workdir}/libmongocrypt/.evergreen/earthly.sh ${args}
```
kevinAlbs, line 360: Are `working_dir` and `workdir` intended to be separate variables?

vector-of-bool: Yes. `workdir` is the directory EVG allocates for the task, but `working_dir` just changes where the subprocess will run. For now, it isn't important to tweak it, since there's only one directory with an Earthfile in it. The similarity in the variable names is unfortunate, I suppose.

PR #561, opened 2023-02-06
MONGOCRYPT-532 Fix i686 (32-bit) build

remicollet:
value taken from RPM %{ix86} definition

```
$ rpm --eval %{ix86}
i386 i486 i586 i686 pentium3 pentium4 athlon geode
```


remicollet:
Fedora rawhide scratch build using this patch
https://koji.fedoraproject.org/koji/taskinfo?taskID=97194696

vector-of-bool (approved):
Thanks for the catch. Debian errors are unrelated. LGTM.

PR #563, opened 2023-02-07
Permit -DENABLE_MONGOC=OFF when compiling libmongocrypt

eramongodb:
Addresses the following error when compiling with `-DENABLE_MONGOC=OFF`:

```
CMake Error at cmake/ImportBSON.cmake:153 (target_include_directories):
  Cannot specify include directories for target "mongoc_static" which is not
  built by this project.
  cmake/ImportBSON.cmake:162 (_import_bson)
  CMakeLists.txt:22 (include)
```

The conditional presence of `mongoc_static` is already being handled correctly further below in ImportBSON.cmake:

```cmake
if (TARGET mongoc_static)
   # And an alias to the mongoc target for use in some test cases
   add_library (_mongocrypt::mongoc ALIAS mongoc_static)
endif ()
```

As a drive-by fix, also addresses a -Wtautological-compare warning in mc-fle2-payload-iev.c which may hinder compilation from source by the C Driver.

### DIFF
--- a/cmake/ImportBSON.cmake
+++ b/cmake/ImportBSON.cmake
@@ -149,12 +149,14 @@ function (_import_bson)
       else ()
          add_subdirectory ("${MONGOCRYPT_MONGOC_DIR}" _mongo-c-driver EXCLUDE_FROM_ALL)
       endif ()
-      # Workaround: Embedded mongoc_static does not set its INCLUDE_DIRECTORIES for user targets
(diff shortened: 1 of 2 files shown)


vector-of-bool (approved):
I had seen this issue a while back but hadn't bothered to push a fix. This can help speed up builds when we don't need mongoc. LGTM. Thanks!

kevinAlbs (approved):

PR #565, opened 2023-02-08
fix(NODE-5053): enforce empty map for kmsProvider auto credentials

nbbeeken:
#### What is changing?

We strictly check for an existing object but with zero keys for a kmsProviders credentials before attempting auto fetching.

##### Is there new documentation needed for these changes?

Yes, this could be surprising to existing AWS users that were able to leave kmsProviders empty. They need to provide an 'aws' key set to `{}`

The kmsProviders settings needs to have a key that exists but is set to an empty object in order to enable auto obtaining credentials. Otherwise future credential providers will both be obtained without actually requesting it.

- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`

### DIFF
--- a/bindings/node/lib/credentialsProvider.js
+++ b/bindings/node/lib/credentialsProvider.js
@@ -1,18 +1,27 @@
 'use strict';
 
-let awsCredentialProviders = null;
-
 /**
- * Load cloud provider credentials for the user provided KMS providers.
- * Credentials will only attempt to get loaded if they do not exist
- * and no existing credentials will get overwritten.
- *
- * @param {Object} kmsProviders - The user provided KMS providers.
- * @returns {Promise} The new kms providers.
- *
  * @ignore
+ * Auto credential fetching should only occur when the provider is defined on the kmsProviders map
+ * and the settings are an empty object.
+ *
+ * This is distinct from a nullish provider key.
+ *
+ * @param {string} provider
+ * @param {object} kmsProviders
  */
-async function loadCredentials(kmsProviders) {
+function isEmptyCredentials(provider, kmsProviders) {
+  return (
+    provider in kmsProviders &&
+    kmsProviders[provider] != null &&
+    typeof kmsProviders[provider] === 'object' &&
+    Object.keys(kmsProviders[provider]).length === 0
+  );
+}
+
+let awsCredentialProviders = null;
+/** @ignore */
+async function loadAWSCredentials(kmsProviders) {
   if (awsCredentialProviders == null) {
     try {
       // Ensure you always wrap an optional require in the try block NODE-3199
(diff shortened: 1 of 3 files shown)


bindings/node/test/credentialsProvider.test.js
```diff
@@ -82,81 +73,114 @@ describe('#loadCredentials', function () {
...
+          it('does not refresh credentials', async function () {
+            const providers = await loadCredentials(kmsProviders);
+            expect(providers).to.deep.equal(kmsProviders);
+          });
         });
 
-        it('refreshes ony the aws credentials', async function () {
-          const providers = await loadCredentials(kmsProviders);
-          expect(providers).to.deep.equal({
+        context.skip('when aws does not exist', function () {
```
nbbeeken, line 132: Should I just delete this block?

bindings/node/lib/credentialsProvider.js
```diff
@@ -1,18 +1,27 @@
...
+function isEmptyCredentials(provider, kmsProviders) {
+  return (
+    provider in kmsProviders &&
+    kmsProviders[provider] != null &&
+    typeof kmsProviders[provider] === 'object' &&
+    Object.keys(kmsProviders[provider]).length === 0
+  );
+}
+
+let awsCredentialProviders = null;
```
addaleax, line 22: Btw, it is so nice that this package does dependency injection for the driver package – It might even be worth adapting that to this dependency as well … :)

addaleax (approved):

bindings/node/lib/credentialsProvider.js
baileympearson: How come you think this is a beneficial pattern?  The DI for the driver is something I've wanted to remove for a while

addaleax: Well, from a very practical point, it’s just a necessity in cases where the driver is bundled into something but this package isn’t (which would be common in bundling scenarios given that this package is a native addon package).

But also just philosophically (and I’m pretty sure you already know this), I think DI is a great design pattern that solves more issues than it creates :slightly_smiling_face:

addaleax: Just to clarify, this thought is not directly related to the changes in this PR and should absolutely not block it :)

nbbeeken: We should consider this though, using it for testing too could be 🔥 

For GCP we don't plan to use an SDK (and likely not for azure either) so this would only be for AWS but still, valid use case even for one dep.

bindings/node/test/credentialsProvider.test.js
```diff
@@ -2,62 +2,56 @@
...
-        const providers = await loadCredentials(kmsProviders);
-        expect(providers).to.deep.equal({
-          aws: {
-            accessKeyId: accessKey,
-            secretAccessKey: secretKey,
-            sessionToken: sessionToken
-          }
-        });
-      });
+  context('aws', () => {
```
durran, line 34:
```suggestion
  context('when using aws', () => {
```

bindings/node/test/credentialsProvider.test.js
durran: Yes I'd just remove it.

durran (changes requested):

bindings/node/test/credentialsProvider.test.js
nbbeeken: done

bindings/node/test/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('extension returns an object equal in shape to the default except for extension', () => {
+    const extensionResult = mongodbClientEncryption.extension(require('mongodb'));
+    const expectedExports = EXPECTED_EXPORTS.filter(exp => exp !== 'extension');
+    const exportsDefault = Object.keys(mongodbClientEncryption).filter(exp => exp !== 'extension');
+    expect(extensionResult).to.have.all.keys(expectedExports);
+    expect(extensionResult).to.have.all.keys(exportsDefault);
+  });
+});
```
nbbeeken, line 29: The thread about DI made me look in to the exports and how they work with `extension`, I just want to make sure that this doesn't break in the future by accident

bindings/node/test/credentialsProvider.test.js
nbbeeken: done

durran (approved):

PR #567, opened 2023-02-09
MONGOCRYPT-534 Fix type detection: No trying to guess the type of 'uint64_t'

vector-of-bool:
Instead, rely on `unsigned long long` being the same _size_ as `uint64_t`, and just do the `clzll` on that value.


addaleax (approved):

kevinAlbs (approved):

rcsanchez97 (approved):
LGTM.  I also asked Kyle to push the necessary associated changes to the Debian packaging.  Once you merge this, the `debian-package-build` and `debian-package-build-i386` tasks should turn green again.

PR #569, opened 2023-02-10
[MONGOCRYPT-536] A Toggle for Decimal128 Range Support

vector-of-bool:
The configure-time option `MONGOCRYPT_ENABLE_DECIMAL128` can be used to enable/disable Decimal128 range support. This is enabled by default.

- Wrap Decimal128 code in conditions on whether Decimal128 is available.
- Only supports IntelDFP for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,13 @@ include (GNUInstallDirs)
 
 include (LTO)
 include (ImportBSON)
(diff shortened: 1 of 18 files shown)


rcsanchez97 (approved):
LGTM, modulo my comment in Slack about the tasks that are still failing.

kevinAlbs (approved):

PR #577, opened 2023-02-16
Reformat all the files

vector-of-bool:
Changes:
- Applies new formatting rules to all C and C++ code.
- Defines a check-format EVG task. While this could be annoying as a PR gate, it discourages unformatted code from passing through and then later being automatically formatted as part of an unrelated change and making a mess of the git-blame.
- Defines a turnkey "format everything" script that makes it easier for devs to ensure they match the exact formatting checked by CI (including a pinned clang-format version).

A Git `blame.ignoreRevsFile` requires a specific commit to ignore. Squash-merging this PR will generate an unpredictable commit hash. This PR either needs to be merge with a different method to allow ignoring the commit `a531580` (which contains the reformat operation) or a subsequent PR will need to be created to add the ignore-revs file for the commit hash.


Earthfile
```diff
@@ -343,6 +343,15 @@ packaging-full-test:
...
     BUILD +rpm-runtime-test
 
+check-format:
+    FROM python:3.11.2-slim-buster
+    RUN pip install pipx
+    COPY etc/clang-format*.sh /X/etc/
+    COPY .evergreen/init.sh /X/.evergreen/
+    RUN bash /X/etc/clang-format.sh  # Does nothing, but warms the cache
+    COPY --dir .clang-format src /X/
+    RUN bash /X/etc/clang-format-all.sh --dry-run -Werror
```
eramongodb, line 353: Can `bash` be omitted when invoking the script(s) to encourage using/respecting shebangs within the file(s)?

etc/clang-format-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
+
+files="$(find "$LIBMONGOCRYPT_DIR/src" -type f -name '*.c' -o -name '*.h' -o -name '*.cpp' -o -name '*.hpp')"
+readarray -t files <<< "$files"
+bash "$LIBMONGOCRYPT_DIR/etc/clang-format.sh" \
```
eramongodb, line 8: Can `bash` be omitted when invoking the script to encourage using/respecting shebangs within the file?

etc/clang-format.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if ! run_python -c ''; then
+  fail "No Python found?"
+fi
+
+if ! run_python -c 'import pipx'; then
+  fail "No PipX found?"
+fi
+
+printf '' | run_python -m pipx run "clang-format==$CLANG_FORMAT_VERSION" "$@"
```
eramongodb, line 16: Suggest documenting the purpose of the `printf ''` being piped to pipx.

Also suggest using `${CLANG_FORMAT_VERSION:?}`; if user-defined, it still shouldn't be null.

etc/clang-format-all.sh
```diff
@@ -0,0 +1,12 @@
...
+    --style="file:$LIBMONGOCRYPT_DIR/.clang-format" \
```
eramongodb, line 9: It seems like clang-format doesn't support an option to guarantee the presence of an EOF newline. Is this something we can add to either `clang-format-all.sh` or `clang-format.sh` to ensure [C standard conformance](https://port70.net/~nsz/c/c99/n1256.html#5.1.1.2p1)?

> A source file that is not empty shall end in a new-line character, which shall not be immediately preceded by a backslash character before any such splicing takes place.

etc/clang-format.sh
```diff
@@ -0,0 +1,16 @@
...
+
+: "${CLANG_FORMAT_VERSION:=15.0.7}"
+export PIPX_HOME="${BUILD_CACHE_DIR}/pipx"
+
+if ! run_python -c ''; then
+  fail "No Python found?"
+fi
+
+if ! run_python -c 'import pipx'; then
+  fail "No PipX found?"
```
kevinAlbs, line 13:
```suggestion
  fail "No PipX found? Try `pip install pipx`"
```

etc/clang-format-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
+
+files="$(find "$LIBMONGOCRYPT_DIR/src" -type f -name '*.c' -o -name '*.h' -o -name '*.cpp' -o -name '*.hpp')"
+readarray -t files <<< "$files"
```
kevinAlbs, line 6: Is there an alternative to `readarray` for older bash? The default bash on M1 macBook is:

```
% bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin21)
```

etc/clang-format-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
```
eramongodb, line 3: The `|| have_command cmd.exe` in:

https://github.com/mongodb/libmongocrypt/blob/73302cd485cd63becec92a29dff5f09f2731b4ab/.evergreen/init.sh#L145

was preventing my attempt to execute this script in WSL2:

```
Uname is [Linux <hostname> 5.15.79.1-microsoft-standard-WSL2 #1 SMP Wed Nov 23 01:01:46 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux]
```

It's not directly related to these changes, but can the check for `cmd.exe` be removed?

etc/clang-format.sh
eramongodb: I am observing the following error when running this script locally:
```
'clang-format==15.0.7' executable script not found in package 'clang-format==15.0.7'. Available executable scripts: clang-format, clang-format-diff.py, git-clang-format
```
Changing `clang-format==$CLANG_FORMAT_VERSION` to simply `clang-format` seems to resolve this issue, but I'm unsure this is the correct fix. Can/should the required clang-format version be specified by another means?

etc/clang-format.sh
```diff
@@ -0,0 +1,16 @@
...
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
+
+: "${CLANG_FORMAT_VERSION:=15.0.7}"
+export PIPX_HOME="${BUILD_CACHE_DIR}/pipx"
+
+if ! run_python -c ''; then
+  fail "No Python found?"
+fi
+
+if ! run_python -c 'import pipx'; then
```
eramongodb, line 12: pipx seems to require Python 3.6 or newer. `run_python` defefined here:

https://github.com/mongodb/libmongocrypt/blob/73302cd485cd63becec92a29dff5f09f2731b4ab/.evergreen/init.sh#L222

seems to simply select `python` which defaults to Python 2 in many environments. Can Python binary selection be improved to ensure a Python 3 binary is used for pipx?

etc/clang-format.sh
vector-of-bool: That's odd. What version of pipx/Python are you using?

Earthfile
vector-of-bool: Yes, but also it can help the error messages be more clear if Bash isn't available (Otherwise you get an ENOENT when trying to execute the file, which can be unintuitive) and doesn't require the file to be executable.

etc/clang-format.sh
eramongodb: Python 3.8.10 with pipx 0.16.4. If I edit `init.sh` to use Python 3.11 instead, it uses pipx 1.1.0 and I no longer observe the error.

vector-of-bool: Ahh. Should we just require a new-enough pipx?

eramongodb: I think so. [Release notes](https://pypa.github.io/pipx/changelog/#0170) suggest this was implemented in 0.17.0, which seems consistent with the observed failure with 0.16.4.

etc/clang-format.sh
vector-of-bool: Yes. Python selection needs to be improved in general, since the filename varies sporadically between platforms. It's very obnoxious.

etc/clang-format-all.sh
vector-of-bool: It's possible, yes, but may require additional scripting. (That seems like a significant omission for clang-format.)

etc/clang-format-all.sh
vector-of-bool: Likely yes. I'll investigate that.

etc/clang-format-all.sh
eramongodb: (Resolving; discussion continuing in other thread: https://github.com/mongodb/libmongocrypt/pull/577#discussion_r1109024061)

etc/clang-format-all.sh
eramongodb: The applied update does not seem to work as intended (case-sensitivity of `grep -q 'Microsoft'` failing to match `/proc/version` contents?), but given this is not directly related to this PR (and I can apply local workarounds anyways), I am not opposed to discarding these changes to simplify this PR.

Note: for reference, the following edit to `init.sh` seems to work:

```diff
- _uname="$(uname -a | tr '[:upper:]' '[:lower:]')"
+ _uname="$(uname | tr '[:upper:]' '[:lower:]')"

- if [[ "$_uname" =~ .*cywin|windows|mingw|msys.* ]] || have_command cmd.exe; then
+ if [[ "$_uname" =~ .*cywin|windows|mingw|msys.* ]] || (have_command cmd.exe && [[ ! "$_uname" =~ .*wsl.* ]]); then
```

Note 2: should `cywin` be `cygwin`?

eramongodb (changes requested):

etc/format-newlines.py
```diff
@@ -0,0 +1,97 @@
+"""
```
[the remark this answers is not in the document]
eramongodb: Note 2: the `tail` + command substitution method doesn't handle the presence of non-printable characters at the end of the file (i.e. `'\0'`), but if there are ever such special characters (I don't know why there ever would be), we'd know pretty quickly via other means:

```bash
$ printf "int main(void) {}\0" >|main.c
$ clang -std=c99 -pedantic main.c
main.c:1:18: warning: null character ignored [-Wnull-character]
int main(void) {}<U+0000>
                 ^
main.c:1:19: warning: no newline at end of file [-Wnewline-eof]
int main(void) {}<U+0000>
                         ^
```

etc/clang-format-all.sh
vector-of-bool: TIL. Unfortunately, version 16 is still in an RC stage, so there is no PyPI package yet. It should be available in March. We can defer that changeset for a month, then bump to the new version once it's available.

etc/clang-format-all.sh
vector-of-bool: Thanks for checking and providing a proper fix. I'll remove these changes and put in a new PR later. And yes, that's a typo.

eramongodb: Note: fixed diff to indicate _addition_ of `-a` to `uname` command rather than removal as otherwise `_uname` will simply be `linux`. The required "5.15.79.1-microsoft-standard-WSL2" portion containing "wsl" is part of the kernel release (`-r`/`--kernel-release`), not the kernel name (`-s`/`--kernel-name`, default. Using `-sr` instead of `-a` should therefore also suffice.

etc/format-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
+
+files="$(find "$LIBMONGOCRYPT_DIR/src" -type f -name '*.c' -o -name '*.h' -o -name '*.cpp' -o -name '*.hpp')"
+# shellcheck disable=SC2206
+IFS=$'\n' files=($files)
+bash "$LIBMONGOCRYPT_DIR/etc/format.sh" \
+    --style="file:$LIBMONGOCRYPT_DIR/.clang-format" \
+    -i \
```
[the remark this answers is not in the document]
vector-of-bool: The `--verbose` option can be specified when the script is invoked, as those arguments are forwarded to `clang-format`

.clang-format
```diff
@@ -1,37 +1,25 @@
 ---
+Language: Cpp
+IndentWidth: 4
 AlignAfterOpenBracket: Align
```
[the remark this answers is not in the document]
eramongodb: Confirming, was this suggestioned missed or rejected? If rejected, feel free to mark as resolved.

.clang-format
```diff
@@ -1,37 +1,25 @@
...
-...
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBraces: Attach
+ColumnLimit: 120
+ContinuationIndentWidth: 2
+IndentCaseLabels: false
+KeepEmptyLinesAtTheStartOfBlocks: false
```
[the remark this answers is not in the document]
eramongodb: Confirming, was this suggestioned missed or rejected? If rejected, feel free to mark as resolved.

.clang-format
```diff
@@ -1,37 +1,25 @@
...
+MaxEmptyLinesToKeep: 2
+PointerAlignment: Right
+QualifierAlignment: Custom
+QualifierOrder: ["static", "inline", "const", "type"]
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
```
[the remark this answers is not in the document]
eramongodb: Resolving in favor of new comments.

etc/format-all.sh
```diff
@@ -0,0 +1,12 @@
...
+    "$@" \
+    -- "${files[@]}"
```
eramongodb, line 12:
```suggestion
    -- "${files[@]:?}"
```

Despite clang-format emitting an error when no files are given when `-i` is present, it still returns an exit code of 0 indicating success:
```bash
$ clang-format -i; echo $?
error: cannot use -i when reading from stdin.
0
```
Suggest asserting `$files` is not null by adding `:?`.

etc/format.sh
```diff
@@ -0,0 +1,16 @@
...
+
+: "${CLANG_FORMAT_VERSION:=15.0.7}"
+export PIPX_HOME="${BUILD_CACHE_DIR}/pipx"
+
+if ! run_python -c ''; then
+  fail "No Python found?"
+fi
+
+# Check that we have a pipx of the proper version:
+run_python -c 'import pkg_resources; pkg_resources.require("pipx>=0.17.0<2.0")'
```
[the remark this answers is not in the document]
vector-of-bool: I'm testing with Python 3.10 and Python 3.11, not anything older than 3.8. `pkg_resources` is discouraged when importlib suffices, but it is not deprecated. The `require()` function is not part of `importlib`.

You have `importlib-metadata` 1.5.0 installed, but pipx requires 3.3.0+ when installed in Python <3.8. Version 3.3.0 was released in Dec 2020. How did you install pipx without it failing due to the unsatisfied `importlib-metadata`?

etc/format-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
+
+files="$(find "$LIBMONGOCRYPT_DIR/src" -type f -name '*.c' -o -name '*.h' -o -name '*.cpp' -o -name '*.hpp')"
```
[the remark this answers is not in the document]
vector-of-bool: Just missed. Added.

.clang-format
```diff
@@ -1,37 +1,25 @@
...
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBraces: Attach
+ColumnLimit: 120
+ContinuationIndentWidth: 2
+IndentCaseLabels: false
+KeepEmptyLinesAtTheStartOfBlocks: false
+MaxEmptyLinesToKeep: 2
```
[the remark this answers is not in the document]
vector-of-bool: I can remove `MaxEmptyLinesToKeep`, but that also includes contexts within definitions (e.g. function bodies). Do we also want to remove spacing in there? I'm fine with that, I just carried it over from the prior format file.

.clang-format
```diff
@@ -1,37 +1,31 @@
...
+PointerAlignment: Right
+QualifierAlignment: Custom
+QualifierOrder: ["static", "inline", "const", "volatile", "type", "restrict"]
+SeparateDefinitionBlocks: Always
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpacesInLineCommentPrefix:
+  Minimum: 1
+  Maximum: -1
```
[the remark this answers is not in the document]
vector-of-bool: Added

.clang-format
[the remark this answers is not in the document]
vector-of-bool: Missed. Added and verified that everything is OK.

.clang-format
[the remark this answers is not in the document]
vector-of-bool: I just moved it down and sorted everything, no "special" options at the top.

etc/format.sh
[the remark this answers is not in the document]
eramongodb: Good catch; unsure. Python when invoked directly seems to indicate `importlib-metadata` version 6.0.0. Manually uninstalling and reinstalling the `importlib-metadata` package seems to have resolved the issue.

I'm in favor of not having to bother with supporting Python<3.8; although I still suggest using recommended packages over `pkg_resources`, I won't consider it a blocker.

etc/format-all.sh
```diff
@@ -0,0 +1,12 @@
...
+. "$(dirname "${BASH_SOURCE[0]}")/../.evergreen/init.sh"
+
+files="$(find "$LIBMONGOCRYPT_DIR/src" "$LIBMONGOCRYPT_DIR/test" -type f -name '*.c' -o -name '*.h' -o -name '*.cpp' -o -name '*.hpp')"
+# shellcheck disable=SC2206
+IFS=$'\n' files=($files)
+bash "$LIBMONGOCRYPT_DIR/etc/format.sh" \
+    --style="file:$LIBMONGOCRYPT_DIR/.clang-format" \
+    -i \
+    "$@" \
+    -- "${files[@]:?}"
```
eramongodb, line 12:
```suggestion
    -- "${files[@]:?}"

# TODO: remove once we can use clang-format 16 InsertNewlineAtEOF.
for file in "${files[@]}"; do
    [[ -z "$(tail -c1 "$file")" ]] || printf "\n" >>"$file"
done
```
Still suggest including a cheap EOF newline test-and-append routine in this script as a workaround for lack of `InsertNewlineAtEOF`. This currently affects 7 source files under the `test` directory relative to the current state of the PR (it seems like 12 files under the `src` directory have already been fixed relative to the target branch in prior revisions).

.clang-format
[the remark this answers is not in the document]
eramongodb: I'm fine with that as well. I think establishing consistent line spacing in all contexts is preferable to leaving in the 2 line separations.

etc/format-newlines.py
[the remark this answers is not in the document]
eramongodb: Resolving in favor of new comment.

.clang-format
```diff
@@ -1,37 +1,33 @@
...
+IndentCaseLabels: false
+IndentWidth: 4
+InsertBraces: true
+KeepEmptyLinesAtTheStartOfBlocks: false
+PointerAlignment: Right
+QualifierAlignment: Custom
+QualifierOrder: ["static", "inline", "const", "volatile", "type", "restrict"]
+SeparateDefinitionBlocks: Always
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
```
eramongodb, line 29:
```suggestion
SpaceBeforeAssignmentOperators: true
SpaceBeforeParens: Always
```
I still want to make sure we want to change from `SpaceBeforeParens: Always` (status quo) to `SpaceBeforeParens: ControlStatements`. Re-posting a comment here to invite further opinions, but I do not consider this suggestion a blocker.

eramongodb (approved):
Minor suggestions remaining; otherwise, LGTM. Improvements to formatting are greatly appreciated. Thank you for the work.

kkloberdanz (approved):

.clang-format
vector-of-bool: Yes, this change was made based on the poll

etc/format-all.sh
vector-of-bool: I'm going to postpone that, as we'll only have to wait a few weeks for LLVM 16, and that is a relatively minor formatting change.

PR #578, opened 2023-02-17
MONGOCRYPT-540 Define QE v2 token types and implemente token derivation

erwee:
This defines and adds derivation functions for the following v2 token types:
* `ServerTokenDerivationLevel1Token = HMAC(indexKey, 2)`
* `ServerDerivedFromDataToken = HMAC(ServerTokenDerivationLevel1Token, value)`

Only `ServerDerivedFromDataToken` will really be used in the v2 payloads. `ServerTokenDerivationLevel1Token` is just an intermediate type used for deriving the `ServerDerivedFromDataToken`.

### DIFF
--- a/src/mc-tokens-private.h
+++ b/src/mc-tokens-private.h
@@ -29,16 +29,19 @@
  * Integers are represented as uint64_t in little-endian.
  *
  * CollectionsLevel1Token = HMAC(RootKey, 1)
+ * ServerTokenDerivationLevel1Token = HMAC(RootKey, 2)  <- new in v2
  * ServerDataEncryptionLevel1Token = HMAC(RootKey, 3)
  *
  * EDCToken = HMAC(CollectionsLevel1Token, 1)
  * ESCToken = HMAC(CollectionsLevel1Token, 2)
- * ECCToken = HMAC(CollectionsLevel1Token, 3)
+ * ECCToken = HMAC(CollectionsLevel1Token, 3)  <- deprecated in v2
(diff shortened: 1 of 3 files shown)


src/mc-tokens-private.h
```diff
@@ -29,16 +29,19 @@
...
  * ESCToken = HMAC(CollectionsLevel1Token, 2)
- * ECCToken = HMAC(CollectionsLevel1Token, 3)
+ * ECCToken = HMAC(CollectionsLevel1Token, 3)  <- deprecated in v2
  * ECOCToken = HMAC(CollectionsLevel1Token, 4)
  *
  * EDCDerivedFromDataToken = HMAC(EDCToken, v)
  * ESCDerivedFromDataToken = HMAC(ESCToken, v)
- * ECCDerivedFromDataToken = HMAC(ECCToken, v)
+ * ECCDerivedFromDataToken = HMAC(ECCToken, v)  <- deprecated in v2
+ * ServerDerivedFromDataToken = HMAC(ServerDataEncryptionLevel1Token, v) <- new
```
kevinAlbs, line 43:
```suggestion
 * ServerDerivedFromDataToken = HMAC(ServerTokenDerivationLevel1Token, v) <- new
```

kevinAlbs (approved):

src/mc-tokens-private.h
erwee: fixed. thanks for catching that!

PR #582, opened 2023-02-23
MONGOCRYPT-537 Use AES-256-CBC/SHA-256 with FLE2v2

sgolemon:
Add FLEv2 ciphers as AES-256-CBC/SHA-256 and converts the one and only place we're actually using them with V2 types.


markbenvenuto (approved):

etc/fle2v2_generate_tests.py
```diff
@@ -0,0 +1,32 @@
+# fle2_aead_generate_tests.py is used to generate the test file: ./test/data/roundtrip/fle2v2-generated.json
```
kevinAlbs, line 1:
```suggestion
# fle2v2_aead_generate_tests.py is used to generate the test file: ./test/data/roundtrip/fle2v2-generated.json
```

etc/fle2_crypto.py
```diff
@@ -25,81 +26,92 @@ def __init__ (self, bytesIn):
...
-    assert (len(Ke) == ENCRYPTION_KEY_LENGTH)
-    assert (len(IV) == IV_LENGTH)
+    assert len(Ke) == ENCRYPTION_KEY_LENGTH
+    assert len(IV) == IV_LENGTH
+    assert (mode == 'CTR') or (mode == 'CBC')
+    modeObj = modes.CTR(IV) if mode == 'CTR' else modes.CBC(IV)
 
-    # S = AES-CTR.Enc(Ke, IV, M)
-    cipher = Cipher(algorithms.AES(Ke), modes.CTR(IV))
+    # S = AES-{mode}.Env(Ke, IV, M)
```
kevinAlbs, line 42:
```suggestion
    # S = AES-{mode}.Enc(Ke, IV, M)
```

etc/fle2_crypto.py
```diff
@@ -25,81 +26,92 @@ def __init__ (self, bytesIn):
...
-    cipher = Cipher(algorithms.AES(Ke), modes.CTR(IV))
-    encryptor = cipher.encryptor()
-    S = encryptor.update(M) + encryptor.finalize()
+def fle2v2_encrypt(M, Ke, IV, Km, AD):
+    """ AES-256-CBC/SHA-256 """
+    return _fle2_encrypt (IV, Ke, M, 'CBC', Km, AD)
 
-    # T = HMAC-SHA256(Km, AD || IV || S)
-    T = _hmacsha256 (Km, AD + IV + S)
+def _fle2_decypt (C, Ke, mode, Km = None, AD = None):
```
kevinAlbs, line 77:
```suggestion
def _fle2_decrypt (C, Ke, mode, Km = None, AD = None):
```

src/mongocrypt-crypto.c
```diff
@@ -1280,6 +1283,9 @@ DECLARE_ALGORITHM (FLE2AEAD, CTR, SHA_256)
 // FLE2 used with ESC/ECOC tokens: AES-256-CTR no HMAC
 DECLARE_ALGORITHM (FLE2, CTR, NONE)
 
+// FLE2 used with FLE2IndexedEncryptedValue: AES-256-CBC HMAC/SHA-256
+DECLARE_ALGORITHM (FLE2v2, CBC, SHA_256)
```
kevinAlbs, line 1287:
```suggestion
DECLARE_ALGORITHM (FLE2v2AEAD, CBC, SHA_256)
```

Since this algorithm includes HMAC.

kevinAlbs (approved):

etc/fle2_crypto.py
sgolemon: 👍

etc/fle2_crypto.py
sgolemon: Huh.  Technically we never run the decrypt function in here. Hope that never bits us in the keester...

etc/fle2v2_generate_tests.py
sgolemon: 👍

src/mongocrypt-crypto.c
sgolemon: I had a premonition this request would come.
Renamed and moved throughout.

sgolemon (commented):
Rebased to make reviews more sane and addressed comments.

src/mc-fle2-insert-update-payload-v2.c
```diff
@@ -322,16 +322,16 @@ mc_FLE2InsertUpdatePayloadV2_decrypt (_mongocrypt_crypto_t *crypto,
...
+   if (!fle2v2->do_decrypt (crypto,
+                            &iup->userKeyId,
+                            user_key,
+                            &ciphertext,
+                            &iup->plaintext,
+                            &bytes_written,
+                            status)) {
       return NULL;
    }
    return &iup->plaintext;
```
erwee, line 337: `iup->plaintext.len` needs to be set to `bytes_written` here, since `bytes_written` can now be smaller than the allocated size.

erwee (changes requested):

src/mc-fle2-insert-update-payload-v2.c
sgolemon: Gorram, excellent catch.

erwee (approved):

sgolemon:
Pushed as e656245b7ebc742df210c8156b9aac41bdd5d113

PR #583, opened 2023-03-01
feat(NODE-5076): add support for Azure KMS credential auto refresh

baileympearson:
## What's changing

- the credential providers have been broken in two separate modules, because the azure module contains a lot of code that doesn't need to live with the other credential refreshing.
- a generic HTTP get method has been added
- azure KMS credentials are automatically fetched if provided a `{}` in the kms providers
- a global credential cache has been added specifically for Azure KMS
- a new non-enumerable ~symbol~ property is exported from mongodb-client-encryption for use in driver prose tests


bindings/node/lib/errors.js
```diff
@@ -49,8 +49,27 @@ class MongoCryptCreateEncryptedCollectionError extends MongoCryptError {
   }
 }
 
+/**
+ * @class
+ * An error indicating that mongodb-client-encryption failed to auto-refresh Azure KMS credentials.
+ */
+class MongoCryptAzureKMSRequestError extends Error {
```
addaleax, line 56: Intentionally not extending from `MongoCryptError`?

baileympearson: No, good catch 👍

bindings/node/test/providers/credentialsProvider.test.js
```diff
@@ -0,0 +1,542 @@
...
+    afterEach(() => tokenCache.resetCache());
+    afterEach(() => sinon.restore());
+    context('credential caching', () => {
+      const cache = tokenCache;
+
+      beforeEach(() => {
+        cache.resetCache();
+      });
+
+      afterEach(() => sinon.restore());
```
nbbeeken, line 259: Doubling up on restore hooks

bindings/node/test/providers/credentialsProvider.test.js
```diff
@@ -0,0 +1,542 @@
...
+      });
+    });
+
+    context('error handling', () => {
+      afterEach(() => sinon.restore());
+      context('when the request times out', () => {
+        before(() => {
+          sinon
+            .stub(utils, 'get')
+            .callsFake(() => Promise.reject(new MongoCryptNetworkTimeoutError()));
```
nbbeeken, line 423: you could use `rejects` and just pass the error I think, and same for resolves

nbbeeken (changes requested):

bindings/node/lib/index.js
```diff
@@ -12,30 +12,47 @@ function loadDefaultModule() {
...
   MongoCryptError,
   MongoCryptCreateEncryptedCollectionError,
-  MongoCryptCreateDataKeyError
+  MongoCryptCreateDataKeyError,
+  MongoCryptAzureKMSRequestError,
+  MongoCryptNetworkTimeoutError
 } = require('./errors');
 
+const { fetchAzureKMSToken } = require('./providers');
+const AZURE_PROSE_TESTING_SYMBOL = Symbol.for('@@mdb.azureKMSRefreshProseTest');
```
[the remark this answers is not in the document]
baileympearson: Do you mean directly import from the node_modules?  It could, but that doesn't seem like good practice.

- that goes against all practice we currently have established in how we test mongodb-client-encryption in the driver
- the current implementation documents clearly and with tests the expected behavior to prevent breakages

Unless you feel strongly, I'm going to leave this as-is

bindings/node/lib/providers/azure.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param { {body: string, status: number }} response
+ * @returns { Promise<{ accessToken: string, expiresOnTimestamp: number } >}
+ */
+async function parseResponse(response) {
+  const { status, body: rawBody } = response;
+
+  /**
+   * @type { { access_token?: string, expires_in?: string} }
+   */
+  const body = (() => {
```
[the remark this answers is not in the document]
baileympearson: Not `const` specifically, but yeah.  I dislike the declare outside scope + assign in scope pattern

bindings/node/lib/providers/utils.js
```diff
@@ -0,0 +1,33 @@
...
+      .get(url, options, response => {
+        response.setEncoding('utf8');
+        let body = '';
+        response.on('data', chunk => (body += chunk));
+        response.on('end', () => {
+          clearTimeout(timeoutId);
+          resolve({ status: response.statusCode, body });
+        });
+      })
+      .on('error', error => reject(error))
```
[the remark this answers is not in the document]
baileympearson: done!

bindings/node/lib/errors.js
```diff
@@ -49,8 +49,27 @@ class MongoCryptCreateEncryptedCollectionError extends MongoCryptError {
...
+   * @param {string} message
+   * @param {object | undefined} body
+   */
+  constructor(message, body) {
+    super(message);
+    this.body = body;
+  }
+}
+
+class MongoCryptNetworkTimeoutError extends MongoCryptError {}
```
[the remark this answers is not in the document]
baileympearson: Done, let me know if you like the name better

bindings/node/lib/index.js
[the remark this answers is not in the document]
nbbeeken: I meant like this: `require('mongodb-client-encryption/lib/providers/index');` you don't need to put node_modules in the path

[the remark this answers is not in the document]
baileympearson: That doesn't change my mind. Do you feel strongly about this?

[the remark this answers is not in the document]
nbbeeken: I don't feel strongly, this just won't work in the future of our ecosystem, eventually ES modules will reach us, and we can't export symbols (IIUC), but that's a distant future if we wanna stick with this for now.

bindings/node/index.d.ts
```diff
@@ -179,7 +179,8 @@ export interface KMSProviders {
          * If present, an access token to authenticate with Azure.
          */
         accessToken: string;
-      };
```
nbbeeken, line 182: We have new errors exported as well: 
- MongoCryptAzureKMSRequestError,
- MongoCryptKMSRequestNetworkTimeoutError

nbbeeken (changes requested):

nbbeeken (approved):

bindings/node/index.d.ts
```diff
@@ -62,6 +62,16 @@ export class MongoCryptCreateDataKeyError extends MongoCryptError {
   cause: Error;
 }
 
+/**
+ * An error indicating that mongodb-client-encryption failed to auto-refresh Azure KMS credentials.
+ */
+class MongoCryptAzureKMSRequestError extends MongoCryptError {
```
nbbeeken, line 68: needs `export`

nbbeeken (changes requested):

bindings/node/lib/providers/azure.js
```diff
@@ -0,0 +1,155 @@
...
+ * @returns {Promise<{ accessToken: string, expiresOnTimestamp: number }>}
+ */
+async function fetchAzureKMSToken(options = {}) {
+  const { headers, url } = prepareRequest(options);
+  try {
+    const response = await utils.get(url, { headers });
+    return parseResponse(response);
+  } catch (error) {
+    if (error instanceof MongoCryptKMSRequestNetworkTimeoutError) {
+      throw new MongoCryptAzureKMSRequestError('Azure KMS request timed out after 10s');
```
dariakp, line 140: If the goal is just to prepend the timeout error with `"Azure KMS"`, would it be better to do something like `[Azure KMS] ${error.message}`? That way you don't have to maintain the "10s" info in multiple places if we ever decide to change the timeout.

Also, I'm wondering if it's better to put this transformation in a catch on the get `utils.get().catch( <transform error>)` just because this syntax suggests that the timeout error could be thrown from `parseResponse`, too, but that's not the case.

Feel free not to block the alpha release on this, but I recommend considering making both of the above changes for the final version.

dariakp (changes requested):

bindings/node/test/providers/credentialsProvider.test.js
```diff
@@ -0,0 +1,527 @@
...
+          });
+
+          it('throws a MongoCryptKMSRequestError', async () => {
+            const error = await loadCredentials({ azure: {} }).catch(e => e);
+            expect(error).to.be.instanceOf(MongoCryptAzureKMSRequestError);
+            expect(error).to.match(/Malformed JSON body in GET request/);
+          });
+        });
+
+        context('when the request has a body', () => {
```
dariakp, line 441:
```suggestion
        context('when the request has a json body', () => {
```

bindings/node/test/providers/credentialsProvider.test.js
```diff
@@ -0,0 +1,527 @@
...
+        });
+
+        it('throws a MongoCryptKMSRequestError', async () => {
+          const error = await loadCredentials({ azure: {} }).catch(e => e);
+          expect(error).to.be.instanceOf(MongoCryptAzureKMSRequestError);
+        });
+      });
+
+      context('when the request returns a non-200 error', () => {
+        context('when the request has no body', () => {
```
dariakp, line 429: might also add a case for when a request has a non-json body

bindings/node/test/providers/credentialsProvider.test.js
```diff
@@ -0,0 +1,527 @@
...
+
+          it('attaches the body to the error', async () => {
+            const error = await loadCredentials({ azure: {} }).catch(e => e);
+            expect(error).to.have.property('body').to.deep.equal({ error: 'something went wrong' });
+          });
+        });
+      });
+
+      context('when the request returns a 200 response', () => {
+        context('when the request has no body', () => {
```
dariakp, line 461: might want to add a case for when the body is defined but not json

bindings/node/test/providers/credentialsProvider.test.js
```diff
@@ -0,0 +1,527 @@
...
+          it('throws a MongoCryptKMSRequestError', async () => {
+            const error = await loadCredentials({ azure: {} }).catch(e => e);
+            expect(error).to.be.instanceOf(MongoCryptAzureKMSRequestError);
+            expect(error).to.match(/Malformed JSON body in GET request/);
+          });
+        });
+
+        context('when the body has no access_token', () => {
+          beforeEach(() => {
+            sinon.stub(utils, 'get').resolves({ status: 200, body: '{ }' });
```
dariakp, line 475: as a testing best practice, it's better to isolate test conditions by ensuring that the response is valid except for the thing being tested, so I would include a valid `explires_in` here

dariakp (changes requested):

bindings/node/test/providers/credentialsProvider.test.js
baileympearson: done!

bindings/node/test/providers/credentialsProvider.test.js
baileympearson: added!

bindings/node/test/providers/credentialsProvider.test.js
baileympearson: done!

bindings/node/test/providers/credentialsProvider.test.js
baileympearson: done!

bindings/node/lib/providers/azure.js
baileympearson: It also changes the error type to follow our standard <KMS provider><error name> error pattern.  

But I addressed both points.  I moved the error handling into a `catch` (I like this suggestion) and interpolate `error.message` so we're not duplicating the error message.

bindings/node/test/types/index.test-d.ts
```diff
@@ -51,7 +51,7 @@ expectAssignable<RequiredCreateEncryptedCollectionSettings>({
   expectAssignable<KMSProviders['azure']>({ tenantId: 'a', clientId: 'a', clientSecret: 'a', identityPlatformEndpoint: '' });
   expectAssignable<KMSProviders['azure']>({ accessToken: 'a' });
   // automatic TODO(NODE-4537): update azure types to accept automatic credentials
```
nbbeeken, line 53: I think this is the ticket to remove this TODO, correct?

nbbeeken (changes requested):

nbbeeken (approved):

PR #589, opened 2023-03-07
MONGOCRYPT-550 Add support for FLE2UnindexedEncryptedValueV2 encrypt/decrypt

erwee:
In this patch:
* most of the parse/decrypt/encrypt code in `src/mc-fle2-payload-uev.c` have been moved to a new src file `mc-fle2-payload-uev-common.c`
* added `mc-fle2-payload-uev-v2-private.h` for the v2 UEV parse/encrypt/decrypt interface. (Note that unlike the other "v2" header files, the "v2" precedes "private" in the filename.. I think the "private" should always precede the .h extension, and I'm all for renaming the existing header files that have them in the other order.)
* `test/test-mc-fle2-payload-uev-v2.c` unit tests the v2 interface, and is almost an exact carbon copy of `test-mc-fle2-payload-uev.c`.

Integrating the v2 stuff into `mongocrypt-marking.c` and `mongocrypt-ctx-decrypt.c` will be done in MONGOCRYPT-551, after MONGOCRYPT-542 lands.

### DIFF
--- a/src/mc-fle2-payload-uev-common.c
+++ b/src/mc-fle2-payload-uev-common.c
@@ -1,0 +1,222 @@
+/*
+ * Copyright 2023-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mc-fle2-payload-uev-common-private.h"
+#include "mongocrypt-private.h"
+
+bool
+_mc_FLE2UnindexedEncryptedValueCommon_parse (const _mongocrypt_buffer_t *buf,
+                                             uint8_t *fle_blob_subtype,
+                                             uint8_t *original_bson_type,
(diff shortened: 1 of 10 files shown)


src/mc-fle2-payload-uev-common.c
```diff
@@ -0,0 +1,222 @@
...
+                                             _mongocrypt_buffer_t *ciphertext,
+                                             mongocrypt_status_t *status)
+{
+   BSON_ASSERT_PARAM (buf);
+   BSON_ASSERT_PARAM (fle_blob_subtype);
+   BSON_ASSERT_PARAM (original_bson_type);
+   BSON_ASSERT_PARAM (key_uuid);
+   BSON_ASSERT_PARAM (ciphertext);
+
+   uint32_t offset = 0;
```
markbenvenuto, line 34: You can use the new src/mc-reader-private.h API to reduce all the copy/paste bounds checking. See mc_FLE2IndexedEqualityEncryptedValueTokens_init_from_buffer

markbenvenuto (changes requested):

src/mc-fle2-payload-uev-common.c
erwee: very nice! done.

kevinAlbs (approved):
> added mc-fle2-payload-uev-v2-private.h for the v2 UEV parse/encrypt/decrypt interface. (Note that unlike the other "v2" header files, the "v2" precedes "private" in the filename.. I think the "private" should always precede the .h extension, and I'm all for renaming the existing header files that have them in the other order.)

I have no preference. The `private` is only informational and does not impact [the installed headers](https://github.com/mongodb/libmongocrypt/blob/2ec9c3c4ff94f76bb51a5b4cc2fe9190457d6541/CMakeLists.txt#L620-L623).

The added testing of `FLE2UnindexedEncryptedValue` is appreciated.

PR #592, opened 2023-03-08
MONGOCRYPT-552 Add derivators for ServerZerosEncryptionToken and ServerCountAndContentionFactorEncryptionToken

sgolemon:
Actual work in `src` is small thanks to all the macroing.

The test refactor is similar to what I did in `test-mongocrypt-crypto`.  Main reason for breaking it out was to add a second complete test case coming from the SERVER codebase (server.json) so that we're testing that we get all the same outputs.  Existing test vectors are in `mc.json` with the cases for the new derived values added.


erwee (approved):
P.S. the `ServerCountAndContentionFactorEncryptionToken` and `ServerZerosEncryptionToken` are never used by libmongocrypt. (hence why I didn't see the need to include them in MONGOCRYPT-540). For decrypting the IEV, it only needs to decrypt the "server-encrypted client ciphertext" using the `ServerDataEncryptionLevel1Token`, and ignore "metadata block(s)" that come after it.

sgolemon:
Oh, I was thinking we'd want to validate the metadata to guard against tampering.  Oh well, as you say it doesn't hurt to have the functionality available.

src/mc-tokens-private.h
```diff
@@ -29,23 +29,32 @@
...
  * EDCDerivedFromDataTokenAndCounter = HMAC(EDCDerivedFromDataToken, u)
  * ESCDerivedFromDataTokenAndCounter = HMAC(ESCDerivedFromDataToken, u)
  * ECCDerivedFromDataTokenAndCounter = HMAC(ECCDerivedFromDataToken, u)
+ *
+ * Note: ECC related tokens are used in FLE2v1 only.
+ * ----------------------------------------------------------------------------
+ * Added in FLE2v2:
+ *
+ * ServerTokenDerivationLevel1Token = HMAC(RootKey, 2)
+ * ServerDataEncryptionLevel1Token = HMAC(RootKey, 3)
```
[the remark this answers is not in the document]
sgolemon: huh. The 1/3 gap threw me off there, but you are correct.

src/mc-tokens.c
```diff
@@ -162,3 +145,38 @@ DEF_TOKEN_TYPE (mc_ECCDerivedFromDataTokenAndCounter,
...
+
+/* FLE2v2 */
+
+DEF_TOKEN_TYPE (mc_ServerTokenDerivationLevel1Token,
+                const _mongocrypt_buffer_t *RootKey)
+IMPL_TOKEN_NEW_CONST (mc_ServerTokenDerivationLevel1Token, RootKey, 2)
+
+DEF_TOKEN_TYPE (mc_ServerDataEncryptionLevel1Token,
+                const _mongocrypt_buffer_t *RootKey)
+IMPL_TOKEN_NEW_CONST (mc_ServerDataEncryptionLevel1Token, RootKey, 3)
```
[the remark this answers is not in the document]
sgolemon: Done

src/mc-tokens-private.h
```diff
@@ -29,23 +29,31 @@
  * Integers are represented as uint64_t in little-endian.
  *
  * CollectionsLevel1Token = HMAC(RootKey, 1)
- * ServerTokenDerivationLevel1Token = HMAC(RootKey, 2)  <- new in v2
  * ServerDataEncryptionLevel1Token = HMAC(RootKey, 3)
  *
  * EDCToken = HMAC(CollectionsLevel1Token, 1)
  * ESCToken = HMAC(CollectionsLevel1Token, 2)
- * ECCToken = HMAC(CollectionsLevel1Token, 3)  <- deprecated in v2
```
markbenvenuto, line 37: nit - I feel like we should keep these deprecation notes

sgolemon: I did add the note about ECC related tokens being used with FLE2v1, but you're right that it doesn't explicitly call it a deprecation...  Given these are private APIs I'm not too worried about that semantic distinction, but I'll clarify it since it costs nothing...

src/mc-tokens.c
```diff
@@ -133,14 +125,9 @@ DEF_TOKEN_TYPE (mc_ECCDerivedFromDataToken,
...
 
-DEF_TOKEN_TYPE (
-   mc_ServerDerivedFromDataToken,
-   const mc_ServerTokenDerivationLevel1Token_t *ServerTokenDerivationToken,
-   const _mongocrypt_buffer_t *v)
-IMPL_TOKEN_NEW (
-   mc_ServerDerivedFromDataToken,
-   mc_ServerTokenDerivationLevel1Token_get (ServerTokenDerivationToken),
-   v)
+DEF_TOKEN_TYPE (mc_ServerDataEncryptionLevel1Token,
```
kevinAlbs, line 128: Suggest reverting move to `mc_ServerDataEncryptionLevel1Token` since it is not new to v2.

kevinAlbs (approved):
Test refactor is much appreciated.

PR #601, opened 2023-03-17
MONGOCRYPT-538

sgolemon:
This is notably missing an integration test for IUP-V2, but I'm waiting on that 'cause I *think* @shreyaskalyan is working on unit tests for it.  Everything else seems to work smoothly though, so it's worth starting review since this is... growing.

### DIFF
--- a/src/mongocrypt-ctx-decrypt.c
+++ b/src/mongocrypt-ctx-decrypt.c
@@ -72,7 +72,7 @@ _replace_FLE2IndexedEncryptedValue_with_plaintext (void *ctx,
    CHECK_AND_RETURN (K_KeyId);
    CHECK_AND_RETURN_KB_STATUS (
       _mongocrypt_key_broker_decrypted_key_by_id (kb, K_KeyId, &K_Key));
-   CHECK_AND_RETURN (mc_FLE2IndexedEqualityEncryptedValue_add_K_Key (
+   CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValue_add_K_Key (
       kb->crypt->crypto, iev, &K_Key, status));
    const _mongocrypt_buffer_t *clientValue =
       mc_FLE2IndexedEncryptedValue_get_ClientValue (iev, status);
--- a/src/mongocrypt-ctx-decrypt.c
+++ b/src/mongocrypt-ctx-decrypt.c
@@ -59,25 +59,33 @@ _replace_FLE2IndexedEncryptedValue_with_plaintext (void *ctx,
    BSON_ASSERT_PARAM (in);
    BSON_ASSERT_PARAM (out);
 
+   // Parse the IEV payload to get S_KeyId.
    CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValue_parse (iev, in, status));
    const _mongocrypt_buffer_t *S_KeyId =
       mc_FLE2IndexedEncryptedValue_get_S_KeyId (iev, status);
    CHECK_AND_RETURN (S_KeyId);
+
    CHECK_AND_RETURN_KB_STATUS (
       _mongocrypt_key_broker_decrypted_key_by_id (kb, S_KeyId, &S_Key));
+
+   // Use S_Key to decrypt envelope and get to K_KeyId.
    CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValue_add_S_Key (
       kb->crypt->crypto, iev, &S_Key, status));
    const _mongocrypt_buffer_t *K_KeyId =
       mc_FLE2IndexedEncryptedValue_get_K_KeyId (iev, status);
    CHECK_AND_RETURN (K_KeyId);
+
    CHECK_AND_RETURN_KB_STATUS (
       _mongocrypt_key_broker_decrypted_key_by_id (kb, K_KeyId, &K_Key));
+
+   // Decrypt the actual data value using K_Key.
    CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValue_add_K_Key (
       kb->crypt->crypto, iev, &K_Key, status));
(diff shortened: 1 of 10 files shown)


markbenvenuto (approved):

src/mongocrypt-ctx-decrypt.c
```diff
@@ -39,68 +59,92 @@ _replace_FLE2IndexedEncryptedValue_with_plaintext (void *ctx,
...
+   CHECK_AND_RETURN_KB_STATUS (
+      _mongocrypt_key_broker_decrypted_key_by_id (kb, S_KeyId, &S_Key));
+   CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValue_add_S_Key (
+      kb->crypt->crypto, iev, &S_Key, status));
+   const _mongocrypt_buffer_t *K_KeyId =
+      mc_FLE2IndexedEncryptedValue_get_K_KeyId (iev, status);
+   CHECK_AND_RETURN (K_KeyId);
+   CHECK_AND_RETURN_KB_STATUS (
+      _mongocrypt_key_broker_decrypted_key_by_id (kb, K_KeyId, &K_Key));
+   CHECK_AND_RETURN (mc_FLE2IndexedEqualityEncryptedValue_add_K_Key (
```
kevinAlbs, line 75: Not related to these changes. Suggest renaming `mc_FLE2IndexedEqualityEncryptedValue_add_K_Key` to `mc_FLE2IndexedEncryptedValue_add_K_Key` since it is used for both Equality and Range payloads.

kevinAlbs (approved):
Nice. LGTM. `_test_ctx_wrap_and_feed_key` looks helpful.

Consider adding tests to `_test_encrypt_fle2_explicit` for explicit encryption. I expect explicit encryption to "just work". Explicit encryption produces the placeholders. And the placeholders have not changed in FLE2v2.

src/mongocrypt-ctx-decrypt.c
```diff
@@ -39,68 +59,92 @@ _replace_FLE2IndexedEncryptedValue_with_plaintext (void *ctx,
...
+   mongocrypt_status_t *status)
+{
+   bool ret = false;
+   _mongocrypt_key_broker_t *kb = ctx;
+   mc_FLE2IndexedEncryptedValueV2_t *iev =
+      mc_FLE2IndexedEncryptedValueV2_new ();
+   _mongocrypt_buffer_t S_Key = {0};
+   _mongocrypt_buffer_t K_Key = {0};
 
-   /* Decrypt ClientEncryptedValue. */
```
erwee, line 74: Please add back the `Decrypt X` comments before the `add_S/K_Key` calls. They help a lot with readability.

src/mongocrypt-ctx-decrypt.c
```diff
@@ -39,68 +59,92 @@ _replace_FLE2IndexedEncryptedValue_with_plaintext (void *ctx,
...
+   BSON_ASSERT_PARAM (out);
 
+   CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValueV2_parse (iev, in, status));
+   const _mongocrypt_buffer_t *S_KeyId =
+      mc_FLE2IndexedEncryptedValueV2_get_S_KeyId (iev, status);
+   CHECK_AND_RETURN (S_KeyId);
+   CHECK_AND_RETURN_KB_STATUS (
+      _mongocrypt_key_broker_decrypted_key_by_id (kb, S_KeyId, &S_Key));
+   CHECK_AND_RETURN (mc_FLE2IndexedEncryptedValueV2_add_S_Key (
+      kb->crypt->crypto, iev, &S_Key, status));
```
erwee, line 123: I think this will cause an abort if `status` is NULL. We made `status` mandatory in the `add_S/K_Key` functions, but here it's optional.

erwee (changes requested):

shreyaskalyan (approved):

src/mongocrypt-ctx-decrypt.c
sgolemon: I did have the same thought, actually.  Happy to include it.

src/mongocrypt-ctx-decrypt.c
sgolemon: Fair, done.

src/mongocrypt-ctx-decrypt.c
sgolemon: Good catch. Adjusted to create a temporary status as-needed.

src/mongocrypt-ctx-decrypt.c
```diff
@@ -455,6 +618,32 @@ _collect_K_KeyID_from_FLE2IndexedEncryptedValue (void *ctx,
...
+         ctx, in, status);
+
+   // FLE2v1
+   case MC_SUBTYPE_FLE2IndexedEqualityEncryptedValue:
+   case MC_SUBTYPE_FLE2IndexedRangeEncryptedValue:
+      return _collect_K_KeyID_from_FLE2IndexedEncryptedValue (ctx, in, status);
+
+   default:
+      // Ignore other types.
+      return true;
```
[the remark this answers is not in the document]
sgolemon: No, because we're traversing over all fields and fully expect to run into UEV or IUP fields which don't have K_KeyId in them.

sgolemon (commented):
Addressed comments, then rebased now that MC-531 has landed.

erwee (approved):

src/mongocrypt-ctx-decrypt.c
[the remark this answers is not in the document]
shreyaskalyan: Gotcha

sgolemon:
Merged as 4a544db9274cb84024cfdd718c1afd833f93ece0

PR #603, opened 2023-03-21
MONGOCRYPT-563 add CryptographicUsageMask to Register request

kevinAlbs:
# Summary
- Add `CryptographicUsageMask` attribute to KMIP `Register` request
- Fix error handling of calls to `kms_kmip_request.*new`

Changes were verified with this [Evergreen patch build](https://spruce.mongodb.com/version/6419b6bbe3c331f96531c14f/) of the C driver using this branch of libmongocrypt.

# Background & Motivation

It was [reported on slack](https://mongodb.slack.com/archives/CDM9QEGUT/p1679061262507339) that versions 1.12 and 1.13 of HashiCorp Vault KMIP return an error on the KMIP Register request:
> 
> Error message: Caused by: com.mongodb.crypt.capi.MongoCryptException: Error getting UniqueIdentifer from KMIP Register response: KMIP response error. Result Status (1): Operation Failed. Result Reason (4): Invalid Message. Result Message: result reason: ResultReasonInvalidMessage; additional message: attribute Cryptographic Usage Mask is missing
> 

The addition of the `CryptographicUsageMask` attribute does not fix the error returned by Vault 1.13 reported in MONGOCRYPT-563. The error is no longer returned in development versions of Vault.

The motivation of this change is to improve conformance to the KMIP specification requirement that the `CryptographicUsageMask` attribute be present.

[4.3 Register](http://docs.oasis-open.org/kmip/spec/v1.0/os/kmip-spec-1.0-os.html#_Toc262581211) lists the "Cryptographic Usage Mask" attribute as REQUIRED.

The "Cryptographic Usage Mask" attribute included is not included in the [Register request for the SecretData object created by libmongocrypt](https://github.com/mongodb/libmongocrypt/blob/0caa1d3249d85050d0e3422d602e630aca574e0b/kms-message/src/kms*kmip*request.c#L41-L70).

[3.14 Cryptographic Usage Mask](http://docs.oasis-open.org/kmip/spec/v1.0/os/kmip-spec-1.0-os.html#_Toc262581188) lists "Cryptographic Usage Mask" in "When implicitly set" for the "Register" operation. [3 Attributes](http://docs.oasis-open.org/kmip/spec/v1.0/os/kmip-spec-1.0-os.html#_Toc262581172) defines "When implicitly set" as "Which operations MAY cause this attribute to be set even if the attribute is not specified in the operation request itself?". I interpret this to mean: the server may not implicitly set the `CryptographicUsageMask`, but it may not. If the server does not implicitly set `CryptographicUsageMask`, `CryptographicUsageMask` is required in the `Request`.

XML comments in the test data was generated with the [kmip_dump](https://github.com/kevinAlbs/c-bootstrap/tree/master/tools/kmip_dump) utility.


kkloberdanz (approved):

kms-message/src/kms_kmip_request.c
```diff
@@ -101,15 +106,29 @@ kms_kmip_request_register_secretdata_new (void *reserved,
...
+                                cryptographicUsageMaskStr,
+                                strlen (cryptographicUsageMaskStr));
+      // Use 0 because the Secret Data object is not used in cryptographic
+      // operations on the KMIP server.
+      kmip_writer_write_integer (writer, KMIP_TAG_AttributeValue, 0);
+      kmip_writer_close_struct (writer);
+   }
    kmip_writer_close_struct (writer); /* KMIP_TAG_TemplateAttribute */
    kmip_writer_begin_struct (writer, KMIP_TAG_SecretData);
    /* 0x01 = Password */
```
markbenvenuto, line 124: Nit - 0x2 = Seed

markbenvenuto (approved):

kms-message/src/kms_kmip_request.c
kevinAlbs: Fixed.

PR #605, opened 2023-03-28
Add new merged ref to blame-ignore

vector-of-bool:


.git-blame-ignore-revs
```diff
@@ -1,2 +1,3 @@
 # Mass reformat:
 24cb8cc878006f3ebf94884102e4e95e2529ac49
```
kevinAlbs, line 2:
```suggestion
```

Not needed? `git show 24cb8cc878006f3ebf94884102e4e95e2529ac49` shows `fatal: bad object 24cb8cc878006f3ebf94884102e4e95e2529ac49`

kevinAlbs (approved):
LGTM with possible removal.

.git-blame-ignore-revs
vector-of-bool: It's in my local copy, but only on a separate branch, so yeah, this can be dropped

PR #607, opened 2023-04-04
test(NODE-5168): fix flaky stubbed key material test

durran:
Fixes the stubbed key material test - no need to actually test the exact encrypted bytes.

### DIFF
--- a/bindings/node/test/clientEncryption.test.js
+++ b/bindings/node/test/clientEncryption.test.js
@@ -676,10 +676,6 @@ describe('ClientEncryption', function () {
         .then(encrypted => {
           expect(encrypted._bsontype).to.equal('Binary');
           expect(encrypted.sub_type).to.equal(6);
-          expect(encrypted.toString('base64')).to.equal(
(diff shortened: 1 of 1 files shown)


bindings/node/test/clientEncryption.test.js
```diff
@@ -676,10 +676,6 @@ describe('ClientEncryption', function () {
         .then(encrypted => {
           expect(encrypted._bsontype).to.equal('Binary');
           expect(encrypted.sub_type).to.equal(6);
-          expect(encrypted.toString('base64')).to.equal(
-            'BqvN76sSNJh2EjQSNFZ4kBICTQaVZPWgXp41I7mPV1rLFTtw1tXzjcdSEyxpKKqujlko5TeizkB9hHQ009dVY1+fgIiDcefh+eQrm3CkhQ=='
```
durran, line 680: This passed locally for me but fails on linux on evg with a different value for the encrypted bytes. However, it's probably not important to test for this as long as it gets encrypted and the next assertion checking for the correct decrypted value passes.

nbbeeken (approved):

PR #618, opened 2023-04-14
do not build tests if BUILD_TESTING is disabled

kevinAlbs:
# Summary

- Do not build test executables if `BUILD_TESTING` is disabled

# Background & Motivation

There is a reported error of installing libmongocrypt with Homebrew:
https://mongodb.slack.com/archives/CDM9QEGUT/p1681457334220079


kevinAlbs:
Not needed. Disregard.

PR #619, opened 2023-04-14
Fix build for MacOS Ventura

kkloberdanz:
MONGOCRYPT-571

Fix build for MacOS Ventura
Patch missing headers for Intel DFP


vector-of-bool (approved):

kevinAlbs (approved):
LGTM. Thank you for the quick fix!

PR #623, opened 2023-04-20
Define a mongocrypt::platform target for platform requirements

vector-of-bool:
`mongocrypt::platform` is an `INTERFACE` library that defines usage requirements related to platform features and dependencies on system libraries. When a library links to a libmongocrypt library, it will transitively link to `mongocrypt::platform`, which will propagate usage requirements required for consuming libmongocrypt (e.g. runtime support libraries)

### DIFF
--- a/cmake/mongocrypt-config.cmake
+++ b/cmake/mongocrypt-config.cmake
@@ -1,6 +1,25 @@
 include(CMakeFindDependencyMacro)
 find_dependency(kms_message 0.0.1)
+find_dependency(Threads)
+
 include("${CMAKE_CURRENT_LIST_DIR}/mongocrypt_targets.cmake")
+
+# Link for dlopen():
+set_property(TARGET mongo::mongocrypt::platform APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${CMAKE_DL_LIBS})
+
+# Link for special math functions:
+if (NOT APPLE)
+    find_library (_MONGOCRYPT_M_LIBRARY m)
+    if (_MONGOCRYPT_M_LIBRARY)
+        set_property(TARGET mongo::mongocrypt::platform APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${_MONGOCRYPT_M_LIBRARY}")
+    endif ()
+endif ()
+
+# Special runtime:
+find_library (_MONGOCRYPT_RT_LIBRARY rt)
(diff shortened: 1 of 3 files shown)


cmake/mongocrypt-config.cmake
```diff
@@ -40,3 +40,22 @@ if (_using_system_intel_dfp)
...
+find_dependency(Threads)
+
+# Link for dlopen():
+set_property(TARGET mongo::mongocrypt::platform APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${CMAKE_DL_LIBS})
+
+# Link for special math functions:
+if (NOT APPLE)
+    find_library (_MONGOCRYPT_M_LIBRARY m)
+    if (_MONGOCRYPT_M_LIBRARY)
+        set_property(TARGET mongo::mongocrypt::platform APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${_MONGOCRYPT_M_LIBRARY}")
```
rcsanchez97, line 53:
```suggestion
        set_property (TARGET mongo::mongocrypt::platform APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${_MONGOCRYPT_M_LIBRARY}")
```

rcsanchez97 (approved):
LGTM, with some nit-picky formatting suggestions.

cmake/mongocrypt-config.cmake
```diff
@@ -40,3 +40,24 @@ if (_using_system_intel_dfp)
       PROPERTY IMPORTED_LOCATION "${_MONGOCRYPT_SYSTEM_INTEL_DFP_STATIC}"
       )
 endif ()
+
+find_dependency(Threads)
+
+# Link for dlopen():
+set_property (TARGET mongo::mongocrypt::platform APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${CMAKE_DL_LIBS})
```
kevinAlbs, line 47: Is the `mongo::mongocrypt::platform` target expected to be used outside of `libmongocrypt`?

kevinAlbs (approved):
LGTM with a question intended for my understanding.

cmake/mongocrypt-config.cmake
vector-of-bool: Not directly, but it will appear as a transitive usage requirement on libmongocrypt consumers.

PR #624, opened 2023-04-21
MONGOCRYPT-570 require `min` and `max` for `int`, `long`, and `date`

kevinAlbs:
# Summary

- require min and max for int, long, and date

Verified with a patch build of the C driver using this branch of libmongocrypt: https://spruce.mongodb.com/version/6442a20ae3c3313e5f3dc97e/

# Background & Motivation

Previous motivation for making `min` and `max` optional was to match server behavior. This is noted in https://github.com/mongodb/libmongocrypt/pull/497

> Making `min` and `max` optional was motivated by matching [server behavior](https://github.com/mongodb/mongo/commit/0d9455e8ee5ed764aee6a514d789ae1b5ed0af03#diff-21fdaf455f52e97caa37093646d05b6fc182795e1152ec7d370ae98e14a7ee1cR100) for `double`/`decimal128`. For double and decimal128, min/max/precision must all be set, or all be unset.

This does not match the server behavior for `int`, `long`, and `date`. The server requires `min` and `max` be specified. These [tests with mongosh](https://gist.github.com/kevinAlbs/bbd2a0039d2cf06f19b5799877e35e73) show server behavior.


vector-of-bool (approved):
LGTM

PR #626, opened 2023-04-24
MONGOCRYPT-567 add QEv2 explicit encryption tests

kevinAlbs:
# Summary
- add QEv2 explicit encryption tests

# Background & Motivation

To add test coverage for QEv2. The QEv1 tests are planned for removal in MONGOCRYPT-572.


vector-of-bool (approved):
LGTM

PR #629, opened 2023-04-25
Improve bson_t initialization behavior

kkloberdanz:
MONGOCRYPT-558
MONGOCRYPT-559
MONGOCRYPT-560

This addresses the issues brought up in the above tickets. We will not be removing calls to `BSON_INITIALIZER`, but we are addressing the other issues uncovered in these tickets.

### DIFF
--- a/src/mc-fle2-rfds-private.h
+++ b/src/mc-fle2-rfds-private.h
@@ -48,6 +48,8 @@ bool mc_FLE2RangeFindDriverSpec_parse(mc_FLE2RangeFindDriverSpec_t *spec,
 
 // mc_FLE2RangeFindDriverSpec_to_placeholders creates a new document with
 // placeholders to encrypt.
+//
+// ACHTUNG: `out` MUST be initialized!
 bool mc_FLE2RangeFindDriverSpec_to_placeholders(mc_FLE2RangeFindDriverSpec_t *spec,
                                                 const mc_RangeOpts_t *range_opts,
                                                 int64_t maxContentionCounter,
--- a/src/mc-fle2-rfds-private.h
+++ b/src/mc-fle2-rfds-private.h
@@ -49,7 +49,7 @@ bool mc_FLE2RangeFindDriverSpec_parse(mc_FLE2RangeFindDriverSpec_t *spec,
 // mc_FLE2RangeFindDriverSpec_to_placeholders creates a new document with
 // placeholders to encrypt.
 //
-// ACHTUNG: `out` MUST be initialized!
+// `out` must be initialized by caller.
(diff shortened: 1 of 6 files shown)


src/mc-fle2-rfds-private.h
```diff
@@ -48,6 +48,8 @@ bool mc_FLE2RangeFindDriverSpec_parse(mc_FLE2RangeFindDriverSpec_t *spec,
 
 // mc_FLE2RangeFindDriverSpec_to_placeholders creates a new document with
 // placeholders to encrypt.
+//
+// ACHTUNG: `out` MUST be initialized!
```
kevinAlbs, line 52:
```suggestion
// `out` must be initialized by caller.
```

Suggest removing the warning note, since this may be a preferable pattern, rather than a "gotcha".

I prefer requiring callers initialize the `bson_t`. If the callee initializes `bson_t` it overwrites the flags of the `bson_t`. If the caller passes a heap allocated `bson_t` (e.g. through `bson_new()`) it results in a leak of the `bson_t` struct. Here is an example:

```c
// A common pattern in the C driver and libmongocrypt is to expect an uninitialized `bson_t` as an output parameter.
// This prevents callers from heap allocating `bson_t`.

#include <mongoc/mongoc.h>

void callee_initializes(bson_t *out)
{
    bson_init(out);
    BSON_APPEND_UTF8(out, "foo", "bar");
}

void caller_initializes(bson_t *out)
{
    BSON_APPEND_UTF8(out, "foo", "bar");
}

int main()
{
    mongoc_init();

    // If callee initializes, passing a heap allocated bson_t results in leak.
    {
        bson_t *heap_allocated = bson_new();

        callee_initializes(heap_allocated); // Reinitializes the bson_t as stack allocated.
        bson_destroy(heap_allocated);       // Does not destroy bson_t struct!
    }

    // If caller initializes, passing a heap allocated bson_t is OK.
    {
        bson_t *heap_allocated = bson_new();

        caller_initializes(heap_allocated);
        bson_destroy(heap_allocated); // Destroys the bson_t struct.
    }

    mongoc_cleanup();
}
```

[SWIFT-824](https://jira.mongodb.org/browse/SWIFT-824) is a bug in the Swift driver resulting from this API.

kevinAlbs (approved):
LGTM with a suggested comment change.

I suggest updating the PR title to be "Improve `bson_t` initialization behavior" or similar. The reported leaks require compiling the C driver with `BSON_MEMCHECK`. Initializing a `bson_t` struct twice (without appending in between) does not result in a leak.

vector-of-bool (approved):
LGTM

PR #631, opened 2023-04-26
cleanup doc before exiting on error

kkloberdanz:
MONGOCRYPT-557


src/mongocrypt-ctx-rewrap-many-datakey.c
```diff
@@ -36,6 +36,9 @@ static bool _finalize(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out) {
             bson_t elem = BSON_INITIALIZER;
 
             if (!mongocrypt_ctx_finalize(iter->dkctx, &bin)) {
+                bson_append_array_end(&doc, &array);
```
kevinAlbs, line 39:
```suggestion
                BSON_ASSERT(bson_append_array_end(&doc, &array));
```

To abort on unexpected error.

kevinAlbs (approved):
LGTM with addition of `BSON_ASSERT`

PR #634, opened 2023-04-28
fix: use CMAKE_INSTALL_FULL_* variables for pkg-config files

K900:
This is more correct as ${CMAKE_INSTALL_LIBDIR} may be set to an absolute path, and ${prefix} may not be set at all.


rcsanchez97:
Thank you for your contribution.

> This is more correct as ${CMAKE_INSTALL_LIBDIR} may be set to an absolute path, and ${prefix} may not be set at all.

Can you please cite sources for these claims?

The [documentation for the GNUInstallDirs module](https://cmake.org/cmake/help/v3.12/module/GNUInstallDirs.html) states the following:

```
CMAKE_INSTALL_<dir>

    Destination for files of a given type. This value may be passed to the DESTINATION options of [install()](https://cmake.org/cmake/help/v3.12/command/install.html#command:install) commands for the corresponding file type.
```

and

```
LIBDIR

    object code libraries (lib or lib64 or lib/<multiarch-tuple> on Debian)
```

Nothing about this suggests that `CMAKE_INSTALL_LIBDIR` would be set to an aboslute path.  I suppose that the user could specify an absolute path on the command line via a `-D` option.  However,  that would certainly not be standard and I would go so far as to say that if a user does that, it is done with knowledge and with acceptance of the consequences of overriding a value which isn't meant to be overridden with an absolute path.

Additionally, your statement that `${prefix}` may not be set at all does not make sense.  The input file for the pkg-config scripts look like this:

```
Name: ${PROJECT_NAME}
Description: ${PROJECT_DESCRIPTION}
Version: ${PROJECT_VERSION}
Requires: ${PKG_CONFIG_STATIC_REQUIRES}
prefix=${CMAKE_INSTALL_PREFIX}
includedir=${PKG_CONFIG_INCLUDEDIR}
libdir=${PKG_CONFIG_LIBDIR}
Libs: ${PKG_CONFIG_STATIC_LIBS}
Cflags: ${PKG_CONFIG_STATIC_CFLAGS}
```

You will note that the `prefix=` line is set from the value of `${CMAKE_INSTALL_PREFIX}`.  According to [the CMake documentation](https://cmake.org/cmake/help/v3.12/variable/CMAKE_INSTALL_PREFIX.html):

```
If make install is invoked or INSTALL is built, this directory is prepended onto all install directories. This variable defaults to /usr/local on UNIX and c:/Program Files/${PROJECT_NAME} on Windows. See [CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT](https://cmake.org/cmake/help/v3.12/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html#variable:CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT) for how a project might choose its own default.
```

Based on that, I do not see any possible scenario where the value of `${prefix}` does not get set.

Additionally, your proposed change subverts one of the features of the pkg-config.  By specifying the `prefix=` value in the `.pc` file, it is possible to relocate the entire install tree elsewhere on the filesystem and still have everything work by simply updating the value of `prefix=`.  However, your change results in a `.pc` file that looks like this:

```
Name: mongocrypt
Description: The libmongocrypt client-side field level encryption library.
Version: 1.9.0-20230509+gitaf95177a5c
Requires: libbson-1.0
prefix=/tmp/libmongocrypt
includedir=/tmp/libmongocrypt/include
libdir=/tmp/libmongocrypt/lib
Libs: /tmp/libmongocrypt/lib/libmongocrypt-static.a -lpthread -ldl /tmp/libmongocrypt/lib/libkms_message-static.a /usr/lib/x86_64-linux-gnu/libbidgcc000.a -lssl -lcrypto -lm
Cflags: -I${includedir} -DMONGOCRYPT_STATIC_DEFINE -DKMS_MSG_STATIC
```

(Note that I had specified `-DCMAKE_INSTALL_PREFIX=/tmp/libmongocrypt` in this particular case.)

Your change also leaves the `prefix=` like in the file, which could confusing for someone looking at the file.

Compare that to what the file looked like before your change:

```
Name: mongocrypt
Description: The libmongocrypt client-side field level encryption library.
Version: 1.9.0-20230509+gitaf95177a5c
Requires: libbson-1.0
prefix=/tmp/libmongocrypt
includedir=${prefix}/include/mongocrypt
libdir=${prefix}/lib
Libs: ${prefix}/lib/libmongocrypt-static.a -lpthread -ldl ${prefix}/lib/libkms_message-static.a /usr/lib/x86_64-linux-gnu/libbidgcc000.a -lssl -lcrypto -lm
Cflags: -I${includedir} -DMONGOCRYPT_STATIC_DEFINE -DKMS_MSG_STATIC
```

As it stands, I see no benefit and several drawbacks to your suggested change.

If you can provide a concrete instance where the problems you assert can be observed, then I am willing to reconsider and find a way to at least optionally implement absolute paths in a way resembling what you suggest.

K900:
Sorry for the late reply, real life got in the way for a bit, and I should have explained this better from the start. I've encountered the problem when packaging libmongocrypt for NixOS, which is a distribution with a file system layout that's ... somewhat unconventional. Due to that, we generally set `CMAKE_INSTALL_PREFIX` to `/` (which is what I tried to say with "may not be set") and set `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` to the absolute paths to the places we actually want things to be installed to (which may not share a prefix at all). I understand that it's not really the intended setup for pkg-config, and we're generally OK with patching things to behave correctly under those circumstances, so if you think the ability to move the prefix is valuable, I'll just make the changes on the NixOS side instead.

rcsanchez97:
@K900 Thanks very much for the clarification.  I can understand how in the situation you describe it would be helpful to have more control over the directories that are specified.  As a package maintainer myself (for Debian) I share your desire to avoid carrying a downstream patch if it can be avoided.  That said, my suggestion is that you add something like this at the top of `CMakeLists.txt`:

```
option (ENABLE_BUILD_FOR_NIXOS "Maintainer-only option for preparing NixOS packages" OFF)
```

Then, the remaining changes you initially suggested can be made conditional on the value of that variable.  I would prefer to have the default (`OFF`) case first, along these lines:

```
if (NOT ENABLE_BUILD_FOR_NIXOS)
   # default case
else ()
   # NixOS-specific case
endif ()
```

Then, all you would need to do in your NixOS packaging scripts is to include the option `-DENABLE_BUILD_FOR_NIXOS=ON` when you first invoke CMake.

Would that meet your needs?

Please note that we currently do not test on NixOS and we have no plans to test on it.  That said, we will do what we can to keep future changes from breaking this.

K900:
I think it makes more sense to keep the patch downstream in this case, so you don't have to carry around dead-99%-of-the-time code.

K900:
Sure. Thanks for your explanation!

PR #636, opened 2023-05-02
test(NODE-5238): set min/max range options when required.

durran:
Patch against this exact commit: https://spruce.mongodb.com/version/6451168856234306eb1923f7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/bindings/node/test/clientEncryption.test.js
+++ b/bindings/node/test/clientEncryption.test.js
@@ -711,6 +711,8 @@ describe('ClientEncryption', function () {
         algorithm: 'RangePreview',
         contentionFactor: 0,
         rangeOptions: {
+          min: new BSON.Long(0),
+          max: new BSON.Long(10),
           sparsity: new BSON.Long(1)
         },
         keyId: dataKey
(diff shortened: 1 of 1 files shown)


bindings/node/test/clientEncryption.test.js
```diff
@@ -763,6 +765,8 @@ describe('ClientEncryption', function () {
         queryType: 'rangePreview',
         contentionFactor: 0,
         rangeOptions: {
+          min: new BSON.Int32(0),
+          max: new BSON.Int32(10),
```
addaleax, line 769: I guess index.d.ts also needs an update?

durran: Tests just started failing with https://github.com/mongodb/libmongocrypt/commit/aa145adb538aeaf530e530ee801ef81391e655b4 so needed to update.

addaleax: The types indicate that `min` and `max` are optional, though – is that still accurate? (It's fine if it is, I just assumed that it isn't anymore based on the PR title)

durran: Ah... Good catch. They are required now.

W-A-James (approved):

addaleax (approved):

nbbeeken (approved):

PR #638, opened 2023-05-05
MONGOCRYPT-576 Include the error from `mcr_dll_open` in hard error

kevinAlbs:
# Summary

- Include the error from `mcr_dll_open` when crypt_shared fails to load and an override path is set.

# Background & Motivation

See MONGOCRYPT-576 for motivation for this change.


vector-of-bool (approved):
LGTM

PR #650, opened 2023-05-18
test(NODE-5268): use node 20 in ci

durran:
- Breaks out the prebuild script in the Node bindings to its own task.
- Adds Node 20 to bindings and tests it on Ubuntu 20 distros.
- Only prebuilds and publishes Node bindings on specific distros with Node 16.
- Fixes HTTP header errors on Node 20 tests.

Distros Node publishes from now:
- MacOS M1 arm64 11.00
- MacOS x64 10.14
- Ubuntu 16 arm64
- Ubuntu 16 x64
- Windows x64

### DIFF
--- a/bindings/node/.evergreen/prebuild.sh
+++ b/bindings/node/.evergreen/prebuild.sh
@@ -80,27 +80,6 @@ if [[ -n $NODE_FORCE_PUBLISH ]]; then
 elif [[ "$VERSION_AT_HEAD" != "$VERSION_AT_HEAD_1" ]]; then
   echo "Difference is package version ($VERSION_AT_HEAD_1 -> $VERSION_AT_HEAD)"
   echo "Beginning prebuild"
-
-  if [[ "$OS" == "linux" ]]; then
-    # Handle limiting which linux gets to publish prebuild
-    ARCH=$(uname -m)
-
-    if [[ $DISTRO_ID == "rhel70-small" ]]; then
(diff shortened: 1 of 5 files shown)


bindings/node/.evergreen/prebuild.sh
```diff
@@ -5,6 +5,19 @@ if [ -z ${DISTRO_ID+omitted} ]; then echo "DISTRO_ID is unset" && exit 1; fi
...
+hash -r
+
+NODE_LTS_VERSION=${NODE_LTS_VERSION:-16}
+export NODE_LTS_VERSION=${NODE_LTS_VERSION}
+source ./.evergreen/install-dependencies.sh
+
+# install node dependencies
+echo "Installing package dependencies (includes a static build)"
+bash ./etc/build-static.sh
+
```
baileympearson, line 20: since we explicitly call this script in the `build and publish node` task, we can clean up all the logic in this script that looks at architecture versions to build.  the only logic that we need to keep is the package.json version check

baileympearson: can we also remove the FLE platform matrix below or update it?  looks like it's outdated (we no longer support s390x)

baileympearson (changes requested):

bindings/node/.evergreen/prebuild.sh
durran: I've updated it to be accurate now.

baileympearson (approved):

PR #660, opened 2023-06-15
add self-test for calc_release_version.py

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/task/libmongocrypt_ubuntu2204_64_calc_release_version_selftest_patch_4ba78c5d59eaa8cd691d59bd760f48801b87d26f_648b86ca9ccd4ed2ea43aad2_23_06_15_21_46_51/logs?execution=0

Note that this implementation is such that it won't run on a patch build. As a result, the patch build above includes these additional changes to simulate the effect of what this task will accomplish when executed on the waterfall:

```
diff --git a/.evergreen/calc_release_version_selftest.sh b/.evergreen/calc_release_version_selftest.sh
index e7cf943..1bfad1b 100644
--- a/.evergreen/calc_release_version_selftest.sh
+++ b/.evergreen/calc_release_version_selftest.sh
@@ -6,6 +6,7 @@
 
 set -o errexit
 set -o pipefail
+git stash
 
 function assert_eq () {
     a="$1"
diff --git a/.evergreen/config.yml b/.evergreen/config.yml
index a1b02ac..001029b 100755
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -399,6 +399,8 @@ tasks:
           cd etc
           bash -x ../.evergreen/calc_release_version_selftest.sh
         fi
+        cd etc
+        bash -x ../.evergreen/calc_release_version_selftest.sh
 
 - name: build-and-test-shared-bson
   commands:
```

### DIFF
--- a/.evergreen/calc_release_version_selftest.sh
+++ b/.evergreen/calc_release_version_selftest.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# calc_release_version_selftest.sh is used to test output of calc_release_version.py.
+# run with:
+# cd etc
+# ./calc_release_version_selftest.sh
+
+set -o errexit
+set -o pipefail
+
+function assert_eq () {
+    a="$1"
+    b="$2"
+    if [[ "$a" != "$b" ]]; then
+        echo "Assertion failed: $a != $b"
+        # Print caller
+        caller
+        exit 1
+    fi
+}
+
+SAVED_REF=$(git rev-parse HEAD)
+
+function cleanup () {
+    rm calc_release_version_test.py
+    git checkout $SAVED_REF --quiet
+}
+
+trap cleanup EXIT
+
+# copy calc_release_version.py to a separate file not tracked by git so it does not change on `git checkout`
+cp calc_release_version.py calc_release_version_test.py
+
+echo "Test a tagged commit ... begin"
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -386,6 +386,20 @@ tasks:
   - func: "fetch source"
   - func: "run clang-tidy"
 
+- name: calc-release-version-selftest
```
kevinAlbs, line 389: Similar comment as https://github.com/mongodb/mongo-cxx-driver/pull/979#discussion_r1232516780. Suggest removing from CI and release process.

.evergreen/calc_release_version_selftest.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# calc_release_version_selftest.sh is used to test output of calc_release_version.py.
+# run with:
+# cd etc
```
kevinAlbs, line 4: Suggest moving `calc_release_version_selftest.sh` to the `etc` directory

rcsanchez97:
The last two commits sync this PR with the parallel C++ driver PR.

kevinAlbs (approved):

PR #666, opened 2023-07-07
feat(NODE-5431): add node bindings v6 deprecations

baileympearson:
This PR deprecates any public callback types and overloads of public async functions that take callbacks.  Additionally, this PR notes on any publicly exported classes that the classes will be moved into the driver in the coming major release.

### DIFF
--- a/bindings/node/README.md
+++ b/bindings/node/README.md
@@ -11,8 +11,8 @@ You can install `mongodb-client-encryption` with the following:
 npm install mongodb-client-encryption
 ```
 
-### Development 
-#### Setup 
+### Development
+#### Setup
 
 Run the following command to build libmongocrypt and setup the node bindings for development:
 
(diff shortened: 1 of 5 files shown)


bindings/node/README.md
```diff
@@ -282,12 +283,12 @@ the underlying C++ Bindings.
 
 <a name="ClientEncryption"></a>
 
-## ClientEncryption
-The public interface for explicit in-use encryption
-
+## ~~ClientEncryption~~
```
nbbeeken, line 286: Since the deprecation messages are failing to bubble up here. Can you manually add a new `h2` heading calling out the change? (use bold, place it above the generated section etc.)

nbbeeken (changes requested):

nbbeeken (approved):

PR #667, opened 2023-07-10
feat(NODE-5420)!: remove JS bindings from libmongocrypt

baileympearson:
This is a large PR, but viewing commit-by-commit helps with the changes.  The first commit removes all the JS files.

Summary of changes

- All JS bindings logic has been removed.
- Node testing tasks have been removed from CI.
- The c++ src now lives in the `addon` folder.  Build scripts have been updated to look here.
- all unnecessary dependencies have been removed from the package.json.
- The `index.js` file has been converted to TS, now lives in `src` and contains TS types for the `MongoCrypt` class and the supporting C++ Javascript classes.
- A `prepare` script has been added that builds the TS bindings and puts them in the `lib` folder (copied from the driver).

I suggest viewing this commit's diff: https://github.com/mongodb/libmongocrypt/pull/667/commits/077e61e102cb53098319051314dd74cdf36dddb1


durran (changes requested):

bindings/node/package.json
```diff
@@ -36,44 +38,24 @@
...
-    "prebuild-install": "^7.1.1",
-    "socks": "^2.7.1"
+    "prebuild-install": "^7.1.1"
   },
   "devDependencies": {
-    "@types/sinon": "^10.0.14",
-    "@types/sinon-chai": "^3.2.9",
+    "@types/bindings": "^1.5.1",
+    "@typescript-eslint/eslint-plugin": "^6.0.0",
     "chai": "^4.3.7",
```
durran, line 46: Given there's nothing left in test/ can we also remove chai, chai-subset, and mocha?

durran (changes requested):

durran (approved):

.evergreen/config.yml
```diff
@@ -1003,8 +991,6 @@ buildvariants:
   - build-and-test-and-upload
   - build-and-test-shared-bson
   - build-and-test-asan
-  - build-and-test-node
```
nbbeeken, line 1006: I think it's nice to have task history, even with all the changes we're making. I _think_ this breaks that, maybe leave in build-and-test-node, and instead make the scripting no-op for the time being.

bindings/node/.eslintrc.json
```diff
@@ -58,9 +71,9 @@
   },
   "overrides": [
     {
-      // Settings for javascript test files
+      // Settings for test files
       "files": [
-        "test/**/*.js"
+        "test/**/*.ts"
```
nbbeeken, line 76: since the test directory is going to be from scratch, let's just remove these overrides, both of the disabled rules are valid for tests just not the tests we had present when we introduced the config

bindings/node/tsconfig.json
```diff
@@ -0,0 +1,45 @@
...
+    "noImplicitReturns": true,
+    // TODO(NODE-3659): Enable useUnknownInCatchVariables and add type assertions or remove unnecessary catch blocks
+    "useUnknownInCatchVariables": false
+  },
+  "ts-node": {
+    "transpileOnly": true,
+    "compiler": "typescript-cached-transpile"
+  },
+  "include": [
+    "src/index.ts"
```
nbbeeken, line 43:
```suggestion
    "src/**/*"
```

If we stick with the driver config it's easier to update and less surprising if a new file is added

bindings/node/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  makeDataKeyContext(
+    optionsBuffer: Buffer,
+    options: {
+      keyAltNames: Buffer[];
+      keyMaterial: Buffer;
+    }
+  ): void;
+  makeRewrapManyDataKeyContext(filter: Buffer, encryptionKey: Buffer): void;
+  status: MongoCryptStatus;
+  cryptSharedLibVersionInfo: {
```
nbbeeken, line 70: I feel like there's some quick win documentation we can pull from index.d.ts before it's gone, docs that help engineers get oriented when working on this code are helpful. Essentially, can we give other options/prop the treatment `expressionMode` got?

nbbeeken (changes requested):
Mostly good to go, we can roll back some deletions to save us effort in the next piece of work

bindings/node/src/index.ts
```diff
@@ -0,0 +1,76 @@
+import * as bindings from 'bindings';
+
+const mc = bindings('mongocrypt');
+
+export interface MongoCryptCtor {
```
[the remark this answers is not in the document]
baileympearson: I tried this but we end up with "duplicate identifier" errors, because we also export an object named `MongoCrypt`.  I thought the current approach was better than renaming one or the other.

bindings/node/package.json
```diff
@@ -1,25 +1,27 @@
...
     "binding.gyp"
   ],
   "directories": {
     "lib": "lib"
   },
   "scripts": {
     "install": "prebuild-install --runtime napi --tag-prefix node-v || node-gyp rebuild",
-    "check:lint": "eslint lib test && tsd",
+    "build:ts": "tsc",
+    "check:lint": "eslint src",
```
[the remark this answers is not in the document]
baileympearson: I'd prefer not to leave an empty test directory.  leaving tooling is different.  I can just copy the files back in when/if we decide to add tests

bindings/node/src/index.ts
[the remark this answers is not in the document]
nbbeeken: Yea I used `declare class MongoCryptClass` like my suggestion to avoid that, it just allows us to use static keyword, constructor keyword, other familiar syntax.  But this approach is fine if we can make the naming updates.

bindings/node/test/types/index.test-d.ts
```diff
@@ -1,63 +0,0 @@
...
-  // aws
-  expectAssignable<KMSProviders['aws']>({ accessKeyId: '', secretAccessKey: '' });
-  expectAssignable<KMSProviders['aws']>({ accessKeyId: '', secretAccessKey: '', sessionToken: undefined });
-  expectAssignable<KMSProviders['aws']>({ accessKeyId: '', secretAccessKey: '', sessionToken: '' });
-  // automatic
-  expectAssignable<KMSProviders['aws']>({});
-
-  // azure
-  expectAssignable<KMSProviders['azure']>({ tenantId: 'a', clientId: 'a', clientSecret: 'a' });
-  expectAssignable<KMSProviders['azure']>({ tenantId: 'a', clientId: 'a', clientSecret: 'a' });
```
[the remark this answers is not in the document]
baileympearson: I don't think tests for TS that is effectively constant hold value, so I'm not going to add them unless that's something the team decides it wants.

I also don't think the release test is necessary, since it's an internal package after the next major release.

bindings/node/src/index.ts
```diff
@@ -0,0 +1,76 @@
+import * as bindings from 'bindings';
```
addaleax, line 1: Just a small suggestion, but I think

```suggestion
import bindings from 'bindings';
```

would be preferable here, it’s currently only working because TS transpiles this to the same CJS `require()` call, but in proper ESM this would just be broken

bindings/node/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const mc = bindings('mongocrypt');
+
+interface MongoCryptKMSRequest {
+  addResponse(response: Buffer): void;
+  status: MongoCryptStatus;
+  bytesNeeded: number;
+  kmsProvider: string;
+  endpoint: string;
+  message: Buffer;
```
addaleax, line 11:
```suggestion
  readonly status: MongoCryptStatus;
  readonly bytesNeeded: number;
  readonly kmsProvider: string;
  readonly endpoint: string;
  readonly message: Buffer;
```

(All of these only have getters? ditto below)

bindings/node/src/index.ts
nbbeeken:
```suggestion
import bindings = require('bindings');
```

Alternatively, we can use the verbatim syntax. Then we won't need to enable 'esModuleInterop' nor emit an import helper into the compiled code.

nbbeeken (changes requested):
I can check over the changes Anna requested, and there's some thumbs up but no change (the [export one here](https://github.com/mongodb/libmongocrypt/pull/667#discussion_r1262765006)). I think we only discussed the type tests offline, but post a "from slack: X" if I'm forgetting, ty!

nbbeeken (changes requested):

bindings/node/test/index.test.js
```diff
@@ -1,45 +0,0 @@
...
-  'MongoCryptError',
-  'MongoCryptCreateEncryptedCollectionError',
-  'MongoCryptCreateDataKeyError',
-  'MongoCryptAzureKMSRequestError',
-  'MongoCryptKMSRequestNetworkTimeoutError',
-  'AutoEncrypter',
-  'ClientEncryption'
-];
-
-describe('mongodb-client-encryption entrypoint', () => {
```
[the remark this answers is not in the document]
nbbeeken: From slack: updated matrix in https://jira.mongodb.org/browse/NODE-5424 to cover the intention here

nbbeeken (approved):

PR #671, opened 2023-07-18
PYTHON-3849 [pymongocrypt] Address utcnow deprecation

blink1073:


bindings/python/pymongocrypt/credentials.py
```diff
@@ -63,7 +63,7 @@ def _get_azure_credentials():
     # Credentials are considered expired when: Expiration - now < 1 mins.
     creds = _azure_creds_cache
     if creds:
-        if creds.expires_in - datetime.utcnow() < timedelta(seconds=60):
+        if creds.expires_in - datetime.now(tz=timezone.utc) < timedelta(seconds=60):
```
ShaneHarvey, line 66: Won't this now fail because creds.expires_in is naive and `datetime.now(tz=timezone.utc)` is aware?

ShaneHarvey (changes requested):

bindings/python/pymongocrypt/credentials.py
blink1073: `expires_in` is also using `utc`, I renamed the variable to `expires_utc` for clarity.

ShaneHarvey (approved):
LGTM!

PR #675, opened 2023-07-27
chore: type definition improvements for driver internal usage

baileympearson:


bindings/node/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "6.0.0-alpha.0",
   "description": "Official client encryption module for the MongoDB Node.js driver",
   "main": "lib/index.js",
-  "types": "index.d.ts",
+  "types": "src/index",
```
nbbeeken, line 6: Shouldn't this be the generated `lib/index.d.ts`?

bindings/node/src/index.ts
```diff
@@ -56,21 +60,30 @@ export interface MongoCrypt {
...
+  ): MongoCryptContext;
+  makeRewrapManyDataKeyContext(filter: Uint8Array, encryptionKey?: Uint8Array): MongoCryptContext;
   readonly status: MongoCryptStatus;
   readonly cryptSharedLibVersionInfo: {
     version: bigint;
     versionStr: string;
   } | null;
 }
 
+export type ExplicitEncryptionContextOptions = NonNullable<
```
nbbeeken, line 80: Can we just flip the way this is done, move the options in here, and use the type in the function? (same for `DataKeyContextOptions`?)

bindings/node/src/index.ts
```diff
@@ -11,18 +11,22 @@ interface MongoCryptKMSRequest {
...
+export interface MongoCryptStatus {
   type: number;
   code: number;
   message?: string;
 }
 
-interface MongoCryptContext {
+interface MongoCryptContextCtor {
+  new (...args: unknown[]): MongoCryptContext;
+}
```
nbbeeken, line 22: Can we mark this only for testing as well? Are there any types better than `unknown[]` worth adding? I would colocate it with the export but that's just an 🧅

nbbeeken (changes requested):

bindings/node/src/index.ts
baileympearson: I moved it below and removed the types.  The constructor type might be necessary for testing but if it is, I'll add it then.  This isn't something we ever construct from js-land outside of tests.

One other option is to define it in the tests and just cast the constructor as the appropriate type?

bindings/node/src/index.ts
nbbeeken: Loosely held opinion. I think it is easier for the uninitiated to read types declared under the named type rather than from a calculation of its location in another type.

nbbeeken (approved):

PR #678, opened 2023-07-27
MONGOCRYPT-580 Add package publication for missing os/arch combinations

rcsanchez97:
Full Evergreen patch build: https://spruce.mongodb.com/version/64bae49461837d89197ad676/tasks?limit=100&page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I also manually inspected the differences in a set of RPM packages between the patch build and its base build for RHEL8. This is because the way that RPMs are built works a little differently after updating `packager.py`. The differences that were observed were all expected and the updated `packager.py` produces working RPMs.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1122,7 +1122,7 @@ buildvariants:
     - rhel70-small
 - name: amazon2023
   display_name: "Amazon Linux 2023"
-  run_on: amazon2023.0-large
+  run_on: amazon2023.0-small
   expansions:
     has_packages: true
     packager_distro: amazon2023
@@ -1130,10 +1130,6 @@ buildvariants:
     node_lts_version: 16
   tasks:
   - build-and-test-and-upload
-  - build-and-test-shared-bson
-  - build-and-test-asan
-  - build-and-test-node
-  - test-java
   - name: publish-packages
     distros:
     - rhel70-small
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1120,6 +1120,39 @@ buildvariants:
   - name: publish-packages
     distros:
     - rhel70-small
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -1120,6 +1120,39 @@ buildvariants:
   - name: publish-packages
     distros:
     - rhel70-small
+- name: amazon2023
+  display_name: "Amazon Linux 2023"
+  run_on: amazon2023.0-large
```
kevinAlbs, line 1125:
```suggestion
  run_on: amazon2023.0-small
```

I suggest using the `-small` distros for libmongocrypt tasks. I expect these tasks are not resource intensive (compared with server build or test tasks) and may not benefit much from the `-large` distros.

Comment applies to all new variants.

.evergreen/config.yml
```diff
@@ -1120,6 +1120,39 @@ buildvariants:
...
+    has_packages: true
+    packager_distro: amazon2023
+    packager_arch: x86_64
+    node_lts_version: 16
+  tasks:
+  - build-and-test-and-upload
+  - build-and-test-shared-bson
+  - build-and-test-asan
+  - build-and-test-node
+  - test-java
```
kevinAlbs, line 1136:
```suggestion
```

Suggest only running the `build-and-test-and-upload` and `publish-packages` tasks for the new variants. Those appear to be the only necessary tasks to publish packages.

It may be added maintenance burden for the Java and Node tasks to pass on all variants. Example: each distro may need to ensure all necessary Java and Node dependencies are available.

Comment applies to all new variants.

kevinAlbs (approved):
LGTM. I left a suggestion to use the `-small` variants, and to remove tasks that may not be needed.

PR #679, opened 2023-07-27
assert return of `EVP_CIPHER_CTX_new`

kevinAlbs:
# Background & Motivation
Intends to resolve Coverity reported issue with CID 121624.

`man EVP_CIPHER_CTX_new` on macOS documents:

> EVP_CIPHER_CTX_new() returns a pointer to a newly created EVP_CIPHER_CTX for success and NULL for failure.


rcsanchez97 (approved):
Sorry for the delay. I somehow didn't notice I had this review pending.

PR #681, opened 2023-07-30
[MONGOCRYPT-583] cmake/bson: allow shared-only system lib

tharvik:
see [jira's issue](https://jira.mongodb.org/projects/MONGOCRYPT/issues/MONGOCRYPT-583) for general intro. some comments now

* when loading system, only load the wanted library, so it creates target `bson_shared` if `USE_SHARED_LIBBSON` else `bson_shared`; that allows to have {shared,static}-only system library
* extract a function to actually perform the system library load that can be used by both shared and static; also fail early if lib not found
* now, `_mongocrypt-libbson_for_{shared,static}` links to shared only iff `USE_SHARED_LIBBSON` else to static
  * fun info: one can craft a static mongocrypt with a dynamic libbson (dunno why one would want that)
* move around some code
  * install static lib iff we localy built it
  * set runtime output of shared iff we localy built it
* for enforcing the local build for test, I wasn't really sure of the check you wanted to make, I took a broad one (fails if `test && (system || shared)`) but really depend on your use cases

### DIFF
--- a/cmake/ImportBSON.cmake
+++ b/cmake/ImportBSON.cmake
@@ -109,26 +109,35 @@ if (ENABLE_ONLINE_TESTS AND (MONGOCRYPT_MONGOC_DIR STREQUAL "USE-SYSTEM" OR USE_
    message (FATAL_ERROR "Online tests requires static local build of libbson")
 endif ()
 
+function (_import_system_libbson target library_type library_name)
+   # import a system libbson
+   # target: target to add library under
+   # library_type: STATIC or SHARED
+   # library_name: filename to seek
+
+   set (lib "_MONGOCRYPT_SYSTEM_LIBBSON_${library_type}")
+
+   find_library (${lib} ${library_name})
+   if (${${lib}} STREQUAL "${lib}-NOTFOUND")
+      message (FATAL_ERROR "system ${library_name} not found")
+   endif ()
+   find_path (_MONGOCRYPT_SYSTEM_LIBBSON_INCLUDE_DIR bson/bson.h PATH_SUFFIXES libbson-1.0)
+
+   add_library (${target} ${library_type} IMPORTED)
+
+   set_target_properties (${target} PROPERTIES
+      IMPORTED_CONFIGURATIONS "Release"
+      INTERFACE_INCLUDE_DIRECTORIES "${_MONGOCRYPT_SYSTEM_LIBBSON_INCLUDE_DIR}"
+      IMPORTED_LOCATION "${${lib}}"
+   )
+   set_property (CACHE ${lib} _MONGOCRYPT_SYSTEM_LIBBSON_INCLUDE_DIR PROPERTY ADVANCED TRUE)
+endfunction ()
+
 function (_import_bson)
    if (MONGOCRYPT_MONGOC_DIR STREQUAL "USE-SYSTEM" AND USE_SHARED_LIBBSON)
       message (STATUS "NOTE: Using system-wide libbson library. This is intended only for package maintainers.")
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Sorry for the delay on this one. We have some open tasks to improve the handling of simultaneous/selective static/shared builds in libmongocrypt, libmongoc, and libbson. There's also an unfortunate cyclic dependency-ish between the projects that led to the strange manual import behavior that is currently seen in the code.

Unfortunately, the changes as they currently are may wreak havoc on some of our other users, but I do sympathize with the challenges of packaging according to distribution policies, and I will happily help get this sorted. If I could collect more information: What are the constraints that you are trying to satisfy that are not currently met by the current process?

tharvik:
> Sorry for the delay on this one.

no worries, I'm late too :)

> We have some open tasks to improve the handling of simultaneous/selective static/shared builds in libmongocrypt, libmongoc, and libbson.

ho, I missed thoses then, I hope I didn't duplicate an existing one. can you point me to the related ones you've in mind?

> What are the constraints that you are trying to satisfy that are not currently met by the current process?

when packaging on Gentoo, we try our best to unbundle all dependencies of a given project; in my case, I'm trying to package libmongocrypt which (by default) bundles libbson. we also let users choose if they want a static lib **in addition** to a dynamic library (via the [`static-libs`](https://packages.gentoo.org/useflags/static-libs) [USE flag](https://wiki.gentoo.org/wiki/USE_flag)).

I though by that simply using `-DMONGOCRYPT_MONGOC_DIR=USE-SYSTEM` for unbundling and a combination `-DENABLE_STATIC` and `-DUSE_SHARED_LIBBSON` for the static part would work as I intended. spoiler, it didn't.

* when build a dynamic-only lib, so `-DUSE_SHARED_LIBBSON=YES -DENABLE_STATIC=NO`, it fails with
  ```
  ninja: error: '_MONGOCRYPT_SYSTEM_LIBBSON_STATIC-NOTFOUND', needed by 'mongo_crypt_v1.so', missing and no known rule to make it
  ```
  indeed, system libbson is dynamic-only, no way to find a static version on the system, but I specifically asked for no static. installing a static and dynamic libbson works but that disregards users' wishes to have everything dynamic.

* when build a static and dynamic lib, so `-DUSE_SHARED_LIBBSON=NO -DENABLE_STATIC=YES` (setting the shared libbson flag to true gives the previous error), it fails with 
  ```
  [...]
  -- Using [USE-SYSTEM] as a sub-project for libbson

  CMake Error at cmake/ImportBSON.cmake:148 (add_subdirectory):
    add_subdirectory given source "USE-SYSTEM" which is not an existing
    directory.
    cmake/ImportBSON.cmake:164 (_import_bson)
    CMakeLists.txt:47 (include)

  CMake Error at cmake/ImportBSON.cmake:195 (set_property):
    set_property could not find TARGET bson_shared.  Perhaps it has not yet
    been created.
    CMakeLists.txt:47 (include)

  CMake Error at cmake/ImportBSON.cmake:202 (set_target_properties):
    set_target_properties Can not find target to add properties to: bson_static
    CMakeLists.txt:47 (include)
  ```
  so that's kinda unexpected. I asked for the system lib via this special value of `MONGOCRYPT_MONGOC_DIR` but it is used as a path which is clearly wrong.

to summarize, I want to be able to
 * build `libmongocrypt.so` while having only `libbson.so`
 * and, based on user choice, to also have `libmongocrypt.a` (in this case, `libbson.a` will be provided by pkg manager)

bisht2050:
@tharvik Is this still relevant for you?

tharvik:
haha, a 3y old PR. yeah, that'd still help package mongo-crypt on gentoo.

PR #682, opened 2023-08-04
refactor(NODE-5432): Remove usages of Buffer from Node C++ bindings

baileympearson:
There are three main changes in this PR

- we now lint both C++ and JS in CI.  We rely on the `.clang-format` file in the project root.  The C++ is also linted, so the diff is large.  This is contained in commit.
- Basic unit tests of our C++ bindings have been added.  These unit tests serve to assert that we correctly validate arguments provided from JS into our C++.  The unit tests only cover errors thrown by our C++ logic and intentionally do not test error scenarios that `libmongocrypt` guards against.
- All usages on `Value::IsBuffer` have been removed from the C++ logic and replaced with `Value::IsTypedArray()`.  Note that the _return types_ of FLE APIs that return Buffers are unmodified, because we explicitly construct Buffers to return.


durran (changes requested):
There are a couple distinct errors in CI.

1. The windows run seems to be timing out on the release test.
2. Ubuntu 20 is failing during build with a node-gyp error. I have seen this before but don't recall exactly how I got around it. It did have something to do with either a) prebuild is using an old version of node-gyp, or b) the python version on Ubuntu 20 was having an affect on cflags being readonly.

I checked the C++ code based on the commit that had the actual behavioural changes, not the entire changes with linting. Just the one question there. Tests otherwise look good.

bindings/node/addon/mongocrypt.h
```diff
@@ -18,97 +18,94 @@ extern "C" {
 namespace node_mongocrypt {
 
 struct MongoCryptBinaryDeleter {
-    void operator()(mongocrypt_binary_t* binary) {
+    void operator()(mongocrypt_binary_t *binary) {
```
addaleax, line 21: > We rely on the `.clang-format` file in the project root

… which is targeting C code and not C++, leading to changes like this?

bindings/node/addon/mongocrypt.cc
```diff
@@ -18,113 +19,112 @@ struct InstanceData {
...
 
-Uint8Array BufferWithLengthOf(Env env, mongocrypt_binary_t* binary) {
+Uint8Array BufferWithLengthOf(Env env, mongocrypt_binary_t *binary) {
     size_t len = mongocrypt_binary_len(binary);
     return Buffer<uint8_t>::New(env, len);
 }
 
-void CopyBufferData(mongocrypt_binary_t* out, Uint8Array buffer, size_t count) {
+Uint8Array Uint8ArrayFromValue(Napi::Value v, std::string argument_name) {
+    if (!v.IsTypedArray()) {
```
addaleax, line 58:
```suggestion
    if (!v.IsTypedArray() || v.As<TypedArray>().TypedArrayType() != napi_uint8_array) {
```

(or otherwise I think you’re also fine just using `TypedArray` instead of `Uint8Array` everywhere)

bindings/node/addon/mongocrypt.h
baileympearson: Neal and I thought that the C format style would also apply to c++.  Are there different conventions?  I can revert and reformat the bindings using the old clang format file.

addaleax: I mean, at the end of the day linting rules are personal preferences and you can do whatever you prefer, it's like single vs double quotes in JS.

That being said, yes, it's been a long time since I've seen left-leaning asterisks for pointers in C or right-leaning ones in C++, and, very personally, yes this does just look a bit _odd_ to me. (Random example if you google this: https://stackoverflow.com/questions/2660633/declaring-pointers-asterisk-on-the-left-or-right-of-the-space-between-the-type even uses "C style" and "C++ style" to refer to the two styles, respectively).

addaleax (approved):

durran (approved):

nbbeeken (approved):

PR #690, opened 2023-08-30
MONGOCRYPT-589 Add bulk decrypt benchmark with Java bindings

kevinAlbs:
# Summary

Add a benchmark in the Java bindings to measure throughput of decrypting documents with many encrypted values.

This PR partially resolves MONGOCRYPT-589. A following PR is planned to export the struct definition of `mongocrypt_binary_t` and update bindings to avoid calls to `mongocrypt_binary_data` and `mongocrypt_binary_len`.

# Background & Motivation

The benchmark is intended to measure the workload identified in DRIVERS-2581 and compare future changes.

The benchmark uses the `rhel90-dbx-perf-large` distro suggested in DRIVERS-2666. The results appears stable: [10 executions](https://spruce.mongodb.com/task/libmongocrypt_benchmark_benchmark_java_patch_f6817fa54b4913d42b8f8718577f4884133530fb_64eca5d2a4cf47658d19ef8d_23_08_28_13_49_06/logs?execution=9) produced the same `Decrypting median ops/sec` of `31`.

[Java Microbenchmarking Harness (JMH)](https://github.com/openjdk/jmh) was briefly considered but decidedly not used. JMH does not have native Gradle support. 

The task uses JDK 8. I am not sure if JDK 8 is needed or wanted. JDK 8 was used for consistency with the existing `test-java` task. Use of JDK 8 resulted in an observed hang in Gradle configure documented in MONGOCRYPT-590. A work around was applied in https://github.com/mongodb/libmongocrypt/pull/690/commits/2cb869ae3f963d021cfcc31e370344973aad8727.

### DIFF
--- a/bindings/java/mongocrypt/benchmarks/src/main/java/com/mongodb/crypt/benchmark/BenchmarkRunner.java
+++ b/bindings/java/mongocrypt/benchmarks/src/main/java/com/mongodb/crypt/benchmark/BenchmarkRunner.java
@@ -117,13 +117,13 @@ public class BenchmarkRunner {
                 }
             }
 
-            String created_at = ZonedDateTime.now(ZoneOffset.UTC).format(DateTimeFormatter.ISO_INSTANT);
+            String createdAt = ZonedDateTime.now(ZoneOffset.UTC).format(DateTimeFormatter.ISO_INSTANT);
             // Warm up benchmark and discard the result.
             measureMedianOpsPerSecOfDecrypt(mongoCrypt, encrypted, NUM_WARMUP_SECS);
             // Decrypt `encrypted` and measure ops/sec.
             long medianOpsPerSec = measureMedianOpsPerSecOfDecrypt(mongoCrypt, encrypted, NUM_SECS);
             System.out.printf("Decrypting 1500 fields median ops/sec : %d%n", medianOpsPerSec);
-            String completed_at = ZonedDateTime.now(ZoneOffset.UTC).format(DateTimeFormatter.ISO_INSTANT);
+            String completedAt = ZonedDateTime.now(ZoneOffset.UTC).format(DateTimeFormatter.ISO_INSTANT);
 
             // Print the results in JSON that can be accepted by the `perf.send` command.
             // See https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Commands#perfsend for the expected `perf.send` input.
@@ -131,8 +131,8 @@ public class BenchmarkRunner {
                     Arrays.asList(
                             new BsonDocument()
                                     .append("info", new BsonDocument().append("test_name", new BsonString("java_decrypt_1500")))
-                                    .append("created_at", new BsonString(created_at))
-                                    .append("completed_at", new BsonString(completed_at))
+                                    .append("created_at", new BsonString(createdAt))
+                                    .append("completed_at", new BsonString(completedAt))
                                     .append("artifacts", new BsonArray())
                                     .append("metrics", new BsonArray(Arrays.asList(
                                             new BsonDocument()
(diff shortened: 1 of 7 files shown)


jyemin (commented):
JDK 8 is not needed, and in fact will no longer be supported by the driver fairly soon.  It's fine to run with 11 or 17.

jyemin (changes requested):
Initial feedback

bindings/java/mongocrypt/benchmarks/src/main/java/com/mongodb/crypt/benchmark/BenchmarkRunner.java
```diff
@@ -0,0 +1,178 @@
...
+            -27, -81
+    };
+
+    private static String getFileAsString(final String fileName, String lineSeparator) {
+        try {
+            URL resource = BenchmarkRunner.class.getResource("/" + fileName);
+            if (resource == null) {
+                throw new RuntimeException("Could not find file " + fileName);
+            }
+            File file = new File(resource.toURI());
```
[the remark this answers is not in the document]
jyemin: Even better!

bindings/java/mongocrypt/benchmarks/src/main/java/com/mongodb/crypt/benchmark/BenchmarkRunner.java
```diff
@@ -0,0 +1,178 @@
...
+                .builder()
+                .localKmsProviderOptions(MongoLocalKmsProviderOptions.builder()
+                        .localMasterKey(ByteBuffer.wrap(LOCAL_MASTER_KEY))
+                        .build())
+                .build());
+    }
+
+    private static long measureMedianOpsPerSecOfDecrypt(MongoCrypt mongoCrypt, BsonDocument toDecrypt) {
+        ArrayList<Long> opsPerSecs = new ArrayList<Long>(NUM_SECS);
+        for (int i = 0; i < NUM_WARMUP_SECS + NUM_SECS; i++) {
```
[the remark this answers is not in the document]
jyemin: Ah, I see.  I like it better this way anyway, so thanks for making the change.

jyemin (approved):

bindings/java/mongocrypt/benchmarks/src/main/java/com/mongodb/crypt/benchmark/BenchmarkRunner.java
```diff
@@ -0,0 +1,157 @@
...
+            -27, -81
+    };
+
+    private static String getFileAsString(final String fileName) {
+        try {
+            URL resource = BenchmarkRunner.class.getResource("/" + fileName);
+            if (resource == null) {
+                throw new RuntimeException("Could not find file " + fileName);
+            }
+            return new String(Files.readAllBytes(Paths.get(resource.toURI())));
```
jyemin, line 53: Does using readAllBytes fix something?  I would have thought the previous incarnation would work just as well.

kevinAlbs: `readString` was introduced in Java 11. The `test-java` tasks use JDK 8.

Alternatively: can the `test-java` tasks use JDK 11 or 17? I expected no, since the declared `sourceCompatibility` and `targetCompatibility` is `VERSION_1_8`. I expect coverage is still needed to test with JDK 8.

jyemin: Yes, it's more important to test with the minimum supported Java than the maximum.  We could probably figure out a way to compile the benchmark runner with a newer JDK, but let's not worry about that.

rozza (approved):

PR #691, opened 2023-09-01
MONGOCRYPT-587 compiler warnings

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/64f1354be3c3312f15caa880/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/test/test-mongocrypt-traverse-util.c
+++ b/test/test-mongocrypt-traverse-util.c
@@ -101,7 +101,6 @@ static void _assert_correct_fields(bson_t *bson,
                                    int num_binary,
                                    int num_other,
                                    _util_tester_t *tester) {
-    char name[36];
     bson_iter_t iter;
     bson_iter_t child;
     int i;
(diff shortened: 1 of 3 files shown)


test/test-mongocrypt-traverse-util.c
```diff
@@ -116,28 +116,33 @@ static void _assert_correct_fields(bson_t *bson,
 
     if (tester->parent != NEST_IN_NONE) {
         for (i = 0; i < num_other; i++) {
-            snprintf(name, 13, "other.field%d", i);
+            /* use mod 10, since guaranteed < 10 and to silence -Wformat-truncation */
+            snprintf(name, 13, "other.field%d", i % 10);
             bson_iter_init(&iter, bson);
             BSON_ASSERT(bson_iter_find_descendant(&iter, name, &child));
```
kevinAlbs, line 122: I expect there is no need to reuse `name` since this is test code.

Consider using `bson_strdup_printf` to allocate the keys in these tests to avoid the `mod 10` workaround:

```c
        for (i = 0; i < num_other; i++) {
            char *key = bson_strdup_printf("other.field%d", i);
            bson_iter_init(&iter, bson);
            BSON_ASSERT(bson_iter_find_descendant(&iter, key, &child));
            bson_free(key);
        }
```

kevinAlbs (approved):
LGTM with possible suggestion.

test/test-mongocrypt-traverse-util.c
rcsanchez97: Oh, that's a good idea. It definitely did not occur to me. Change incoming.

vector-of-bool (approved):
Sorry to delay. LGTM!

PR #693, opened 2023-09-05
MONGOCRYPT-589 export `mongocrypt_binary_t`

kevinAlbs:
# Scope

- Export definition of `mongocrypt_binary_t`.
- Read fields of `mongocrypt_binary_t` directly in Java bindings.

# Motivation

Languages interfacing with libmongocrypt may observe slower throughput with frequent calls to C functions.

[HELP-27047](https://jira.mongodb.org/browse/HELP-27047) identifies a problematic workload: decrypting documents with 1500 encrypted values.

Profiling showed significant samples spent in calls to `mongocrypt_binary_data` and `mongocrypt_binary_len`.

[A prototype benchmark](https://github.com/mongodb/DRIVERS-2581) showed improvement by using the struct definition directly, rather than calling `mongocrypt_binary_data` and `mongocrypt_binary_len`:

```
run_test 'Baseline' ... begin
Baseline   (ms): 9.868333
Decrypting (ms): 35.971542
Overhead   (ms): 26.103209
Overhead   (+%): 264.51487804475187
run_test 'Baseline' ... end
run_test 'With Java using mongocrypt_binary_t definition' ... begin
Baseline   (ms): 9.756125
Decrypting (ms): 31.031459
Overhead   (ms): 21.275334
Overhead   (+%): 218.071560173737
run_test 'With Java using mongocrypt_binary_t definition' ... end
```

With these changes, the `benchmark-java` task [reported 40 ops/sec](https://spruce.mongodb.com/task/libmongocrypt_benchmark_benchmark_java_patch_ce14f0cebfe9d33bf5d2f5467a0fb8a36908e80c_64f72794d6d80a9b1bd6c970_23_09_05_13_05_25/trend-charts?execution=0). The previous run [reported 29 ops/sec](https://spruce.mongodb.com/task/libmongocrypt_benchmark_benchmark_java_ce14f0cebfe9d33bf5d2f5467a0fb8a36908e80c_23_09_05_12_41_16/trend-charts?execution=0).

# Rejected Alternatives.

The proposed changes use `getPointer()` and `getInt()` to read the `mongocrypt_binary_t` fields directly. Defining Structure was attempted:

```java
@Structure.FieldOrder({ "data", "len" })
public static class mongocrypt_binary_t extends Structure {
    public Pointer data;
    public int len;
}
```

A local run showed a decrease in ops/sec from the baseline (44  => 36). [Comparing profile snapshots](https://www.jetbrains.com/help/idea/compare-profiler-snapshots.html) showed many samples in translating Structure:

reading a Structure

Red sections are new samples, green are missing samples. Since throughput reduced, I interpret these results to mean: less work was completed, and more of the time spent was translating Structure.

### DIFF
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
@@ -92,7 +92,7 @@ public class CAPI {
             // Representing `uint32_t` values greater than INT32_MAX is represented as a negative `int`.
             // Throw an exception. mongocrypt_binary_t is not expected to use lengths greater than INT32_MAX.
             if (len < 0) {
-                throw new IllegalArgumentException(
+                throw new AssertionError(
                         String.format("Expected mongocrypt_binary_t length to be non-negative, got: %d", len));
             }
             return len;
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
@@ -87,7 +87,16 @@ public class CAPI {
             return this.getPointer().getPointer(0);
         }
         public int len() {
-            return this.getPointer().getInt(Native.POINTER_SIZE);
+            int len = this.getPointer().getInt(Native.POINTER_SIZE);
+            // mongocrypt_binary_t represents length as an unsigned `uint32_t`.
+            // Representing `uint32_t` values greater than INT32_MAX is represented as a negative `int`.
+            // Throw an exception. mongocrypt_binary_t is not expected to use lengths greater than INT32_MAX.
+            if (len < 0) {
+                throw new IllegalArgumentException(
+                        String.format("Expected mongocrypt_binary_t length to be non-negative, got: %d", len));
+            }
+            return len;
+
         }
     }
 
--- a/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
+++ b/bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
@@ -75,7 +75,7 @@ public class CAPI {
     public static class mongocrypt_binary_t extends PointerType {
         // The `mongocrypt_binary_t` struct layout is part of libmongocrypt's ABI:
         // typedef struct _mongocrypt_binary_t {
-        //     uint8_t *data;
+        //     void *data;
         //     uint32_t len;
         // } mongocrypt_binary_t;
         // To improve performance, fields are read directly using `getPointer` and `getInt`.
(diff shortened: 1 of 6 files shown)


src/mongocrypt.h
```diff
@@ -69,8 +69,14 @@ const char *mongocrypt_version(uint32_t *len);
...
  * mongocrypt_binary_t is valid until the parent ctx is destroyed with @ref
  * mongocrypt_ctx_destroy.
+ *
+ * The `mongocrypt_binary_t` struct definition is public.
+ * Consumers may rely on the struct layout.
  */
-typedef struct _mongocrypt_binary_t mongocrypt_binary_t;
+typedef struct _mongocrypt_binary_t {
+    uint8_t *data;
+    uint32_t len;
```
eramongodb, line 78:
```suggestion
    void *data;
    size_t len;
```
Noting that making this public marks a point of no return: the opportunity to change these data members without requiring a subsequent API/ABI break is now. Is there compelling reason/rationale to preserve `uint32_t` as the type of `len` rather than using `size_t` or `uint64_t`?

The `uint8_t*` -> `void*` recommendation is just for better consistency with C convention for a "pointer to a blob of bytes" (i.e. see C stdlib functions such as `memcpy()`, which also saves users an explicit cast to `uint8_t*` from their own pointer-to-blob types such as `char*`). AFAIK it should make no functional difference, although it may complicate our implementation by requiring a cast back to `uint8_t*` before performing pointer arithmetic, which atm seems to only impact [`_crypto_aes_256_ctr_encrypt_decrypt_via_ecb()`](https://github.com/mongodb/libmongocrypt/blob/ce14f0cebfe9d33bf5d2f5467a0fb8a36908e80c/src/mongocrypt-crypto.c#L81) and some test code, sans any potential external uses of `mongocrypt-binary-private.h`.

kkloberdanz (approved):
Looks good to me

bindings/java/mongocrypt/src/main/java/com/mongodb/crypt/capi/CAPI.java
```diff
@@ -73,6 +73,31 @@ public static class mongocrypt_opts_t extends PointerType {
...
+        public Pointer data() {
+            return this.getPointer().getPointer(0);
+        }
+        public int len() {
+            int len = this.getPointer().getInt(Native.POINTER_SIZE);
+            // mongocrypt_binary_t represents length as an unsigned `uint32_t`.
+            // Representing `uint32_t` values greater than INT32_MAX is represented as a negative `int`.
+            // Throw an exception. mongocrypt_binary_t is not expected to use lengths greater than INT32_MAX.
+            if (len < 0) {
+                throw new IllegalArgumentException(
```
jyemin, line 95: Let's use AssertionError instead since this is all internal code.

jyemin (changes requested):
Looks great overall.  Just a small change request.

jyemin (approved):
LGTM

PR #706, opened 2023-10-25
MONGOCRYPT-601 fix build for Alpine on ARM64

kkloberdanz:
MONGOCRYPT-601

Without this fix, building on Alpine on ARM64 results in the compiler error below.
Since we do not support ancient versions of Solaris, the redefinition of `wchar_t` does not seem necessary, therefore it is safe to remove.
```
In file included from /build/cmake-build/_deps/intel_dfp-src/LIBRARY/src/bid_internal.h:41,
                 from /build/cmake-build/_deps/intel_dfp-src/LIBRARY/src/bid_trans.h:61,
                 from /build/cmake-build/_deps/intel_dfp-src/LIBRARY/src/bid128_asin.c:29:
/build/cmake-build/_deps/intel_dfp-src/LIBRARY/src/bid_functions.h:46:15: error: conflicting types for 'wchar_t'; have 'int'
   46 | typedef int   wchar_t;
      |               ^~~~~~~
In file included from /usr/include/stdlib.h:21,
                 from /usr/include/fortify/stdlib.h:23,
                 from /build/cmake-build/_deps/intel_dfp-src/LIBRARY/src/bid_internal.h:33:
/usr/include/bits/alltypes.h:15:18: note: previous declaration of 'wchar_t' with type 'wchar_t' {aka 'unsigned int'}
   15 | typedef unsigned wchar_t;
      |                  ^~~~~~~
```

### DIFF
--- a/cmake/IntelDFP.cmake
+++ b/cmake/IntelDFP.cmake
@@ -45,6 +45,7 @@ FetchContent_Declare (
             ${patch_input_opt} "${PROJECT_SOURCE_DIR}/etc/mongo-inteldfp-s390x.patch"
             ${patch_input_opt} "${PROJECT_SOURCE_DIR}/etc/mongo-inteldfp-MONGOCRYPT-571.patch"
             ${patch_input_opt} "${PROJECT_SOURCE_DIR}/etc/mongo-inteldfp-libmongocrypt-pr-625.patch"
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
> Since we do not support ancient versions of Solaris

I expect any Solaris workarounds can be removed. C driver dropped support for Solaris in 1.8.0 as part of CDRIVER-2230. The MongoDB server also no longer supports Solaris.

vector-of-bool (approved):

PR #708, opened 2023-11-27
MONGOCRYPT-603 Don't return a compile error if dll_path unsupported

NattyNarwhal:
This returns an error to the caller instead of a hard compile error. The 34a6e9d585e4b8fd1c4f18ffa57a08a6f3338546 commit partially addressed this, but this file wasn't touched by it.

This lets mongocrypt work on a platform without dll_path support, like AIX, but CSFLE is untested.


vector-of-bool (approved):
Thank you for your contribution! I have one tweak to make, but nothing big.

PR #710, opened 2023-11-28
Use gradles features to make slf4j optional

rozza:
See: https://docs.gradle.org/current/userguide/feature_variants.html

MONGOCRYPT-602


jyemin (approved):
I like it!  Good to know about this Gradle feature feature ;)

PR #712, opened 2023-12-05
MONGOCRYPT-593 Unskip `test-java` task on `Debian 11.0`

kevinAlbs:
# Summary

Unskip `test-java` task on `Debian 11.0`. Remove `test-java` tasks on `Ubuntu 20.04 arm64` and `Ubuntu 22.04 arm64`.

`test-java` was run on `Debian 11.0` in this patch build: https://spruce.mongodb.com/version/656e42417742ae3e6c6607b6

# Background & Motivation

BUILD-17899 includes the JDK on `Debian 11.0`. Comments suggest `Ubuntu 20.04 arm64` and `Ubuntu 22.04 arm64` are not expected to include JDK.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1485,7 +1485,6 @@ buildvariants:
   - build-and-test-shared-bson
   - build-and-test-asan
   - build-and-test-node
-  # - test-java - Fails with `JAVA_HOME is set to an invalid directory`. Unskip once MONGOCRYPT-593 is resolved.
   - name: publish-packages
     distros:
     - ubuntu2004-small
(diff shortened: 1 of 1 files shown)


jyemin:
> Comments suggest Ubuntu 20.04 arm64 and Ubuntu 22.04 arm64 are not expected to include JDK.

Which comments are you referring to?

kevinAlbs:
> Which comments are you referring to?

These:

> If it's a heavy lift to add arm support to the toolchain, I think we can live without it.

> Added java to debian11

jyemin (approved):

PR #720, opened 2023-12-12
PYTHON-4091 Skip perf benchmark when running test suite

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4091


blink1073 (approved):
Thanks!

ShaneHarvey:
Note we can move this pytest option to pyproject.toml after https://github.com/mongodb/libmongocrypt/pull/721

caseyclements:
One can also use pytest fixtures to create custom marks (e.g. slow) and then expose these as command line arguments

PR #726, opened 2023-12-26
MONGOCRYPT-611 Implement E[DS]CTwiceDerivedToken(Tag|Value)? tokens

sgolemon:
I added this ticket to the epic while working on MONGOCRYPT-608 and unwrapping where these server payload constructions come from and realizing that we'll need the "TwiceDerived" tokens eventually regardless of how 608 gets implemented.  Simple enough and an easy review over the holidays.


src/mc-tokens-private.h
```diff
@@ -44,6 +44,14 @@
  * ESCDerivedFromDataTokenAndCounter = HMAC(ESCDerivedFromDataToken, u)
  * ECCDerivedFromDataTokenAndCounter = HMAC(ECCDerivedFromDataToken, u)
  *
+ * EDCTwiceDerivedToken      = HMAC(EDCDerivedFromDataTokenAndCounter, 1)
+
+ * ESCTwiceDerivedTagToken   = HMAC(ESCDerivedFromDataTokenAndCounter, 1)
+ * ESCTwiceDerivedValueToken = HMAC(ESCDerivedFromDataTokenAndCounter, 2)
+
+ * ECCTwiceDerivedTagToken   = HMAC(ECCDerivedFromDataTokenAndCounter, 1)
+ * ECCTwiceDerivedValueToken = HMAC(ECCDerivedFromDataTokenAndCounter, 2)
```
kevinAlbs, line 53: Suggest removing the ECC tokens.

I expect the ECC tokens are not needed for the V2 protocol. SERVER-74066 notes:
> In V2, since ECC is removed

ECC tokens were removed from protocol documentation in: https://github.com/10gen/mongo-enterprise-modules/commit/abe05cef537c40d14881d19b8bff71bd6baf5c94#diff-5de05088df1ea1a0745044571bfe9c684d033c18c156d81f2e0d4de50c836403L731-L732

test/test-mc-tokens.c
```diff
@@ -152,6 +157,20 @@ static void _mc_token_test_run(_mongocrypt_tester_t *tester, const char *path) {
...
+    mc_##Name##TwiceDerived##Suffix##_t *Name##TwiceDerived##Suffix =                                                  \
+        mc_##Name##TwiceDerived##Suffix##_new(crypt->crypto, Name##DerivedFromDataTokenAndCounter, status);            \
+    ASSERT_OR_PRINT(Name##TwiceDerived##Suffix, status);                                                               \
+    ASSERT_CMPBUF(*mc_##Name##TwiceDerived##Suffix##_get(Name##TwiceDerived##Suffix),                                  \
+                  test.Name##TwiceDerived##Suffix);
+   TEST_TWICE(EDC, Token);
+   TEST_TWICE(ESC, TagToken);
+   TEST_TWICE(ESC, ValueToken);
+   TEST_TWICE(ECC, TagToken);
+   TEST_TWICE(ECC, ValueToken);
```
kevinAlbs, line 171: Run `./etc/format-all.sh` to format and fix the `check-format` task.

kevinAlbs (commented):
LGTM with a format fix and removal of no-longer-needed ECC tokens.

kevinAlbs (approved):

src/mc-tokens-private.h
sgolemon: Yeah, that occurred to me as well.  I kept them in for completeness sake, but I suppose dead code is dead code and it's only serving to make the binary larger in the end. Fine, I'll kill it.

test/test-mc-tokens.c
sgolemon: Ah, that's the script I was looking around for. Thanks!

sgolemon:
evergreen merge

PR #728, opened 2024-01-02
MONGOCRYPT-580 treat "arm64" as "aarch64" on RPM-based distros

rcsanchez97:
for the purposes of creating packages published in the PPA

Follow-up for ef5b0b54adaadf3b8a1a151aa31b91702bf17187

After merging this into `master`, I will also cherry-pick to `r1.8`.


kevinAlbs (approved):

rcsanchez97:
The actual problem is actually something else. New PR incoming.

PR #734, opened 2024-01-05
PYTHON-4115 [pymongocrypt] Bundle libmongocrypt 1.8.4

blink1073:


bindings/python/CHANGELOG.rst
```diff
@@ -5,6 +5,8 @@ Changes in Version 1.8.0
 ------------------------
 
 - Update from manylinux2010 to manylinux2014 wheels.
+- Bundle libmongocrypt 1.8.2 in release wheels.
```
ShaneHarvey, line 8: 1.8.2->1.8.4

ShaneHarvey (commented):
Should we add rhel-70-arm64 to the test matrix too to make sure it actually works?

bindings/python/CHANGELOG.rst
blink1073: Fixed

blink1073:
> Should we add rhel-70-arm64 to the test matrix too to make sure it actually works?

I added `rhel-82-arm64` and `rhel-70-64-bit` tests.

ShaneHarvey (approved):

PR #738, opened 2024-01-23
MONGOCRYPT-616 add `mongocrypt_is_crypto_available`

kevinAlbs:
# Summary

Add `mongocrypt_is_crypto_available` to report whether libmongocrypt is built with native crypto support.

# Background & Motivation

libmongocrypt can be built without native crypto support with the CMake option `DISABLE_NATIVE_CRYPTO`. If native crypto is disabled, crypto primitives must be supplied with callbacks (e.g. `mongocrypt_setopt_crypto_hooks` and friends).

DRIVERS-2718 proposes having driver bindings prefer using libmongocrypt with native crypto when available for the observed performance imporvement. This may require driver check if libmongocrypt is built with native crypto at runtime:

> Add a method in CAPI like bool isCryptoAvailable(). Change the wrapper to call that method and only register crypto callbacks if it returns false.

### DIFF
--- a/src/mongocrypt.c
+++ b/src/mongocrypt.c
@@ -891,6 +891,14 @@ bool mongocrypt_init(mongocrypt_t *crypt) {
     return _try_enable_csfle(crypt);
 }
 
+bool mongocrypt_is_crypto_available(void) {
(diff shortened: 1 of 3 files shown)


jyemin (approved):
LGTM!

PR #739, opened 2024-01-25
PYTHON-4172 Support named KMS providers

ShaneHarvey:
https://jira.mongodb.org/browse/PYTHON-4172

This change extends the python-side kms provider validation rules to the new named providers. Note that we already pass kms_providers directly to libmongocrypt so this change isn't a blocker. All users need to do is ensure they have a libmongocrypt with [MONGOCRYPT-605](https://jira.mongodb.org/browse/MONGOCRYPT-605) installed (>= v1.9.0).

This change also fixes a bug in the python binding which blocks named kms providers from being used in create_data_key. The bug results in this `EncryptionError: unknown kms_provider: aws:name1` error:
```python
    return target.create_data_key(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shane/git/mongo-python-driver/pymongo/encryption.py", line 747, in create_data_key
    with _wrap_encryption_errors():
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/Users/shane/git/mongo-python-driver/pymongo/encryption.py", line 104, in _wrap_encryption_errors
    raise EncryptionError(exc) from None
pymongo.errors.EncryptionError: unknown kms_provider: aws:name1
```

### DIFF
--- a/bindings/python/CHANGELOG.rst
+++ b/bindings/python/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 1.9.0
+------------------------
+
+- Add support for named KMS providers like "local:name1".
+  This feature requires libmongocrypt >= 1.9.0.
+
 Changes in Version 1.8.0
 ------------------------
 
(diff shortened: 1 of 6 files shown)


blink1073 (approved):
LGTM!

PR #744, opened 2024-02-09
SERVER-85740 Extend libmongocrypt to support trimFactor passthrough

marksg07:
Extends the FLE2RangeInsertSpec and RangeOpts structures to support the passthrough and bounds checking of trim factor. Does not actually use the trim factor in the query or insert algos yet.

### DIFF
--- a/src/mc-rangeopts.c
+++ b/src/mc-rangeopts.c
@@ -18,33 +18,33 @@
 
 #include "mc-check-conversions-private.h"
 #include "mongocrypt-private.h"
-#include "mongocrypt-util-private.h" // mc_bson_type_to_string
+#include "mongocrypt-util-private.h"  // mc_bson_type_to_string
 
-#include <float.h> // DBL_MAX
+#include <float.h>  // DBL_MAX
 
 // Common logic for testing field name, tracking duplication, and presence.
-#define IF_FIELD(Name, ErrorPrefix)                                                                                    \
-    if (0 == strcmp(field, #Name)) {                                                                                   \
-        if (has_##Name) {                                                                                              \
-            CLIENT_ERR("%sUnexpected duplicate field '" #Name "'", ErrorPrefix);                                       \
-            return false;                                                                                              \
-        }                                                                                                              \
+#define IF_FIELD(Name, ErrorPrefix)                                              \
+    if (0 == strcmp(field, #Name)) {                                             \
+        if (has_##Name) {                                                        \
+            CLIENT_ERR("%sUnexpected duplicate field '" #Name "'", ErrorPrefix); \
+            return false;                                                        \
+        }                                                                        \
         has_##Name = true;
 
-#define END_IF_FIELD                                                                                                   \
-    continue;                                                                                                          \
+#define END_IF_FIELD \
+    continue;        \
     }
 
-#define CHECK_HAS(Name, ErrorPrefix)                                                                                   \
-    if (!has_##Name) {                                                                                                 \
-        CLIENT_ERR("%sMissing field '" #Name "'", ErrorPrefix);                                                        \
-        return false;                                                                                                  \
+#define CHECK_HAS(Name, ErrorPrefix)                            \
+    if (!has_##Name) {                                          \
+        CLIENT_ERR("%sMissing field '" #Name "'", ErrorPrefix); \
+        return false;                                           \
     }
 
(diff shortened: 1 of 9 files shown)


src/mc-rangeopts.c
```diff
@@ -324,6 +351,141 @@ bool mc_RangeOpts_appendMax(const mc_RangeOpts_t *ro,
...
+    // For each type, we use getTypeInfo to get the total number of values in the domain (-1)
+    // which tells us how many bits are needed to represent the whole domain.
+    // note - can't use a switch statement because of -Werror=switch-enum
+    if (valueType == BSON_TYPE_INT32) {
+        int32_t value = 0;
+        mc_optional_int32_t rmin = {false, 0}, rmax = {false, 0};
+        if (ro->min.set) {
+            value = bson_iter_int32(&ro->min.value);
+            rmin = OPT_I32(value);
+            rmax = OPT_I32(bson_iter_int32(&ro->max.value));
```
kevinAlbs, line 372:
```suggestion
            BSON_ASSERT (ro->max.set);
            rmax = OPT_I32(bson_iter_int32(&ro->max.value));
```

Suggest asserting `ro->max` is set before access. If `min` is set and `max` is not set, I expect an error will be returned before in `mc_RangeOpts_to_FLE2RangeInsertSpec` and `mc_RangeOpts_parse`.

Same suggestion applies to other accesses of `ro->max` in this function.

src/mc-rangeopts.c
```diff
@@ -324,6 +351,162 @@ bool mc_RangeOpts_appendMax(const mc_RangeOpts_t *ro,
...
+bool mc_RangeOpts_appendTrimFactor(const mc_RangeOpts_t *ro,
+                                   bson_type_t valueType,
+                                   const char *fieldName,
+                                   bson_t *out,
+                                   mongocrypt_status_t *status) {
+    BSON_ASSERT_PARAM(ro);
+    BSON_ASSERT_PARAM(fieldName);
+    BSON_ASSERT_PARAM(out);
+    BSON_ASSERT(status || true);
+
```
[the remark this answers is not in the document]
marksg07: What would be an acceptable way to do something like this with libmongocrypt? In the server, we have a feature flag which bounds the use of trimFactor, but I wasn't sure about what the equivalent would be in libmongocrypt.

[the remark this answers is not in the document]
kevinAlbs: I do not think there is an established pattern aside from `mongocrypt_setopt_fle2v2`. I expect it can be a private and short-lived flag. See https://github.com/mongodb/libmongocrypt/commit/68a41291bf9839dc1423ce77c0b97e6ca54c244f https://github.com/mongodb/libmongocrypt/commit/49109dca07e259ef2cfa8f7e5fc976f82280ee61 on [this branch](https://github.com/mongodb/libmongocrypt/compare/master...kevinAlbs:libmongocrypt:marksg07/server-85740?expand=1) for one possible way.

test/test-mc-rangeopts.c
```diff
@@ -104,15 +108,185 @@ static void test_mc_RangeOpts_to_FLE2RangeInsertSpec(_mongocrypt_tester_t *teste
...
+             "min" : {"$numberDecimal" : "0"},
+             "max" : {"$numberDecimal" : "100"},
+             "precision" : 1,
+             "sparsity" : {"$numberLong" : "1"}
+         }),
+         .v = RAW_STRING({"v" : {"$numberDecimal" : "0"}}),
+         .expectError = "Trim factor (10) must be less than the total number of bits (10) used to represent any "
+                        "element in the domain."},
+
+#if MONGOCRYPT_HAVE_DECIMAL128_SUPPORT
```
kevinAlbs, line 271: Move `#if MONGOCRYPT_HAVE_DECIMAL128_SUPPORT` above the "tf bound check passes correctly for decimal" test so decimal tests do not run when libmongocrypt is not built with decimal128 support.

marksg07: Good catch, changed

src/mc-rangeopts.c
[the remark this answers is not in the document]
marksg07: Thanks so much for this! I've merged those commits.

src/mc-rangeopts.c
marksg07: Fixed.

markbenvenuto (approved):

kevinAlbs (approved):

marksg07:
evergreen merge

PR #759, opened 2024-02-28
MONGOCRYPT-617 Use trim factor in the edges + mincover algorithms

marksg07:
Please recommend any tests that you think should be added; there were a few existing tests (such as `test-mongocrypt-ctx-encrypt`) which I wasn't able to figure out how to expand to fit trim factor tests in.

### DIFF
--- a/src/mc-fle2-encryption-placeholder-private.h
+++ b/src/mc-fle2-encryption-placeholder-private.h
@@ -50,6 +50,8 @@ typedef struct {
     // precision determines the number of digits after the decimal point for
     // floating point values.
     mc_optional_uint32_t precision;
+    // trimFactor determines how many root levels of the hypertree to trim.
+    mc_optional_uint32_t trimFactor;
 } mc_FLE2RangeFindSpecEdgesInfo_t;
 
 /** FLE2RangeFindSpec represents the range find specification that is encoded
@@ -74,7 +76,7 @@ typedef struct {
     mc_FLE2RangeOperator_t secondOperator;
 } mc_FLE2RangeFindSpec_t;
 
-bool mc_FLE2RangeFindSpec_parse(mc_FLE2RangeFindSpec_t *out, const bson_iter_t *in, mongocrypt_status_t *status);
+bool mc_FLE2RangeFindSpec_parse(mc_FLE2RangeFindSpec_t *out, const bson_iter_t *in, bool use_range_v2, mongocrypt_status_t *status);
 
 /** mc_FLE2RangeInsertSpec_t represents the range insert specification that is
  * encoded inside of a FLE2EncryptionPlaceholder. See
@@ -90,9 +92,11 @@ typedef struct {
     // precision determines the number of digits after the decimal point for
     // floating point values.
     mc_optional_uint32_t precision;
+    // trimFactor determines how many root levels of the hypertree to trim.
+    mc_optional_uint32_t trimFactor;
 } mc_FLE2RangeInsertSpec_t;
 
-bool mc_FLE2RangeInsertSpec_parse(mc_FLE2RangeInsertSpec_t *out, const bson_iter_t *in, mongocrypt_status_t *status);
+bool mc_FLE2RangeInsertSpec_parse(mc_FLE2RangeInsertSpec_t *out, const bson_iter_t *in, bool use_range_v2, mongocrypt_status_t *status);
 
 /** FLE2EncryptionPlaceholder implements Encryption BinData (subtype 6)
  * sub-subtype 0, the intent-to-encrypt mapping. Contains a value to encrypt and
--- a/src/mc-fle2-encryption-placeholder-private.h
+++ b/src/mc-fle2-encryption-placeholder-private.h
@@ -76,7 +76,10 @@ typedef struct {
     mc_FLE2RangeOperator_t secondOperator;
 } mc_FLE2RangeFindSpec_t;
 
-bool mc_FLE2RangeFindSpec_parse(mc_FLE2RangeFindSpec_t *out, const bson_iter_t *in, bool use_range_v2, mongocrypt_status_t *status);
+bool mc_FLE2RangeFindSpec_parse(mc_FLE2RangeFindSpec_t *out,
+                                const bson_iter_t *in,
+                                bool use_range_v2,
+                                mongocrypt_status_t *status);
 
 /** mc_FLE2RangeInsertSpec_t represents the range insert specification that is
  * encoded inside of a FLE2EncryptionPlaceholder. See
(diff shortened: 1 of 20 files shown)


src/mc-fle2-encryption-placeholder-private.h
```diff
@@ -90,9 +95,14 @@ typedef struct {
     // precision determines the number of digits after the decimal point for
     // floating point values.
     mc_optional_uint32_t precision;
+    // trimFactor determines how many root levels of the hypertree to trim.
```
markbenvenuto, line 98: nit - hypergraph

src/mc-fle2-encryption-placeholder.c
```diff
@@ -405,26 +437,50 @@ bool mc_FLE2RangeInsertSpec_parse(mc_FLE2RangeInsertSpec_t *out, const bson_iter
...
     CHECK_HAS(max)
     // Do not error if precision is not present. Precision optional and only
     // applies to double/decimal128.
+
+    if (out->trimFactor.set && !use_range_v2) {
+        // Once `use_range_v2` is default true, this block may be removed.
+        CLIENT_ERR("invalid FLE2RangeInsertSpec: trimFactor is not supported for QE range v1");
+        return false;
+    }
+    // Do not error if trim factor is not present. It is optional as it requires a feature flag,
```
markbenvenuto, line 482: It should be required once you land the server changes. There will never be a case where we don't send it to the server for range even when it is just 0.

src/mc-fle2-encryption-placeholder.c
```diff
@@ -282,13 +300,24 @@ static bool mc_FLE2RangeFindSpecEdgesInfo_parse(mc_FLE2RangeFindSpecEdgesInfo_t
     CHECK_HAS(indexMax)
     // Do not error if precision is not present. Precision optional and only
     // applies to double/decimal128.
+
+    if (out->trimFactor.set && !use_range_v2) {
+        // Once `use_range_v2` is default true, this block may be removed.
+        CLIENT_ERR("invalid FLE2RangeFindSpecEdgesInfo: 'trimFactor' is not supported for QE range v1");
+        return false;
+    }
+    // Do not error if trim factor is not present. It is optional as it requires a feature flag,
```
markbenvenuto, line 309: It should be required once you land the server changes. There will never be a case where we don't send it to the server for range even when it is just 0.

src/mc-rangeopts.c
```diff
@@ -191,7 +191,7 @@ bool mc_RangeOpts_parse(mc_RangeOpts_t *ro, const bson_t *in, bool use_range_v2,
     if (ro->trimFactor.set) {
         if (!use_range_v2) {
             // Once `use_range_v2` is default true, this block may be removed.
-            CLIENT_ERR("trimFactor is not supported for QE range v1");
+            CLIENT_ERR("%strimFactor is not supported for QE range v1", error_prefix);
```
markbenvenuto, line 194: Change it to be `%s: ...` for readability unless you can guarantee `error_prefix` always has a trailing space.

markbenvenuto (changes requested):

src/mc-fle2-encryption-placeholder.c
marksg07: Changed the logic such that:

- For parse_RangeOpts, trim factor is optional when use_range_v2 is enabled, and causes an error if present otherwise.
- For RangeOpts_to_InsertSpec, when use_range_v2 is enabled, if trim factor is not present, appends trim factor = 0 to the insert spec. If use_range_v2 is disabled, appends nothing.
- For parse_InsertSpec and parse_FindSpecEdgesInfo, trim factor is required when use_range_v2 is enabled, and causes an error if present otherwise.

Also changed the testing around to test edge cases of use_range_v2 + trim factor presence/absence more thoroughly.

src/mc-fle2-encryption-placeholder.c
marksg07: see above

src/mc-rangeopts.c
marksg07: Kinda got fed up with the inconsistencies in error prefixing and just added a new macro to do it, let me know what you think of the approach. Changed it in this file and fle2-encryption-placeholder.

src/mc-fle2-encryption-placeholder.c
```diff
@@ -58,7 +63,7 @@ bool mc_FLE2EncryptionPlaceholder_parse(mc_FLE2EncryptionPlaceholder_t *out,
 
     mc_FLE2EncryptionPlaceholder_init(out);
     if (!bson_validate(in, BSON_VALIDATE_NONE, NULL) || !bson_iter_init(&iter, in)) {
-        CLIENT_ERR("invalid BSON");
+        CLIENT_ERR_PREFIXED("invalid BSON");
```
kevinAlbs, line 66: The trailing comma appears to cause MSVC build errors. [Example](https://spruce.mongodb.com/task/libmongocrypt_windows_test_build_and_test_and_upload_patch_0fbf9a0cfb215998c5dbec621c49e554a6344420_65e0c16f3627e0800d1475d0_24_02_29_17_40_00/logs?execution=0):

> mc-fle2-encryption-placeholder.c(66): error C2059: syntax error: ')'

VS Code shows this expands to:
```
_mongocrypt_set_error(status, MONGOCRYPT_STATUS_ERROR_CLIENT, 1, "Error parsing FLE2EncryptionPlaceholder" ": " "invalid BSON",)
```

Suggest using `%s` for use of `CLIENT_ERR_PREFIXED` that only have one argument:
```suggestion
        CLIENT_ERR_PREFIXED("%s", "Invalid BSON");
```

src/mc-range-mincover-generator.template.h
```diff
@@ -131,11 +133,18 @@ static inline DECORATE_NAME(MinCoverGenerator)
         CLIENT_ERR("Sparsity must be > 0");
         return NULL;
     }
+    size_t maxlen = (size_t)BITS - DECORATE_NAME(mc_count_leading_zeros)(max);
+    if (trimFactor != 0 && trimFactor >= maxlen) {
+        CLIENT_ERR("Trim factor must be less than the number of bits (%ld) used to represent an element of the domain",
```
kevinAlbs, line 138:
```suggestion
        CLIENT_ERR("Trim factor must be less than the number of bits (%zu) used to represent an element of the domain",
```

src/mc-range-edge-generation.c
```diff
@@ -23,29 +23,42 @@
...
     BSON_ASSERT_PARAM(leaf);
     if (sparsity < 1) {
         CLIENT_ERR("sparsity must be 1 or larger");
         return NULL;
     }
+
+    const size_t leaf_len = strlen(leaf);
+    if (trimFactor != 0 && trimFactor >= leaf_len) {
+        // We append a total of leaf_len + 1 (for the root) - trimFactor edges. When this number is equal to 1, we
+        // degenerate into equality, which is not desired, so trimFactor must be less than leaf_len.
```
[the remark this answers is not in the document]
marksg07: This is true; however, even without setting the trim factor > 0 here, this pretty much ends up being a degenerate case since we only have the root node + leaves. I think that we should probably consider re-evaluating having a maximum value set for sparsity. I'll get some feedback on what people think on this. Thank you for pointing this out.

src/mc-range-mincover-generator.template.h
marksg07: Fixed

markbenvenuto (approved):

src/mc-fle2-encryption-placeholder.c
marksg07: So apparently there is a long-standing bug with the MSVC preprocessor which causes this incorrect trailing comma for certain cases of variadic expansion (searching "msvc variadic macro preprocessing bug" yields about 12 years of people complaining about this). One solution is to avoid using macros with only variadic arguments from other macros, which is why I've added the `fmt` leading argument to the CLIENT_ERR macro. This seems to fix the issue for Windows.

kevinAlbs (approved):

PR #762, opened 2024-02-29
MONGOCRYPT-622 publish Debian 12 packages

kevinAlbs:
This PR adds PPA publishing to Debian 12 (both arm64 and x86_64).
Development packages were uploaded in [this patch build](https://spruce.mongodb.com/version/65d8f2291e2d17c1f5afdae5/), and can be tested with Earthly:
```
./.evergreen/earthly.sh +test-deb-packages-from-ppa --env=deb12 --distro="debian bookworm"
```

Release packages are expected with the 1.9.1 release.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1241,6 +1241,7 @@ buildvariants:
     has_packages: true
     packager_distro: debian12
     packager_arch: x86_64
+    python: python3
   tasks:
   - build-and-test-and-upload
   - name: publish-packages
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1235,7 +1235,7 @@ buildvariants:
     distros:
     - ubuntu2004-small
 - name: debian12
-  display_name: "Debian 12.0"
+  display_name: "Debian 12"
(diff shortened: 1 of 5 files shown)


.evergreen/config.yml
```diff
@@ -1230,6 +1234,19 @@ buildvariants:
   - name: publish-packages
     distros:
     - ubuntu2004-small
+- name: debian12
+  display_name: "Debian 12.0"
```
rcsanchez97, line 1238:
```suggestion
  display_name: "Debian 12"
```

I think it would be a good idea if we didn't specify the minor version in the variant description, keeping it to the major version. Our evergreen hosts are actually running 12.1 right now, and the distinction isn't really meaningful in our case.

rcsanchez97 (approved):
LGTM.

I initially had suggestions to change "12.0" -> "12" in Earthfile. However, it seems like earthly might distinguish between the various minor releases in a meaningful way, so I dropped those suggestions. However, if possible, I think it benefits us to not have to drag around the minor versions for the Debian releases, since they are only rarely meaningful.

PR #766, opened 2024-03-05
MONGOCRYPT-621 Rename rangePreview to range

marksg07:
Related main repo changes: https://github.com/10gen/mongo/pull/19645
Related enterprise changes: https://github.com/10gen/mongo-enterprise-modules/pull/1994

The new "range" query/index type works with both range V1 and V2. When we are using range V2, we can't use "rangePreview" as either an index type or query type. 

For range V1, the "range" query type is compatible with the "rangePreview" index type; this is for the case where we are using an old collection with "rangePreview" query type since we want to encourage using the new name. The alternative behavior would be that when using range V1, we force users to always use "rangePreview". I thought this was undesirable if we want to treat that name as deprecated. Conversely, the "rangePreview" query type is incompatible with the "range" index type; in this case, we should always be using "range" queries.

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -23,6 +23,7 @@
 #include "mongocrypt-marking-private.h"
 #include "mongocrypt-traverse-util-private.h"
 #include "mongocrypt-util-private.h" // mc_iter_document_as_bson
+#include "mongocrypt.h"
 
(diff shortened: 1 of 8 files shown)


marksg07:
evergreen retry

markbenvenuto (approved):

kevinAlbs (approved):

PR #772, opened 2024-03-18
MONGOCRYPT-619 Extend EdgeTokenSetV2 to take new ECOC tokens

marksg07:
When using Range V2, for range queries, the encryptedTokens (marking: "p") field gets an extra byte denoting whether the token set is for a leaf edge.


markbenvenuto (approved):

sgolemon (approved):
Just optional comments...

test/test-mongocrypt-marking.c
```diff
@@ -896,7 +902,212 @@ static void test_mc_get_mincover_from_FLE2RangeFindSpec(_mongocrypt_tester_t *te
...
+    ASSERT(mongocrypt_status_ok(status));
+
+    mc_CollectionsLevel1Token_destroy(collectionsLevel1Token);
+    _mongocrypt_buffer_cleanup(&tokenKey);
+    mongocrypt_status_destroy(status);
+    return ecocToken;
+}
+
+static void test_mc_marking_to_ciphertext(_mongocrypt_tester_t *tester) {
+    // Test that whether range V2 is enabled or disabled, the ciphertext matches our expectations.
```
kevinAlbs, line 1065:
```suggestion
    if (!_aes_ctr_is_supported_by_os) {
        printf("Common Crypto with no CTR support detected. Skipping.");
        return;
    }
    
    // Test that whether range V2 is enabled or disabled, the ciphertext matches our expectations.
```

I expect this will resolve task failures on macOS 10.14. macOS 10.14 does not support CTR mode encryption (MONGOCRYPT-440).

kevinAlbs (commented):
Try `git merge master` to update the PR branch. That may pull in test fixes from: https://github.com/mongodb/libmongocrypt/commit/f4cef01e319b2aec028499d2d20d3bf19e23538b and resove tasks failing due to the error: `cp: cannot stat '../debian/*': No such file or directory`.

marksg07:
@kevinAlbs Thanks for the suggestions, I've implemented them and the patch looks free of errors so I'll merge.

PR #777, opened 2024-03-20
MONGOCRYPT-620 Update compact + cleanup passthroughs for Range V2

marksg07:
- When Range V2 is enabled, compact and cleanup on fields supporting range queries will be sent with the new AnchorPaddingTokenRoot passed through.
- When Range V2 is enabled, compact gets the encrypted field config passed through. 

Updated tests for these new features, and also to make sure that the anchor padding factor field is passed through successfully for compact.


kevinAlbs (approved):
LGTM with a case-insensitive comparison.

markbenvenuto (approved):
+1 Kevin's suggestions

PR #779, opened 2024-04-01
MONGOCRYPT-663 Add cmake option to enable Range V2

marksg07:
When the option is enabled, newly created mongocrypt_t objects will have use_range_v2 set to true. In testing, we add a new _mongocrypt_init_for_testing which disables this option before calling mongocrypt_init, which is needed because there are a multitude of tests which implicitly rely on the range V1 behavior (ex. by testing the actual encrypted payload blob against a specific encrypted payload blob). As part of MONGOCRYPT-661 (enabling this flag by default), we will audit these tests and decide which ones to update for the new behavior, which to leave as-is (while explicitly disabling range V2) and which to remove.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,11 @@ option (ENABLE_PIC
 )
 option (ENABLE_BUILD_FOR_PPA "Maintainer-only option for preparing PPA build" OFF)
 option (ENABLE_ONLINE_TESTS "Enable online tests and the csfle utility" ON)
+# TODO MONGOCRYPT-661 When range V2 is default, remove this option.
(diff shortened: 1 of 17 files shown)


markbenvenuto (approved):

kevinAlbs (approved):

marksg07:
Hi @kevinAlbs , would you have any insight on why the Node build targets are failing: https://parsley.mongodb.com/evergreen/libmongocrypt_ubuntu2004_64_build_and_test_node_patch_0d692524ee8c6a6c9bd6576d24adf87a11518dfa_660afa05aa15ba0006e6e069_24_04_01_18_16_39/0/task?bookmarks=0,1304 ?

PR #781, opened 2024-04-15
PYTHON-4364 Add async support to pymongocrypt

NoahStapp:
`crendentials.py` still does synchronous network IO. Adding an asynchronous version requires that we migrate away from `requests` to a library that has async support. We'd also need to create an asynchronous `mongocrypt_ctx`, which seems beyond the reach of the Python binding.


blink1073:
We could use https://github.com/aio-libs/aiohttp

blink1073:
Maybe we should use `unasync` here as well?  This PR seems like it is dancing around doing just that.

bindings/python/pymongocrypt/synchronous/state_machine.py
```diff
@@ -112,7 +112,7 @@ def run_state_machine(ctx, callback):
 
     :Parameters:
       - `ctx`: A :class:`MongoCryptContext`.
-      - `callback`: A :class:`MongoCryptCallback`.
+      - `callback`: A :class:`AsyncMongoCryptCallback`.
```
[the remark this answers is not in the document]
NoahStapp: Looks like `:class:AsyncMongoCryptCallback` is one token, so it doesn't get picked up by `unasync`.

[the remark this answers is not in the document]
blink1073: Can we add that full thing as a token replacement?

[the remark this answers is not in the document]
NoahStapp: I was mistaken about docstrings, the entire docstring is actually a single token. I guess we will need a docstring injector like we have in motor, then.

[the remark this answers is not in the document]
NoahStapp: I've added a very basic docstring synchronizer that searches lines instead of tokens for now.

bindings/python/pymongocrypt/asynchronous/mongocrypt.py
```diff
@@ -0,0 +1,683 @@
...
+          - `callback`: A :class:`MongoCryptCallback`.
+        """
+        self.__opts = options  # type: MongoCryptOptions
+        self.__callback = callback
+        self.__crypt = None
+
+        if not isinstance(options, MongoCryptOptions):
+            raise TypeError("options must be a MongoCryptOptions")
+
+        if not isinstance(callback, (MongoCryptCallback, AsyncMongoCryptCallback)):
```
blink1073, line 58: Shouldn't this be one or the other, and replaced?

bindings/python/pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
     "cffi>=1.12.0,<2",
     "cryptography>=40",
     "packaging>=21.0",
-    "requests<3.0.0",
+    "httpx<0.25.0",
```
blink1073, line 46: The current version is 0.27, does that one not work for us?

bindings/python/synchro.py
```diff
@@ -0,0 +1,49 @@
...
+
+from unasync import Rule, unasync_files
+
+replacements = {
+    "asynchronous": "synchronous",
+    "AsyncMongoCryptCallback": "MongoCryptCallback",
+    "AsyncExplicitEncrypter": "ExplicitEncrypter",
+    "AsyncAutoEncrypter": "AutoEncrypter",
+    "AsyncClient": "Client",
+    "aclose": "close"
```
blink1073, line 14: Is this actually used anywhere?

bindings/python/pyproject.toml
NoahStapp: 0.25.0 dropped support for Python 3.7, we need this until we upgrade to supporting >= 3.8.

bindings/python/synchro.py
NoahStapp: Which one?

blink1073: `aclose`

bindings/python/pyproject.toml
blink1073: We can let pip handle that for us: https://github.com/encode/httpx/blob/4b85e6c3898b94e686b427afd83138c87520b479/pyproject.toml#L9

bindings/python/synchro.py
NoahStapp: In `crendentials.py`:

```
    try:
        response = await client.get(url, headers=headers)
    except Exception as e:
        msg = "unable to retrieve GCP credentials: %s" % e
        raise MongoCryptError(msg)
    finally:
        await client.aclose()
```

blink1073: Ah, so an httpx API, got it, thanks!

bindings/python/pymongocrypt/asynchronous/mongocrypt.py
```diff
@@ -0,0 +1,682 @@
...
+        with MongoCryptBinaryIn(document) as binary:
+            if not lib.mongocrypt_ctx_mongo_feed(self.__ctx, binary.bin):
+                self._raise_from_status()
+
+    def complete_mongo_operation(self):
+        """Completes the mongo operation."""
+        if not lib.mongocrypt_ctx_mongo_done(self.__ctx):
+            self._raise_from_status()
+
+    async def ask_for_kms_credentials(self):
```
blink1073, line 325: I think we should make kms_providers a property, hoist this whole file up to the top level, and call `_ask_for_kms_credentials(ctx.kms_providers)`.

NoahStapp: A property? Can you explain what you mean? `_ask_for_kms_credentials` does network IO so it needs to have both async and sync versions.

blink1073: The only async aspect of this file is this method.  Instead, I'm suggesting we have `MongoCryptContext.kms_providers` and then instead of calling `ctx.ask_for_kms_credentials`, we call the appropriate `_ask_for_kms_credentials` with `ctx.kms_providers`.  Only the `_ask_for_kms_credentials` function needs to be async/syncified.

NoahStapp: Ahh makes sense. The only other part that differs is this line: 

```
        if not isinstance(callback, AsyncMongoCryptCallback):
            raise TypeError("callback must be a AsyncMongoCryptCallback")
```

If we're fine with that checking for both sync and async callbacks, we can do this change.

blink1073: Yeah I think that's fine.

blink1073 (approved):
Nice work!

bindings/python/pymongocrypt/asynchronous/auto_encrypter.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2019-present MongoDB, Inc.
```
ShaneHarvey, line 1: Here and in all the other added files:
2019 -> 2024

bindings/python/synchro.py
```diff
@@ -0,0 +1,48 @@
+from os import listdir
```
ShaneHarvey, line 1: Missing license.

bindings/python/pyproject.toml
```diff
@@ -43,7 +43,8 @@ dependencies = [
     "cffi>=1.12.0,<2",
     "cryptography>=40",
     "packaging>=21.0",
-    "requests<3.0.0",
+    "httpx",
+    "unasync"
```
ShaneHarvey, line 47: This is only a build-time dependency so should it be included here?

bindings/python/pymongocrypt/synchronous/state_machine.py
```diff
@@ -14,6 +14,7 @@
```
ShaneHarvey, line 14: These imports need to continue to work so that pymongo doesn't break:
```python
try:
    from pymongocrypt.auto_encrypter import AutoEncrypter  # type:ignore[import]
    from pymongocrypt.errors import MongoCryptError  # type:ignore[import]
    from pymongocrypt.explicit_encrypter import ExplicitEncrypter  # type:ignore[import]
    from pymongocrypt.mongocrypt import MongoCryptOptions  # type:ignore[import]
    from pymongocrypt.state_machine import MongoCryptCallback  # type:ignore[import]

    _HAVE_PYMONGOCRYPT = True
except ImportError:
    _HAVE_PYMONGOCRYPT = False
    MongoCryptCallback = object
```

ShaneHarvey (changes requested):

bindings/python/pymongocrypt/synchronous/credentials.py
```diff
@@ -72,11 +75,14 @@ def _get_azure_credentials():
...
     headers = { "Metadata": "true", "Accept": "application/json" }
+    client = httpx.Client()
     try:
-        response = requests.get(url, headers=headers)
+        response = client.get(url, headers=headers)
     except Exception as e:
         msg = "Failed to acquire IMDS access token: %s" % e
         raise MongoCryptError(msg)
+    finally:
+        client.close()
```
ShaneHarvey, line 85: We used to implicitly use request's process wide connection pooling but now we're opening a new connection for each request. Is this going to have a negative perf impact?

ShaneHarvey: This might not matter because libmongocrypt caches the creds so this method is called infrequently.

bindings/python/.evergreen/test.sh
```diff
@@ -82,6 +82,7 @@ for PYTHON_BINARY in "${PYTHONS[@]}"; do
     createvirtualenv $PYTHON_BINARY .venv
     python -m pip install --prefer-binary -r test-requirements.txt
     python -m pip install -v -e .
+    python synchro.py
```
ShaneHarvey, line 85: Could you open a new ticket to add a PR check to ensure we ran synchro.py before committing? It could even be part of pre-commit?

As part of that ticket we should be sure to remove this line otherwise we could be testing uncommitted code (eg when we forgot to run synchro.py before committing).

bindings/python/pymongocrypt/synchronous/state_machine.py
ShaneHarvey: Can you add a test case for this? EG:
```python
def test_pymongo_imports(self):
    from pymongocrypt.auto_encrypter import AutoEncrypter  # type:ignore[import]
    from pymongocrypt.errors import MongoCryptError  # type:ignore[import]
    from pymongocrypt.explicit_encrypter import ExplicitEncrypter  # type:ignore[import]
    from pymongocrypt.mongocrypt import MongoCryptOptions  # type:ignore[import]
    from pymongocrypt.state_machine import MongoCryptCallback  # type:ignore[import]

```

ShaneHarvey (commented):
Have you run the pymongo encryption tests with this change?

NoahStapp:
> Have you run the pymongo encryption tests with this change?

Yes, all passing locally

bindings/python/release.sh
```diff
@@ -38,16 +38,20 @@ function get_libmongocrypt() {
 }
 
 function build_wheel() {
+    python -m pip install unasync
     python -m pip install --upgrade pip build
     python -m build --wheel
     rm -rf build libmongocrypt pymongocrypt/*.so pymongocrypt/*.dll pymongocrypt/*.dylib
+    python synchro.py
```
ShaneHarvey, line 45: This seems wrong. Aren't we suppose to run synchro _before_ committing the code? So there should never be a need to run it here.

ShaneHarvey (approved):

blink1073:
Check-manifest is failing:

```
[2024/04/22 18:57:05.040] + check-manifest -v
[2024/04/22 18:57:05.040] Removing drivers-evergreen-tools/
[2024/04/22 18:57:05.040] lists of files in version control and sdist do not match!
[2024/04/22 18:57:05.040] listing source files under version control: 30 files and directories
[2024/04/22 18:57:05.040] missing from sdist:
[2024/04/22 18:57:05.040] building an sdist: pymongocrypt-1.9.2.dev0.tar.gz: 29 files and directories
[2024/04/22 18:57:05.040] copying source files to a temporary directory
[2024/04/22 18:57:05.040] building a clean sdist: pymongocrypt-1.9.2.dev0.tar.gz: 29 files and directories
[2024/04/22 18:57:05.040] suggested MANIFEST.in rules:
[2024/04/22 18:57:05.040]   include *.py
[2024/04/22 18:57:05.040]   synchro.py
```

I suggest we add an `exclude:` for it.

PR #782, opened 2024-04-16
MONGOCRYPT-625 address Coverity issues

kevinAlbs:
This PR addresses some issues identified by Coverity. See individual commits for the CID.

# Background & Motivation
There are many Coverity issues referencing uninitialized access of the `bson_iter_t::value` field:

```c
bson_iter_t iter1;
BSON_ASSERT (bson_iter_init (&iter1, &bson)); // Does not initialize `iter1.value`
bson_iter_t iter2 = iter1; // Assigns uninitialized `value`.
```

I expect this is OK in practice. I do not expect `bson_iter_t::value` is meant to be directly accessed. The `bson_iter_t::value` is initialized and returned in `bson_iter_value`. Regardless, this PR errs towards safety and zero-initializes the `bson_iter_t` where issues were reported.

Once https://github.com/mongodb/mongo-c-driver/pull/1587 is released in libbson, libmongocrypt can upgrade the libbson dependency to avoid this issue in the future.

### DIFF
--- a/kms-message/src/kms_gcp_request.c
+++ b/kms-message/src/kms_gcp_request.c
@@ -87,7 +87,7 @@ kms_gcp_request_oauth_new (const char *host,
       req->crypto.sign_ctx = opt->crypto.sign_ctx;
    }
 
-   jwt_signature = malloc (SIGNATURE_LEN);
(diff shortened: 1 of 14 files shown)


vector-of-bool (approved):
LGTM

PR #783, opened 2024-04-18
DRIVERS-1541 Retry KMS decrypt requests on transient errors

adriandole:
Retry GCP, AWS, and Azure requests that encounter transient errors.
Evergreen testing with the C driver implementation: https://spruce.mongodb.com/version/66ac145668f87200077d4ad0/tasks

Scope doc: https://docs.google.com/document/d/1JUVC2-Pe8mfpgemHyb4ihsIOzIe_QFQPnYDJV2idVY4/edit

### DIFF
--- a/test/test-mongocrypt-datakey.c
+++ b/test/test-mongocrypt-datakey.c
@@ -440,7 +440,7 @@ static void _test_create_datakey_with_retry(_mongocrypt_tester_t *tester) {
         ASSERT_OK(kms_ctx, ctx);
         // Expect no sleep is requested before any error.
         ASSERT_CMPINT64(mongocrypt_kms_ctx_usleep(kms_ctx), ==, 0);
-        // Feed a retryable HTTP error.
+        // Mark a network error.
         ASSERT_OK(mongocrypt_kms_ctx_fail(kms_ctx), kms_ctx);
         // Expect KMS request is returned again for a retry.
         kms_ctx = mongocrypt_ctx_next_kms_ctx(ctx);


src/mongocrypt-kms-ctx.c
```diff
@@ -50,6 +50,7 @@ typedef struct {
 #define SHA256_LEN 32
 #define DEFAULT_HTTPS_PORT "443"
 #define DEFAULT_KMIP_PORT "5696"
+static const int kms_max_attempts = 10;
```
kevinAlbs, line 53:
```suggestion
static const int kms_max_attempts = 3;
```

Suggest using a max of 3. I noted rationale in scope (avoid a long worst-case, agree with AWS defaults)

kevinAlbs (commented):
Leaving quick initial feedback. I suggest `git merge master` to pull in https://github.com/mongodb/libmongocrypt/commit/cb4d4faa08c0cdc1aeb26eee22c930dc20e103bc. That may resolve many Evergreen task failures. The other task failures (e.g. `clang-tidy`) may have been introduced in this PR.

src/mongocrypt-key-broker.c
```diff
@@ -817,6 +817,10 @@ mongocrypt_kms_ctx_t *_mongocrypt_key_broker_next_kms(_mongocrypt_key_broker_t *
         return NULL;
     }
 
+    if (kb->keys_returned->kms.should_retry) {
```
kevinAlbs, line 820: `keys_returned` is a linked list. I expect this only retries the first KMS request if there are multiple, and may need to loop:
```c
// Check if any requests need retry
for (key_returned_t *ptr = kb->keys_returned; ptr != NULL; ptr = ptr->next) {
    if (ptr->kms.should_retry) {
        ptr->kms.should_retry = false;
        return &ptr->kms;
    }
}
```

Here is a suggested test of multiple KMS requests: https://github.com/mongodb/libmongocrypt/commit/fbabc573219915d0997b44127bc23693773317ab#diff-70d771e7c35927baab41e7a4472474753d07fbb2cccf73640626663ad3a3b771R5095. Test data was generated with the `csfle` utility.

kevinAlbs (changes requested):

src/mongocrypt-kms-ctx.c
```diff
@@ -995,6 +1091,30 @@ static bool _ctx_done_kmip_decrypt(mongocrypt_kms_ctx_t *kms_ctx) {
...
+        return false;
+    }
+
+    if (kms->attempts >= kms_max_attempts) {
+        mongocrypt_status_t *status = kms->status;
+        CLIENT_ERR("KMS request failed after %d retries due to a network error", kms_max_attempts);
+        return false;
+    }
+
+    /* Not idempotent and can't be retried. */
```
kevinAlbs, line 1105: Suggest using an allow-list (rather than a block-list) to determine eligibility. This may help to prevent retrying non-idempotent request added in the future.

src/mongocrypt-kms-ctx.c
```diff
@@ -995,6 +1091,30 @@ static bool _ctx_done_kmip_decrypt(mongocrypt_kms_ctx_t *kms_ctx) {
     return ret;
 }
 
+bool mongocrypt_kms_ctx_fail(mongocrypt_kms_ctx_t *kms) {
```
kevinAlbs, line 1094: The proposed retry protocol requires retrying network errors immediately after calling `mongocrypt_kms_ctx_fail`. In pseudo-code:

```python
for kms_ctx in mongocrypt_ctx_next_kms_ctx (ctx):
    while True: # Loop to retry network errors.
        try:
            do_socket_operation (kms_ctx)
        except NetworkException as ne:
            if mongocrypt_kms_ctx_fail (kms_ctx):
                continue # Retry request immediately
```

Retryable HTTP errors require less special handling, as they are returned again in `mongocrypt_ctx_next_kms_ctx`.

Suggest changing `mongocrypt_kms_ctx_fail` to set `should_retry` to true. That may help simplify the driver integration. `mongocrypt_ctx_next_kms_ctx` will return the retryable KMS context, regardless of network or HTTP retry. 

Along with the suggestion for an allow-list, and having `mongocrypt_kms_ctx_fail` only return `false` on non-retryable error:

```c
bool mongocrypt_kms_ctx_fail(mongocrypt_kms_ctx_t *kms) {
    if (!kms) {
        return false;
    }

    mongocrypt_status_t *status = kms->status;

    if (!kms->retry_enabled) {
        CLIENT_ERR("KMS request failed due to network error");
        return false;
    }

    if (kms->attempts >= kms_max_attempts) {
        CLIENT_ERR("KMS request failed after %d retries due to a network error", kms_max_attempts);
        return false;
    }

    // Check if request type is retryable. Some requests are non-idempotent and cannot be safely retried.
    _kms_request_type_t retryable_types[] = {MONGOCRYPT_KMS_AWS_ENCRYPT,
                                             MONGOCRYPT_KMS_AWS_DECRYPT,
                                             MONGOCRYPT_KMS_AZURE_WRAPKEY,
                                             MONGOCRYPT_KMS_AZURE_UNWRAPKEY,
                                             MONGOCRYPT_KMS_GCP_ENCRYPT,
                                             MONGOCRYPT_KMS_GCP_DECRYPT};
    bool is_retryable = false;
    for (size_t i = 0; i < sizeof(retryable_types) / sizeof(retryable_types[0]); i++) {
        if (retryable_types[i] == kms->req_type) {
            is_retryable = true;
            break;
        }
    }
    if (!is_retryable) {
        CLIENT_ERR("KMS request failed due to network error");
        return false;
    }

    // Mark KMS context as retryable. Return again in `mongocrypt_ctx_next_kms_ctx`.
    set_retry(kms);

    // Reset intermediate state of parser.
    if (kms->parser) {
        kms_response_parser_reset(kms->parser);
    }
    return true;
}
```

src/mongocrypt-ctx-datakey.c
```diff
@@ -40,7 +40,7 @@ static mongocrypt_kms_ctx_t *_next_kms_ctx(mongocrypt_ctx_t *ctx) {
     BSON_ASSERT_PARAM(ctx);
 
     dkctx = (_mongocrypt_ctx_datakey_t *)ctx;
-    if (dkctx->kms_returned) {
+    if (!dkctx->kms.should_retry && dkctx->kms_returned) {
         return NULL;
     }
     dkctx->kms_returned = true;
```
kevinAlbs, line 46:
```suggestion
    dkctx->kms.should_retry = false; // Reset retry state.
    dkctx->kms_returned = true;
```



Suggest adding a test for creating a data key with retry:

```c
static void _test_creating_datakey_with_retry(_mongocrypt_tester_t *tester) {
    // Test that an HTTP error is retried.
    {
        mongocrypt_t *crypt = _mongocrypt_tester_mongocrypt(TESTER_MONGOCRYPT_DEFAULT);
        mongocrypt_ctx_t *ctx = mongocrypt_ctx_new(crypt);
        ASSERT_OK(
            mongocrypt_ctx_setopt_key_encryption_key(ctx,
                                                     TEST_BSON("{'provider': 'aws', 'key': 'foo', 'region': 'bar'}")),
            ctx);
        ASSERT_OK(mongocrypt_ctx_datakey_init(ctx), ctx);
        ASSERT_STATE_EQUAL(mongocrypt_ctx_state(ctx), MONGOCRYPT_CTX_NEED_KMS);
        mongocrypt_kms_ctx_t *kms_ctx = mongocrypt_ctx_next_kms_ctx(ctx);
        ASSERT_OK(kms_ctx, ctx);
        // Expect no sleep is requested before any error.
        ASSERT_CMPINT64(mongocrypt_kms_ctx_usleep(kms_ctx), ==, 0);
        // Feed a retryable HTTP error.
        ASSERT_OK(mongocrypt_kms_ctx_feed(kms_ctx, TEST_FILE("./test/data/rmd/kms-decrypt-reply-429.txt")), kms_ctx);
        // Expect KMS request is returned again for a retry.
        kms_ctx = mongocrypt_ctx_next_kms_ctx(ctx);
        ASSERT_OK(kms_ctx, ctx);
        // Feed a successful response.
        ASSERT_OK(mongocrypt_kms_ctx_feed(kms_ctx, TEST_FILE("./test/data/kms-aws/encrypt-response.txt")), kms_ctx);
        ASSERT_OK(mongocrypt_ctx_kms_done(ctx), ctx);
        _mongocrypt_tester_run_ctx_to(tester, ctx, MONGOCRYPT_CTX_DONE);
        mongocrypt_ctx_destroy(ctx);
    }

    // Test that a network error is retried.
    {
        mongocrypt_t *crypt = _mongocrypt_tester_mongocrypt(TESTER_MONGOCRYPT_DEFAULT);
        mongocrypt_ctx_t *ctx = mongocrypt_ctx_new(crypt);
        ASSERT_OK(
            mongocrypt_ctx_setopt_key_encryption_key(ctx,
                                                     TEST_BSON("{'provider': 'aws', 'key': 'foo', 'region': 'bar'}")),
            ctx);
        ASSERT_OK(mongocrypt_ctx_datakey_init(ctx), ctx);
        ASSERT_STATE_EQUAL(mongocrypt_ctx_state(ctx), MONGOCRYPT_CTX_NEED_KMS);
        mongocrypt_kms_ctx_t *kms_ctx = mongocrypt_ctx_next_kms_ctx(ctx);
        ASSERT_OK(kms_ctx, ctx);
        // Expect no sleep is requested before any error.
        ASSERT_CMPINT64(mongocrypt_kms_ctx_usleep(kms_ctx), ==, 0);
        // Feed a retryable HTTP error.
        ASSERT_OK(mongocrypt_kms_ctx_fail(kms_ctx), kms_ctx);
        // Expect KMS request is returned again for a retry.
        kms_ctx = mongocrypt_ctx_next_kms_ctx(ctx);
        ASSERT_OK(kms_ctx, ctx);
        // Feed a successful response.
        ASSERT_OK(mongocrypt_kms_ctx_feed(kms_ctx, TEST_FILE("./test/data/kms-aws/encrypt-response.txt")), kms_ctx);
        ASSERT_OK(mongocrypt_ctx_kms_done(ctx), ctx);
        _mongocrypt_tester_run_ctx_to(tester, ctx, MONGOCRYPT_CTX_DONE);
        mongocrypt_ctx_destroy(ctx);
    }
}
```

test/test-mongocrypt-datakey.c
```diff
@@ -399,6 +399,60 @@ static void _test_datakey_custom_key_material(_mongocrypt_tester_t *tester) {
...
+            mongocrypt_ctx_setopt_key_encryption_key(ctx,
+                                                     TEST_BSON("{'provider': 'aws', 'key': 'foo', 'region': 'bar'}")),
+            ctx);
+        ASSERT_OK(mongocrypt_ctx_datakey_init(ctx), ctx);
+        ASSERT_STATE_EQUAL(mongocrypt_ctx_state(ctx), MONGOCRYPT_CTX_NEED_KMS);
+        mongocrypt_kms_ctx_t *kms_ctx = mongocrypt_ctx_next_kms_ctx(ctx);
+        ASSERT_OK(kms_ctx, ctx);
+        // Expect no sleep is requested before any error.
+        ASSERT_CMPINT64(mongocrypt_kms_ctx_usleep(kms_ctx), ==, 0);
+        // Feed a retryable HTTP error.
```
kevinAlbs, line 443:
```suggestion
        // Mark a network error.
```

Sorry, that was my goof (from https://github.com/mongodb/libmongocrypt/pull/783#discussion_r1705927422).

kevinAlbs (approved):
LGTM

PR #790, opened 2024-04-22
MONGOCRYPT-668 Add pre-commit hooks to Python binding

NoahStapp:


blink1073:
We should run this in our test runner as part of this PR, no?

NoahStapp:
> We should run this in our test runner as part of this PR, no?

Whoops good catch, added.

blink1073:
`./.evergreen/test.sh: line 78: pre-commit: command not found`

blink1073:
We should skip `pre-commit` on Windows (`ERROR: Failed building wheel for shellcheck_py`).

NoahStapp:
> We should skip `pre-commit` on Windows (`ERROR: Failed building wheel for shellcheck_py`).

Is `shellcheck_py` known to not support Windows? The failure is due to the same `urllib` error as before for macOS 10.14.

blink1073:
> Is shellcheck_py known to not support Windows?

It has a wheel but it gets downloaded from source in pre-commit.  I think it is fine to only run pre-commit on *nix platforms.

blink1073 (approved):
LGTM!

PR #794, opened 2024-04-26
Pre-commit formatting for Python binding

NoahStapp:


.pre-commit-config.yaml
```diff
@@ -9,9 +9,10 @@ repos:
   - id: check-toml
   - id: check-yaml
   - id: check-json
-    exclude: .evergreen/atlas_data_lake/test.driverdata.json
+    exclude: '.evergreen/atlas_data_lake/test.driverdata.json'
```
blink1073, line 12: Let's revert changes to this file and `pyproject.toml` and make them part of the PR that adds `.git-ignore-revs`.

NoahStapp: The `pyproject.toml` changes are adding the ruff rules, wouldn't we want that in this PR?

blink1073: So that it doesn't get ignored as part of this commit.

NoahStapp: Ah right!

blink1073 (approved):
LGTM!

PR #799, opened 2024-05-01
Add GraalVM reachability metadata

stIncMale:
Now that I discovered there is a way to get metadata origins from the tracing agent, it has became simple to know what metadata should go where. This PR reflects that new knowledge.

JAVA-5407


jyemin (approved):
LGTM

PR #805, opened 2024-05-09
MONGOCRYPT-416 build on Alpine Linux 3.18 (via Earthly)

rcsanchez97:
This PR adds amd64 and arm64 builds of libmongocrypt for Alpine Linux 3.18 (a musl-based distro).

The approach uses the same `"tar and upload libmongocrypt libraries"` function in the Evergreen configuration that is used by other tasks. This way, consumers who want the Alpine artifacts can retrieve them based on the build variant name, as with artifacts built for other platforms and distros.

Evergreen patch build: https://spruce.mongodb.com/version/663d35d51c3e9b0007936faf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I confirmed that the artifacts produced by these tasks are for the correct architectures:

```
roberto@build01:~/Downloads$  find libmongocrypt-a* -name '*.so'
libmongocrypt-amd64/sharedbson/lib/libmongocrypt.so
libmongocrypt-amd64/sharedbson/lib/libkms_message.so
libmongocrypt-amd64/nocrypto/lib/libmongocrypt.so
libmongocrypt-amd64/nocrypto/lib/libkms_message.so
libmongocrypt-amd64/lib/libmongocrypt.so
libmongocrypt-amd64/lib/libkms_message.so
libmongocrypt-arm64/sharedbson/lib/libmongocrypt.so
libmongocrypt-arm64/sharedbson/lib/libkms_message.so
libmongocrypt-arm64/nocrypto/lib/libmongocrypt.so
libmongocrypt-arm64/nocrypto/lib/libkms_message.so
libmongocrypt-arm64/lib/libmongocrypt.so
libmongocrypt-arm64/lib/libkms_message.so
roberto@build01:~/Downloads$  file libmongocrypt-amd64/lib/libmongocrypt.so
libmongocrypt-amd64/lib/libmongocrypt.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=f087b766de9529b170f29d130da7a2988f79edf2, with debug_info, not stripped
roberto@build01:~/Downloads$  file libmongocrypt-arm64/lib/libmongocrypt.so
libmongocrypt-arm64/lib/libmongocrypt.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=f087b766de9529b170f29d130da7a2988f79edf2, with debug_info, not stripped
```

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -411,6 +411,17 @@ tasks:
   - func: "create packages and repos"
   - func: "upload packages and repos"
 
+- name: build-with-earthly
+  commands:
+  - func: "fetch source"
+  - func: "earthly"
+    vars:
+      args: +build --env=${earthly_env}
+  - func: "earthly"
+    vars:
+      args: --artifact +build/libmongocrypt-install ${workdir}/install
+  - func: "tar and upload libmongocrypt libraries"
+
 - name: build-deb-packages-with-earthly
   commands:
   - func: "fetch source"
@@ -1602,6 +1613,22 @@ buildvariants:
   tasks:
   - benchmark-java
   - benchmark-python
+
+- name: alpine-amd64-earthly
+  display_name: "Alpine Linux 3.18 amd64 (via Earthly)"
+  expansions:
+    earthly_env: alpine
+  tasks:
+  - name: build-with-earthly
+    run_on: ubuntu2204-small
+
+- name: alpine-arm64-earthly
+  display_name: "Alpine Linux 3.18 arm64 (via Earthly)"
+  expansions:
+    earthly_env: alpine
+  tasks:
+  - name: build-with-earthly
+    run_on: ubuntu2204-arm64-small
 
 - name: debian11-arm64-earthly
   display_name: "Debian 11 arm64 (via Earthly)"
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -411,6 +411,17 @@ tasks:
...
+  commands:
+  - func: "fetch source"
+  - func: "earthly"
+    vars:
+      args: +build --env=${earthly_env}
+  - func: "earthly"
+    vars:
+      args: --artifact +build/libmongocrypt-install ${workdir}/install
+  - func: "tar and upload libmongocrypt libraries"
+
```
kevinAlbs, line 424: Suggest updating the `upload-all` task to include the Alpine tasks.

The tarball uploaded in the `upload-all` task appears [referenced in Python bindings](https://github.com/mongodb/libmongocrypt/tree/6b328b46d2489526b7cc013a9b3d36eddfe3ab62/bindings/python#installing-libmongocrypt), and [consumed in C# bindings](https://github.com/mongodb/libmongocrypt/blob/6b328b46d2489526b7cc013a9b3d36eddfe3ab62/bindings/cs/Scripts/build.cake#L23).

Aside: eventually I would like to remove/revise the `upload-all` task. The tarball is ~2GB, but I expect consumers only need a subset of builds. The task depends on slow-to-schedule distros.

kevinAlbs (approved):
LGTM with an update to the `upload-all` task.

PR #806, opened 2024-05-10
MONGOCRYPT-685 minimize IntelDFP sources

kevinAlbs:
# Summary

Minimize the source included from IntelDFP.

Evergreen patch build: https://spruce.mongodb.com/version/663e1a3da99334000725025d

# Background & Motivation

This was originally intended to address Coverity issues related to IntelDFP. However, the Static code analysis requirements may not apply to third-party code (see [slack thread](https://mongodb.slack.com/archives/C05N744KWQ4/p1715344849758739)).

Regardless, it may help reduce unnecessary issues reported in Coverity. Removing the unnecessary source has the added benefit of reducing the binary size:

```bash
% du -h ./libmongocrypt-before/lib/libmongocrypt.so
 10M    ./libmongocrypt-before/lib/libmongocrypt.so
% du -h ./libmongocrypt-after/lib/libmongocrypt.so
6.0M    ./libmongocrypt-after/lib/libmongocrypt.so
```

To determine the minimal set of source files, this [Python script was used](https://gist.github.com/kevinAlbs/3fc04867c64aaf9f32284c8d0337de25).


rcsanchez97 (approved):
Wow, that is a lot of unused code (which adds substantially to the size of the compiled library).

LGTM!

PR #808, opened 2024-05-15
Fix KMIP operations reference in comment

addaleax:
Just a tiny comment fix, but this was apparently lying around in my local checkout of libmongocrypt for a couple months 🙂  

Refs: http://docs.oasis-open.org/kmip/spec/v1.0/os/kmip-spec-1.0-os.html#_Ref242030690


kevinAlbs (approved):
Thank you :)

PR #809, opened 2024-05-15
fix(NODE-6173): add rangeV2 feature flag toggle to Node.js bindings

addaleax:
See the ticket for motivation/why it would be convenient to have this earlier than the 'main' range V2 work 🙂

### DIFF
--- a/bindings/node/addon/mongocrypt.cc
+++ b/bindings/node/addon/mongocrypt.cc
@@ -514,6 +514,10 @@ MongoCrypt::MongoCrypt(const CallbackInfo& info)
         mongocrypt_setopt_bypass_query_analysis(_mongo_crypt.get());
     }
 
+    if (options.Get("rangeV2").ToBoolean()) {
+        mongocrypt_setopt_use_range_v2(_mongo_crypt.get());
+    }
+
     mongocrypt_setopt_use_need_kms_credentials_state(_mongo_crypt.get());
 
     // Initialize after all options are set.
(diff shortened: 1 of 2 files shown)


lerouxb (approved):

bindings/node/addon/mongocrypt.cc
```diff
@@ -514,6 +514,10 @@ MongoCrypt::MongoCrypt(const CallbackInfo& info)
         mongocrypt_setopt_bypass_query_analysis(_mongo_crypt.get());
     }
 
+    if (options.Get("rangeV2").ToBoolean()) {
```
kmruiz, line 517: What happens if rangeV2 is provided? Does it return a false? If it can return null an additional check would be nice.

addaleax: I don't think I _fully_ understand your question (probably a "not" missing somewhere?), but this is the equivalent of `if (!!options.rangeV2)` in JS – we do a similar check right above this one for `bypassQueryAnalysis`. This type of check is fine for boolean values, including optional ones.

addaleax: `Null` and `Undefined` are represented in Node-API as distinct values, not a C/C++ `nullptr`, so that's all good, and `.ToBoolean()` corresponds directly to the JS standard [`ToBoolean()`](https://tc39.es/ecma262/#sec-toboolean) operation 🙂

nbbeeken (approved):
It appears this option will just "pass through" for `ClientEncryption` but `AutoEncryption` would need to support it explicitly, will you be needing that?

addaleax:
@nbbeeken Yeah, not ideal for this that `AutoEncrypter` doesn't pass through unknown options, but this is something we can work around I'd say

PR #810, opened 2024-05-16
MONGOCRYPT-680 attach Augmented SBOM to release

kevinAlbs:
# Summary

Add an `sbom-download` target to download an Augmented SBOM. Add step to release to attach Augmented SBOM.

Partially resolves MONGOCRYPT-680.

# Background & Motivation

Addresses the requirement:

> Drivers MUST publish Augmented SBOM documents (produced by Silk processing SBOM Lite documents) alongside releases.

Related: I expect it is not needed to run `silkbomb upload`. Onboarding to [SilkBomb > Generate SBOM from purls](https://docs.devprod.prod.corp.mongodb.com/mms/python/src/sbom/silkbomb/docs/ONBOARDING/#generate-sbom-from-purls) notes:
> (Preferred) Commit the generated CycloneDx SBOM to source control. Silk will automatically get the latest version of your SBOM every night.

The SBOM Lite is committed to `etc/cyclonedx.sbom.json`.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -477,3 +477,24 @@ sbom-generate:
         --sbom-out cyclonedx.sbom.json
     # Save the result back to the host:
     SAVE ARTIFACT /s/cyclonedx.sbom.json AS LOCAL etc/cyclonedx.sbom.json
+
+# sbom-download:
+#   Download the Augmented SBOM file from Silk.
+#
+# See https://wiki.corp.mongodb.com/display/DRIVERS/Using+AWS+Secrets+Manager+to+Store+Testing+Secrets for instructions to get secrets from AWS Secrets Manager. Secrets are available under `drivers/libmongocrypt`.
+# See https://docs.devprod.prod.corp.mongodb.com/mms/python/src/sbom/silkbomb/ for documentation of silkbomb.
+sbom-download:
+    FROM artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0
+    # Alias the silkbom executable to a simpler name:
+    RUN ln -s /python/src/sbom/silkbomb/bin /usr/local/bin/silkbomb
+    WORKDIR /s
+    # Download the Augmented SBOM file:
+    RUN --secret silk_client_id --secret silk_client_secret \
+        SILK_CLIENT_ID=${silk_client_id} \
+        SILK_CLIENT_SECRET=${silk_client_secret} \
+        silkbomb download \
+        --sbom-out cyclonedx.augmented.sbom.json \
+        --silk-asset-group libmongocrypt
+    # Save the result back to the host:
+    SAVE ARTIFACT /s/cyclonedx.augmented.sbom.json AS LOCAL etc/cyclonedx.augmented.sbom.json
+    RUN echo "Augmented SBOM saved to etc/cyclonedx.augmented.sbom.json"
--- a/Earthfile
+++ b/Earthfile
@@ -493,7 +493,7 @@ sbom-download:
     FROM +silkbomb
     WORKDIR /s
     # Download the Augmented SBOM file:
-    RUN --secret silk_client_id --secret silk_client_secret \
+    RUN --no-cache --secret silk_client_id --secret silk_client_secret \
         SILK_CLIENT_ID=${silk_client_id} \
         SILK_CLIENT_SECRET=${silk_client_secret} \
         silkbomb download \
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -477,3 +477,24 @@ sbom-generate:
...
+
+# sbom-download:
+#   Download the Augmented SBOM file from Silk.
+#
+# See https://wiki.corp.mongodb.com/display/DRIVERS/Using+AWS+Secrets+Manager+to+Store+Testing+Secrets for instructions to get secrets from AWS Secrets Manager. Secrets are available under `drivers/libmongocrypt`.
+# See https://docs.devprod.prod.corp.mongodb.com/mms/python/src/sbom/silkbomb/ for documentation of silkbomb.
+sbom-download:
+    FROM artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0
+    # Alias the silkbom executable to a simpler name:
+    RUN ln -s /python/src/sbom/silkbomb/bin /usr/local/bin/silkbomb
```
vector-of-bool, line 489: Recommend: I suggest consolidating this and the above duplicate lines into a separate target `xyz` for obtaining the SilkBomb tool, then replacing this with `FROM +xyz`

Earthfile
```diff
@@ -477,3 +477,24 @@ sbom-generate:
...
+    # Download the Augmented SBOM file:
+    RUN --secret silk_client_id --secret silk_client_secret \
+        SILK_CLIENT_ID=${silk_client_id} \
+        SILK_CLIENT_SECRET=${silk_client_secret} \
+        silkbomb download \
+        --sbom-out cyclonedx.augmented.sbom.json \
+        --silk-asset-group libmongocrypt
+    # Save the result back to the host:
+    SAVE ARTIFACT /s/cyclonedx.augmented.sbom.json AS LOCAL etc/cyclonedx.augmented.sbom.json
+    RUN echo "Augmented SBOM saved to etc/cyclonedx.augmented.sbom.json"
```
vector-of-bool, line 500: Consider: Since this file is handled manually and not committed to git, I recommend parameterizing the output path, i.e. `ARG --required out`, then `SAVE ARTIFACT … AS LOCAL $out`, thus the usage becomes `earthly … +sbom-download --out=somewhere.json`, requiring the user to be cognizant of the resulting file without the `RUN echo …`, and the `.gitignore` can remain unmodified for this PR.

Earthfile
```diff
@@ -477,3 +477,24 @@ sbom-generate:
...
+#
+# See https://wiki.corp.mongodb.com/display/DRIVERS/Using+AWS+Secrets+Manager+to+Store+Testing+Secrets for instructions to get secrets from AWS Secrets Manager. Secrets are available under `drivers/libmongocrypt`.
+# See https://docs.devprod.prod.corp.mongodb.com/mms/python/src/sbom/silkbomb/ for documentation of silkbomb.
+sbom-download:
+    FROM artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0
+    # Alias the silkbom executable to a simpler name:
+    RUN ln -s /python/src/sbom/silkbomb/bin /usr/local/bin/silkbomb
+    WORKDIR /s
+    # Download the Augmented SBOM file:
+    RUN --secret silk_client_id --secret silk_client_secret \
```
vector-of-bool, line 492: Important: Since this depends on opaque external resources, use `--no-cache` to inhibit Earthly's caching. Otherwise it will always output the same file after executing the first time on a user's machine.

vector-of-bool (changes requested):
Two minor suggestions, but one important small change is needed to preventing bad caching.

vector-of-bool:
I find it unpleasant that this depends on an uncontrolled external resource, i.e. Silk ingests a file that is unrelated to the contents of the local source tree (presumably it would only pull from the `master` branch on a scheduled basis).

I don't think the SBOM for this project would change often, so it may be a non-issue here (and probably for C and C++ as well) — Just something that I think about.

Earthfile
kevinAlbs: Neat. Done.

Earthfile
kevinAlbs: Oof. Great catch. Updated.

Earthfile
kevinAlbs: I like that idea. Added `out` arg and reverted changes to `.gitignore`.

vector-of-bool (approved):
LGTM

PR #815, opened 2024-05-23
CSHARP-5100: Update Libmongocrypt cake build script

adelinowona:
Updating some things in the cake build script that were failing and making the build process frustrating.


BorisDog (approved):
LGTM

PR #818, opened 2024-05-28
MONGOCRYPT-685 update libbson to 1.27.1

kevinAlbs:
# Summary

Update bundled libbson to 1.27.1

Apply workarounds to resolve errors on RHEL 6.2 with gcc 4.4.7:
- Define `__STDC_LIMIT_MACROS` when compiling with C++.
- Disable GCC diagnostic macros in functions for gcc < 4.6.

Remove workarounds for older libbson:
- Replace `CONCAT` macro with `BSON_CONCAT`.
- Remove `BSON_SUBTYPE_ENCRYPTED` (now defined in libbson).
- Remove workaround for CDRIVER-3340.

Verified with this patch build: https://spruce.mongodb.com/version/6650b53f05094a0007522646

# Background

## Motivation

Updating bundled libbson is to address an issue reported in Skyk that was fixed in libbson 1.25.0: https://security.snyk.io/vuln/SNYK-UNMANAGED-MONGODBMONGOCDRIVER-6156642

## Workaround: define `__STDC_LIMIT_MACROS` for C++ targets

Defining `__STDC_LIMIT_MACROS` resolves an error on [this patch](https://parsley.mongodb.com/evergreen/libmongocrypt_rhel_62_64_bit_build_and_test_and_upload_patch_6b328b46d2489526b7cc013a9b3d36eddfe3ab62_6647a9b5a30c2b00076ff94f_24_05_17_19_02_14/0/task?bookmarks=0,705) when compiling a C++ test target: `bson-cmp.h:144: error: 'INT8_MIN' was not declared in this scope`. The stdint.h header on a spawned RHEL 6.2 host includes:
```c
/* The ISO C99 standard specifies that in C++ implementations these
   macros should only be defined if explicitly requested.  */
#if !defined __cplusplus || defined __STDC_LIMIT_MACROS
```

## Workaround: remove diagnostic pragma on old GCC

Removing the `GCC diagnostic` pragma resolves an error on [this patch](https://spruce.mongodb.com/task/libmongocrypt_rhel_62_64_bit_build_and_test_and_upload_patch_6b328b46d2489526b7cc013a9b3d36eddfe3ab62_664ea13f2a08f80007e8b58f_24_05_23_01_51_59/logs?execution=0): `mongoc-aggregate.c:97: error: #pragma GCC diagnostic not allowed inside functions`. I expect this is unavailable on GCC 4.6 and older. [GCC 4.6 changelog](https://gcc.gnu.org/gcc-4.6/changes.html) notes:
> Support for selectively enabling and disabling warnings via #pragma GCC diagnostic has been added

Due to the workarounds, MONGOCRYPT-688 was filed to propose dropping RHEL 6.2 support for libmongocrypt.

## Adding a runtime check

Removing the workaround of CDRIVER-3340 motivated adding a compile-time check to ensure libbson is newer than 1.16.0 (to include fix of CDRIVER-3340). The check for 1.16.0 is a precaution to ensure using a system install of libbson is 1.16.0 or newer. When libbson is loaded from the system, [a version check is not performed](https://github.com/mongodb/libmongocrypt/blob/7f396b84308ae0428bcdf4f55f9b387ecfe24981/cmake/ImportBSON.cmake#L110-L128).

### DIFF
--- a/cmake/ImportBSON.cmake
+++ b/cmake/ImportBSON.cmake
@@ -142,6 +142,11 @@ function (_import_bson)
       set (ENABLE_EXTRA_ALIGNMENT ${_extra_alignment_default} CACHE BOOL "Toggle extra alignment of bson_t")
       # We don't want the subproject to find libmongocrypt
       set (ENABLE_CLIENT_SIDE_ENCRYPTION OFF CACHE BOOL "Disable client-side encryption for the libmongoc subproject")
+      if (BUILD_VERSION)
+         # Both libmongocrypt and C driver support setting a `BUILD_VERSION`. Do not reuse the `BUILD_VERSION` for libmongocrypt when building the C driver.
+         # Set the `BUILD_VERSION` in this file to only apply to the C driver subproject.
+         set (BUILD_VERSION ${MONGOC_FETCH_TAG_FOR_LIBBSON})
+      endif()
       # Add the subdirectory as a project. EXCLUDE_FROM_ALL to inhibit building and installing of components unless requested
       # SYSTEM (on applicable CMake versions) to prevent warnings (particularly from -Wconversion/-Wsign-conversion) from the C driver code
       if (CMAKE_VERSION VERSION_GREATER 3.25)
--- a/cmake/ImportBSON.cmake
+++ b/cmake/ImportBSON.cmake
@@ -150,6 +150,8 @@ function (_import_bson)
       # Disable building tests in C driver:
       set (ENABLE_TESTS OFF)
       set (BUILD_TESTING OFF)
+      # Disable counters in C driver. Counters are not supported on all platforms.
+      set (ENABLE_SHM_COUNTERS OFF)
       # Add the subdirectory as a project. EXCLUDE_FROM_ALL to inhibit building and installing of components unless requested
       # SYSTEM (on applicable CMake versions) to prevent warnings (particularly from -Wconversion/-Wsign-conversion) from the C driver code
       if (CMAKE_VERSION VERSION_GREATER 3.25)
--- a/cmake/ImportBSON.cmake
+++ b/cmake/ImportBSON.cmake
@@ -142,11 +142,9 @@ function (_import_bson)
       set (ENABLE_EXTRA_ALIGNMENT ${_extra_alignment_default} CACHE BOOL "Toggle extra alignment of bson_t")
       # We don't want the subproject to find libmongocrypt
       set (ENABLE_CLIENT_SIDE_ENCRYPTION OFF CACHE BOOL "Disable client-side encryption for the libmongoc subproject")
-      if (BUILD_VERSION)
-         # Both libmongocrypt and C driver support setting a `BUILD_VERSION`. Do not reuse the `BUILD_VERSION` for libmongocrypt when building the C driver.
-         # Set the `BUILD_VERSION` in this file to only apply to the C driver subproject.
-         set (BUILD_VERSION ${MONGOC_FETCH_TAG_FOR_LIBBSON})
-      endif()
+      # Set `BUILD_VERSION` so C driver does not use a `BUILD_VERSION` meant for libmongocrypt.
+      # Both libmongocrypt and C driver support setting a `BUILD_VERSION` to override the version.
+      set (BUILD_VERSION ${MONGOC_FETCH_TAG_FOR_LIBBSON})
       # Disable building tests in C driver:
       set (ENABLE_TESTS OFF)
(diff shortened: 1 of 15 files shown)


rcsanchez97 (commented):
LGTM

cmake/ImportBSON.cmake
```diff
@@ -142,6 +142,16 @@ function (_import_bson)
       set (ENABLE_EXTRA_ALIGNMENT ${_extra_alignment_default} CACHE BOOL "Toggle extra alignment of bson_t")
       # We don't want the subproject to find libmongocrypt
       set (ENABLE_CLIENT_SIDE_ENCRYPTION OFF CACHE BOOL "Disable client-side encryption for the libmongoc subproject")
+      if (BUILD_VERSION)
+         # Both libmongocrypt and C driver support setting a `BUILD_VERSION`. Do not reuse the `BUILD_VERSION` for libmongocrypt when building the C driver.
+         # Set the `BUILD_VERSION` in this file to only apply to the C driver subproject.
+         set (BUILD_VERSION ${MONGOC_FETCH_TAG_FOR_LIBBSON})
+      endif()
```
vector-of-bool, line 149: Is there a reason to only set the build version if its already defined for the parent project? This could also be `unset(BUILD_VERSION)` to clear the value from the scope and let the sub-project infer its own value, or alternatively it can unconditionally set `BUILD_VERSION` to the fetch tag value.

cmake/Patch.cmake
vector-of-bool, line 1: If this is moved to a separate module, I'd prefer it more module-like so that it doesn't peek at variables in the includer's scope and behave different when included multiple times. For example:

```cmake
find_program(GIT_EXECUTABLE git)
find_program(PATCH_EXECUTABLE patch)

#[[
    Form a new Patch-applying command for the given inputs

    make_patch_command(
        <outvar>
        [DIRECTORY <dir>]
        [STRIP_COMPONENTS <N>]
        PATCHES [<file> ...]
    )
]]
function(make_patch_command out)
    cmake_parse_arguments(PARSE_ARGV 1 patch "" "DIRECTORY;STRIP_COMPONENTS" "PATCHES")
    if(GIT_COMMAND)
        # git ...
        set(cmd ${GIT_EXECUTABLE})

        if(patch_DIRECTORY)
            # git --work-tree=...
            list(APPEND cmd --work-tree=${patch_DIRECTORY})
        endif()
        # git ... apply ...
        list(APPEND cmd apply)
        # git ... apply -pN ...
        if(patch_STRIP_COMPONENTS)
            list(APPEND cmd -p${patch_STRIP_COMPONENTS})
        endif()
        # git ... apply ...
        foreach(p IN LISTS patch_PATCHES)
            # git ... apply ... <file>
            list(APPEND cmd ${p})
        endforeach()
    else()
        # patch ...
        set(cmd ${PATCH_EXECUTABLE})
        if(patch_DIRECTORY)
            # patch --dir=...
            list(APPEND cmd --dir=${patch_DIRECTORY})
        endif()
        # patch ... -pN ...
        if(patch_STRIP_COMPONENTS)
            list(APPEND cmd -p${patch_STRIP_COMPONENTS})
        endif()
        foreach(p IN LISTS patch_PATCHES)
            # patch ... ---input=<file>
            list(APPEND cmd --input=${p})
        endforeach()
    endif()
    set("${out}" "${cmd}" PARENT_SCOPE)
endfunction()
```

and then callers can use it via `make_patch_command`

cmake/ImportBSON.cmake
kevinAlbs: No. Updated to unconditionally set `BUILD_VERSION`.

cmake/Patch.cmake
kevinAlbs: I like it. Thank you for the suggestion! Updated.

vector-of-bool (approved):
LGTM. I discovered two simplifications to the patch command function.

rcsanchez97 (approved):
LGTM

cmake/Patch.cmake
```diff
@@ -1,14 +1,56 @@
...
+            list(APPEND cmd --dir=${patch_DIRECTORY})
+        endif()
+        # patch ... -pN ...
+        if(patch_STRIP_COMPONENTS)
+            list(APPEND cmd -p${patch_STRIP_COMPONENTS})
+        endif()
+        foreach(p IN LISTS patch_PATCHES)
+            # patch ... ---input=<file>
+            list(APPEND cmd --input=${p})
+        endforeach()
```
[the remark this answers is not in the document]
kevinAlbs: Neat, thank you.

PR #821, opened 2024-05-30
PYTHON-4466 Fix python executables used in EVG test

blink1073:


bindings/python/.evergreen/test.sh
```diff
@@ -29,14 +29,23 @@ if [ "Windows_NT" = "$OS" ]; then # Magic variable in cygwin
...
-    MACOS_VER=$(sw_vers -productVersion)
-    PYTHONS=(
-          "/Library/Frameworks/Python.framework/Versions/3.8/bin/python3"
-          "/Library/Frameworks/Python.framework/Versions/3.9/bin/python3"
-          "/Library/Frameworks/Python.framework/Versions/3.10/bin/python3"
-          "/Library/Frameworks/Python.framework/Versions/3.11/bin/python3"
-          "/Library/Frameworks/Python.framework/Versions/3.12/bin/python3"
-          )
+    MACOS_ARCH=$(uname -p)
+    if [[ $MACOS_ARCH =~ "arm" ]]; then
```
ShaneHarvey, line 33: We don't test on non-arm anymore so I think we can remove this if/else.

ShaneHarvey (changes requested):

bindings/python/.evergreen/test.sh
blink1073: Updated

ShaneHarvey (approved):

PR #830, opened 2024-06-12
Use native cryptography where available

jyemin:
* Only register crypto Java callbacks if mongocrypt_is_crypto_available returns false
* Package crypto-enabled shared libraries on Mac and Windows
* Continue to package crypto-disabled shared libraries on Linux due to OpenSSL incompatibilities. Applications can force use of a crypto-enabled shared library by manually installing it in the system path or else specify the path via jna.library.path system variable.

JAVA-5306

### DIFF
--- a/bindings/java/mongocrypt/build.gradle.kts
+++ b/bindings/java/mongocrypt/build.gradle.kts
@@ -139,7 +139,7 @@ tasks.register<Copy>("unzipJava") {
     outputs.upToDateWhen { false }
     from(tarTree(resources.gzip("${jnaDownloadsDir}/libmongocrypt-java.tar.gz")))
     include(jnaMapping.keys.flatMap {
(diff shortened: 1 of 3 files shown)


bindings/java/mongocrypt/build.gradle.kts
```diff
@@ -139,7 +139,7 @@ tasks.register<Copy>("unzipJava") {
     outputs.upToDateWhen { false }
     from(tarTree(resources.gzip("${jnaDownloadsDir}/libmongocrypt-java.tar.gz")))
     include(jnaMapping.keys.flatMap {
-        listOf("${it}/nocrypto/**/libmongocrypt.so", "${it}/nocrypto/**/libmongocrypt.dylib", "${it}/nocrypto/**/mongocrypt.dll" )
+        listOf("${it}/nocrypto/**/libmongocrypt.so", "${it}/lib/**/libmongocrypt.dylib", "${it}/lib/**/mongocrypt.dll" )
```
jyemin, line 142: Only use nocrypto variant on linux platforms

kevinAlbs (approved):

rozza (approved):
LGTM

PR #850, opened 2024-06-27
CSHARP-4889: Support linux distros using libdl.so.2

adelinowona:


bindings/cs/MongoDB.Libmongocrypt/LibraryLoader.cs
```diff
@@ -168,6 +168,19 @@ private class LinuxLibrary : ISharedLibraryLoader
...
+
+            static LinuxLibrary()
+            {
+                try
+                {
+                    Libdl1.dlerror();
+                    _use_libdl1 = true;
+                }
+                catch
+                {
```
BorisDog, line 181: minor: I'd add `_use_libdl1 = false` in catch, just to readability (also technically it's better to initialize the variable)

BorisDog (approved):
LGTM

bindings/cs/MongoDB.Libmongocrypt/LibraryLoader.cs
adelinowona: Done

PR #852, opened 2024-06-27
MONGOCRYPT-699 Omit `encryptionInformation` in compact if no range fields are used

kevinAlbs:
See MONGOCRYPT-699. Bug was discovered by a PHP driver test receiving an error running `compactStructuredEncryptionData` on non-range fields in server 7.0 after upgrading libmongocrypt:

```
Command 'compactStructuredEncryptionData' requires field 'encryptionInformation' when range fields are present
```

A drivers integration test for running compact with range fields is proposed in https://github.com/mongodb/specifications/pull/1605


blink1073 (approved):
LGTM!

PR #858, opened 2024-07-10
MONGOCRYPT-695 Remove upper bound assertions on FLE2 range precision

adriandole:
Removes maximum precision assertions.

Corresponding server PR:https://github.com/10gen/mongo/pull/24903

### DIFF
--- a/src/mc-range-encoding.c
+++ b/src/mc-range-encoding.c
@@ -225,7 +225,7 @@ bool mc_getTypeInfoDouble(mc_getTypeInfoDouble_args_t args,
         }
 
         if (args.precision.set && args.precision.value > INT32_MAX) {
-            CLIENT_ERR("Precision cannot be greater than %d, got %u", INT32_MAX, args.precision.value);
+            CLIENT_ERR("Precision cannot be greater than %" PRId32 ", got %" PRIu32, INT32_MAX, args.precision.value);
             return false;
         }
     }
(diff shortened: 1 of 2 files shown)


src/mc-range-encoding.c
```diff
@@ -223,6 +223,11 @@ bool mc_getTypeInfoDouble(mc_getTypeInfoDouble_args_t args,
                        args.value);
             return false;
         }
+
+        if (args.precision.set && args.precision.value > INT32_MAX) {
+            CLIENT_ERR("Precision cannot be greater than %d, got %u", INT32_MAX, args.precision.value);
+            return false;
+        }
```
kevinAlbs, line 230: Add same check to `mc_getTypeInfoDecimal128`.

kevinAlbs (commented):
The expected error for encoding double with precision 309 does not currently match expectations in https://github.com/10gen/mongo/pull/23926:

```c++
    // precisions greater than 308 (after conversion to uint) will result in infinity
    // when calculating 10^prc
    ASSERT_EIBB_OVERFLOW(1, 2, 1, 309, 13830554455654793216ULL);
```

However, https://github.com/10gen/mongo/pull/23926 shows merge conflicts. Considering changes SERVER-91571, I also expect an error. Out of caution, I suggest waiting for https://github.com/10gen/mongo/pull/23926 to be merged prior to merging to ensure both implementations match.

kevinAlbs (approved):

PR #861, opened 2024-07-12
MONGOCRYPT-661 enable "range" by default

kevinAlbs:
# Summary

This PR partially resolves MONGOCRYPT-661:

- Enable "range" protocol by default.
- Deprecate `mongocrypt_setopt_use_range_v2` as it is no longer necessary.

# Background & Motivation

Patch build: https://spruce.mongodb.com/version/66912f5f6df6a60007ec877e

As implemented, libmongocrypt does not currently support both "range" and "rangePreview" simultaneously. Instead, `mongocrypt_setopt_use_range_v2` controls whether "range" is supported. Server 8.0 drops the experimental "rangePreview". Drivers are expected to drop "rangePreview" as well.

I expect removing "rangePreview" in a minor release is OK. "rangePreview" was documented as unstable in libmongocrypt and [in drivers](https://jira.mongodb.org/browse/DRIVERS-2286) as "experimental only". For historical precedent: the unstable QEv1 protocol was updated to the incompatible QEv2 protocol in minor libmongocrypt release for [DRIVERS-2435](https://jira.mongodb.org/browse/DRIVERS-2435).

Documentation noting "range" as unstable is expected to be short-lived. Once remaining "range" protocol changes are completed, a subseqent PR is planned to document "range" as stable.

This PR includes updates to language bindings to fix tests and update documentation. Notably, Python's `enable_range_v2` is not modified. It may be OK to deprecate or remove as it is no-longer-needed (rangeV2 is always enabled).


ShaneHarvey:
> This PR includes updates to language bindings to fix tests and update documentation. Notably, Python's enable_range_v2 is not modified. It may be OK to deprecate or remove as it is no-longer-needed (rangeV2 is always enabled).

I added that so existing apps that use rangePreview on MongoDB <=7.0 wouldn't break when upgrading to the latest version of pymongocrypt. If the plan is to break "rangePreview" across the board then that you can go ahead and remove the enable_range_v2 argument and property. Users will just need to pin to an older version of pymongocrypt or upgrade to "range".

ShaneHarvey (approved):
Python changes LGTM.

vbabanin (approved):
Java changes LGTM!

rozza (approved):
Java code - LGTM

PR #867, opened 2024-07-18
MONGOCRYPT-698 Make trimFactor and sparsity optional

adriandole:


src/mc-range-mincover.c
```diff
@@ -20,7 +20,7 @@
 
 #include "mc-array-private.h"
 #include "mc-range-edge-generation-private.h" // mc_count_leading_zeros_u32
-#include "mc-range-encoding-private.h"        // mc_getTypeInfo32
+#include "mc-range-encoding-private.h"        // mc_getTypeInfo32, mc_trimFactorDefault
```
kevinAlbs, line 23:
```suggestion
#include "mc-range-encoding-private.h"        // mc_getTypeInfo32, trimFactorDefault
```

kevinAlbs (approved):

PR #869, opened 2024-07-19
MONGOCRYPT-712 Improve double handling with precision

markbenvenuto:
Min and max integer of a double is -2^53 and 2^53. Remember double has a separate sign bit unlike integers. Javascript's Number constants are not correct for doubles.

Depends on Shreyas' change to compute the right max value.


test/test-mc-range-encoding.c
```diff
@@ -470,6 +474,25 @@ static void _test_RangeTest_Encode_Double(_mongocrypt_tester_t *tester) {
...
+                           .precision = OPT_U32_C(0),
+                           .expect = 0,
+                           .expectMax = OPT_U64_C(36028797018963967)},
+                          {.value = DOUBLE_MIN_SAFE_INT + 1,
+                           .max = OPT_DOUBLE_C(DOUBLE_MAX_SAFE_INT),
+                           .min = OPT_DOUBLE_C(DOUBLE_MIN_SAFE_INT),
+                           .precision = OPT_U32_C(0),
+                           .expect = 1,
+                           .expectMax = OPT_U64_C(36028797018963967)},
+                         {.value = DOUBLE_MAX_SAFE_INT,
```
shreyaskalyan, line 490: Add test cases for value 0-10. Doesn't matter what the actual expect is (should be 9007199254740992-9007199254741002) but I want to know that they are all different values.

shreyaskalyan (changes requested):

test/test-mc-range-encoding.c
```diff
@@ -487,6 +487,42 @@ static void _test_RangeTest_Encode_Double(_mongocrypt_tester_t *tester) {
...
+                           .precision = OPT_U32_C(0),
+                           .expect = 9007199254740995,
+                           .expectMax = OPT_U64_C(36028797018963967)},
+                          {.value = 4,
+                           .max = OPT_DOUBLE_C(DOUBLE_MAX_SAFE_INT),
+                           .min = OPT_DOUBLE_C(DOUBLE_MIN_SAFE_INT),
+                           .precision = OPT_U32_C(0),
+                           .expect = 9007199254740996,
+                           .expectMax = OPT_U64_C(36028797018963967)},
+                          {.value = 5,
```
shreyaskalyan, line 520: After thinking this through a bit more, although we know that this is probably correct, we still need a few more test cases. We basically need to show that the imprecision that we see in the range `[2^53 < abs(n) < 2^63]` does not impact the precision in the range of `[-(2^53) < val < 2^53]`. To make that happen, we should probably run a few test cases.

We should test some random bounds where `min E [-2^53, -2^54, -2^55, ... -2^63]` and `max E [2^53, 2^54, 2^55, ... 2^63]` and show that when `value = [0, 1, 2, 3, 4, 5]`, we have `.expect = [max + value]`. 

I know that we've poured over the math, but I still have some qualms about us possibly losing precision here. I also have some general qualms about floating point imprecision in a fractional number, but I am not sure that those will ever really go away :(

shreyaskalyan (commented):
If my comment makes sense, please do incorporate it. I mean for it to be more of a discussion rather than a prescription.

test/test-mc-range-encoding.c
markbenvenuto: Doubles lose precision for their significant between 2^52 and 2^63. The lose precision at 2^(i-52) where i = bits. I generated test cases from the server code and added a few more test cases. One for each number between 52-62 with values from 0-4. I did not add 40 tests, just 10 because 40 would be a waste.

shreyaskalyan (approved):

src/mc-range-encoding.c
```diff
@@ -354,8 +354,9 @@ bool mc_getTypeInfoDouble(mc_getTypeInfoDouble_args_t args,
     if (use_precision_mode) {
         // Take a number of xxxx.ppppp and truncate it xxxx.ppp if precision = 3.
         // We do not change the digits before the decimal place.
-        double v_prime = trunc(args.value * exp10Double(args.precision.value)) / exp10Double(args.precision.value);
-        int64_t v_prime2 = (int64_t)((v_prime - args.min.value) * exp10Double(args.precision.value));
+        int64_t v_prime = (int64_t)(trunc(args.value * exp10Double(args.precision.value)));
+        int64_t scaled_min = (int64_t)(args.min.value * exp10Double(args.precision.value));
+        int64_t v_prime2 = v_prime - scaled_min;
```
erwee, line 359: I thought of another test case where this might fail:
```js
let min = 900719925474099.0; // <-- exactly representable as double
let max = 900719925474100.0; // <-- exactly representable
let val = 900719925474099.6; // <-- exactly representable
let other_val = 900719925474099.5; // <-- exactly representable
let prc = 1;

// multiplying each by 10^prc = 10, it follows that
let scaled_min = 9007199254740990; // <-- exactly representable
let scaled_max = 9007199254741000; // <-- exactly representable
let scaled_val = 9007199254740996; // <-- exactly representable
let scaled_other_val = 9007199254740996; // <-- 9007199254740995 is NOT exactly representable; rounded to +1
```
It follows that both `scaled_val` and `scaled_other_val` return the same `v_prime2` value of `6`,  even though they are unique inputs

markbenvenuto: Both `9007199254740995` and `9007199254740996` are the same number in double (both 9007199254740996). So once the precision calculation kicks in, we lose accuracy by scaling it up from `900719925474099.6` (error -0.025) and `900719925474099.5` (error 0). The precision calculation is always going to have error in it because it `double`. This is why this double math is so frustrating to me because it is "easy" to run into rounding.

See https://flop.evanau.dev/double-converter

erwee: I still think we need to assert that `scaled_max` does not exceed 2^53 (the rounding threshold), so that we can be sure that the value never runs into rounding when scaled by `10^prc`.

That way this equation will always be true before proceeding: `(value * 10^prc) <= scaled_max <= 2^53`

shreyaskalyan: Yea, I'm with Mark on this one. `other_val` cannot be expected to be a different value than `val` unfortunately, even if the range is so small. 

This will however have to be something that we have to be very careful to clarify in the documentation, and in how we teach users to use this feature. It may even be necessary for us to allow users to enable and disable the precision optimization mode, since as Erwin said it's counter intuitive for `900719925474099.5` and `900719925474099.6` to yield different values when they are representable as doubles.

erwee: Okay.. I thought about this some more, and I think my assertion on `scaled_max` being less than `2^53` is not enough to ensure there are no collisions when mapping two values within the range.  Looking at how floating point multiplication works, I can imagine how it's still possible for two distinct double values, `x` and `y`, to produce the same result when multiplied by a constant (`10^prc`), and for the result to still be less than `2^53`.  At any rate, I don't have the brainwaves right now to be able to prove (or refute) this, or provide an example.. Probably wise to consult the crypt team, though, right?

markbenvenuto: I think the assertion around 2^53 for precison mode is still useful. There are cases though that can produce imprecise results due to rounding even with it. Floating point does have rounding and some numbers cannot be represented precisely. 

When a user types `9007199254740995` or `9007199254740996` in C/C++, they think the numbers are different but they are really the same in doubles. I think we will just need to document to users that setting precision may be imprecise due to floating point rounding.

erwee (approved):

shreyaskalyan (approved):
Took another pass - LGTM stands

PR #880, opened 2024-08-20
clarify lifetime of returned `mongocrypt_kms_ctx_t`

kevinAlbs:
Motivated by question about expected lifetime of `mongocrypt_kms_ctx_t` with respect to `mongocrypt_ctx_t` ([see slack](https://mongodb.slack.com/archives/C0668RJBA0J/p1724089719861209)).

Calling `mongocrypt_ctx_destroy (ctx)` [frees](https://github.com/mongodb/libmongocrypt/blob/d05dbf3c0d6e03da8c0b25a6fd0214c1165f9214/src/mongocrypt-key-broker.c#L1117) associated `mongocrypt_kms_ctx`'s.

### DIFF
--- a/src/mongocrypt.h
+++ b/src/mongocrypt.h
@@ -1082,6 +1082,8 @@ typedef struct _mongocrypt_kms_ctx_t mongocrypt_kms_ctx_t;
  * If KMS handles are being handled synchronously, the driver can reuse the same
  * TLS socket to send HTTP requests and receive responses.
  *
+ * The returned KMS handle must not outlive `ctx`.
+ *
  * @param[in] ctx A @ref mongocrypt_ctx_t.
  * @returns a new @ref mongocrypt_kms_ctx_t or NULL.
  */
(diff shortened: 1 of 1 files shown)


src/mongocrypt.h
```diff
@@ -1082,6 +1082,8 @@ typedef struct _mongocrypt_kms_ctx_t mongocrypt_kms_ctx_t;
  * If KMS handles are being handled synchronously, the driver can reuse the same
  * TLS socket to send HTTP requests and receive responses.
  *
+ * The returned KMS handle must not outlive `ctx`.
```
addaleax, line 1085:
```suggestion
 * The returned KMS handle does not outlive `ctx`.
```

`must not` sounds like it's a requirement on the user/caller (i.e. they need to destroy the KMS handle before destroying the context), but it's not, it's just a description of the lifetime relationship between these objects that libmongocrypt already provides (right?)

kevinAlbs: Agreed. Thank you for the suggestion.

addaleax (approved):

PR #886, opened 2024-09-05
MONGOCRYPT-666 document installation via extrepo

rcsanchez97:


kevinAlbs (approved):
LGTM. Tested instructions with a docker container.

PR #890, opened 2024-10-03
CDRIVER-5702 remove use of `bson-cmp.h`

kevinAlbs:
# Summary

Remove use of `bson-cmp.h`. Partially resolves MONGOCRYPT-724.

Verified with [this patch build](https://spruce.mongodb.com/version/66fd4469a87c330007048826).

# Background & Motivation

CDRIVER-5703 plans to make libbson's `bson-cmp.h` private. libmongocrypt supports using a system install of libbson (via `USE_SHARED_LIBBSON`). I expect this is needed to support the Debian package [libmongocrypt-dev](https://packages.debian.org/trixie/libmongocrypt-dev), which [depends on libbson-dev](https://github.com/mongodb/libmongocrypt/blob/e98085b8e1de1803b3ceb827a03924cfb8985de3/debian/control#L11). I expect the Debian package cannot use private API of the libbson source through that package. Therefore, this PR adds a modified copy of `bson-cmp.h` to `mlib/cmp.h`.

### DIFF
--- a/src/mlib/cmp.h
+++ b/src/mlib/cmp.h
@@ -1,0 +1,173 @@
+/*
+ * Copyright 2009-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <bson/bson-prelude.h>
+
+
+#ifndef BSON_CMP_H
+#define BSON_CMP_H
+
+
+#include <bson/bson-compat.h> /* ssize_t */
+#include <bson/bson-macros.h> /* BSON_CONCAT */
+
+#include <limits.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+
+BSON_BEGIN_DECLS
+
+
+/* Based on the "Safe Integral Comparisons" proposal merged in C++20:
+ * http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0586r2.html
+ *
+ * Due to lack of type deduction in C, relational comparison functions (e.g.
+ * `cmp_less`) are defined in sets of four "functions" according to the
+ * signedness of each value argument, e.g.:
+ *  - bson_cmp_less_ss (signed-value, signed-value)
+ *  - bson_cmp_less_uu (unsigned-value, unsigned-value)
+ *  - bson_cmp_less_su (signed-value, unsigned-value)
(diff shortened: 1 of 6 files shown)


src/mlib/cmp.h
eramongodb, line 1: Suggest `src/mlib/cmp.h` -> `src/mc-cmp-private.h` and add a `src/mc-cmp.c` that contains the `ssize_t` static assertions from [test-bson-cmp.c](https://github.com/mongodb/mongo-c-driver/blob/c38b8b0855b558a900f225935ada2f6852313b3e/src/libbson/tests/test-bson-cmp.c#L219).

eramongodb (commented):
> A new pattern of internal-only headers seemed excessive.

I am in favor of making the header(s) fully internal when able.

Recommend also importing the [test cases](https://github.com/mongodb/mongo-c-driver/blob/master/src/libbson/tests/test-bson-cmp.c).

rcsanchez97 (approved):
The changes look good to me. I don't feel strongly about Ezra's suggestions, but I am also not at all opposed to them. That said, the actual option which points the libmongocrypt build at a system installation of the libmongoc/libbson is `-DMONGOCRYPT_MONGOC_DIR=USE-SYSTEM`. Technically, this can either be used with the static or the shared libmongoc/libbson. The `USE-SYSTEM` value is used by both the Debian package the RPM. I just wanted to mention it for the sake of correctness/completeness.

src/mlib/cmp.h
eramongodb: Recommendation to define the assertions in a source file is to avoid redundant re-evaluation of the assertion + noisy duplicate errors (in each source file that happens to include the header) on potential assertion failure. There is already precedent for static assertions in header files in `src/mc-range-edge-generation-private.h` however, so I do not mind its current placement in the private header.

test/test-mc-cmp.c
```diff
@@ -0,0 +1,304 @@
...
+// `test-mc-cmp.h` is a modified copy of `test-bson-cmp.c` from libbson 1.28.0.
+
+#include "test-mongocrypt.h"
+
+#include "mc-cmp-private.h"
+
+static void test_mc_cmp_equal(_mongocrypt_tester_t *tester) {
+    BSON_UNUSED(tester);
+
+    BSON_ASSERT(mc_cmp_equal_ss(0, 0));
```
eramongodb, line 26:
```suggestion
    ASSERT(mc_cmp_equal_ss(0, 0));
```
Use test assertion instead of contract assertion in tests.

test/test-mc-cmp.c
```diff
@@ -0,0 +1,304 @@
...
+ */
+
+// `test-mc-cmp.h` is a modified copy of `test-bson-cmp.c` from libbson 1.28.0.
+
+#include "test-mongocrypt.h"
+
+#include "mc-cmp-private.h"
+
+static void test_mc_cmp_equal(_mongocrypt_tester_t *tester) {
+    BSON_UNUSED(tester);
```
eramongodb, line 24: No other libmongocrypt code appears to use `BSON_UNUSED`. Consider `(void)tester;` instead.

eramongodb (approved):
Minor suggestions remaining; otherwise, LGTM.

rcsanchez97 (approved):

PR #893, opened 2024-10-08
Remove Java binding and build variants

vbabanin:
### Description
Following the migration and integration of the `mongocrypt Java binding` into the `mongo-java-driver` repository, this PR aims to remove the Java binding and any associated build variants from this repository.

For the time being, we will retain the `upload-java` task as it supports faster local builds in `the mongo-java-driver`. 

JAVA-5585


.evergreen/config.yml
```diff
@@ -188,32 +181,6 @@ functions:
       params:
         file: ${PYMONGO_DIR}/xunit-results/TEST-*.xml
 
-  "publish java":
```
rozza, line 191: @vbabanin don't we need this task as it creates the jar that we download?

kevinAlbs (approved):

rozza (approved):

kevinAlbs (approved):

PR #895, opened 2024-10-18
MONGOCRYPT-724 remove use of `bson_as_json`

kevinAlbs:
# Summary

Replace use of the soon-to-be-deprecated `bson_as_json` with `bson_as_relaxed_extended_json`.

# Background & Motivation

CDRIVER-5700 plans to deprecate `bson_as_json` due to producing Legacy Extended JSON by default. libmongocrypt's use of `bson_as_json` appears limited to logging and tests. Calls are updated to use the portable Relaxed Extended JSON format (rather than preserve the Legacy Extended JSON format).


vector-of-bool (approved):
LGTM

PR #896, opened 2024-10-18
MONGOCRYPT-728 Add FLE2IndexedRangeEncryptedValueV2 serializer

joshbsiegel:
# Summary
This PR adds `mc_FLE2IndexedRangeEncryptedValueV2_serialize` along with the appropriate testing integration. 

# Background
See [SPM-3042](https://jira.mongodb.org/browse/SPM-3042). This new function is necessary for [SERVER-95887](https://jira.mongodb.org/browse/SERVER-95887) as a part of eliminating duplicated code for FLE2 support.

# Metadata
Previously, metadata sent into the `FLE2IndexedRangeEncryptedValueV2` parser was ignored as `libmongocrypt` doesn't have a use for it. Now that it's necessary to reserialize the encrypted data for server use, a new field is added to `_mc_FLE2IndexedEncryptedValueV2_t` and the metadata is saved while parsing.

# Testing
After parsing the test's payload, we now send the parsing function output into the new serializer to ensure that the serialized output is the same as the initial payload.

### DIFF
--- a/src/mc-fle2-payload-iev-private-v2.h
+++ b/src/mc-fle2-payload-iev-private-v2.h
@@ -130,4 +130,18 @@ bool mc_FLE2IndexedRangeEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2_t 
                                                const _mongocrypt_buffer_t *buf,
                                                mongocrypt_status_t *status);
 
+/*
+ * Serializes an mc_FLE2IndexedEncryptedValueV2_t into a buffer.
+ *
+ * The serialized output follows the same layout as the input `buf` to
+ * mc_FLE2IndexedRangeEncryptedValueV2_parse, allowing for round-trip
+ * conversions between the serialized and parsed forms.
+ *
+ * Retuns an error if the input structure is not valid, or if the buffer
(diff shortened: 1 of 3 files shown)


joshbsiegel:
The Evergreen failures are consistent with what is seen in other PRs: https://github.com/mongodb/libmongocrypt/pull/895

joshbsiegel:
I see now that [MONGOCRYPT-508](https://jira.mongodb.org/browse/MONGOCRYPT-508) and [MONGOCRYPT-510](https://jira.mongodb.org/browse/MONGOCRYPT-510) are meant to add parsing and serialization of the tokens in FLE2IndexedRangeEncryptedValue. This PR simply parses and serializes the raw tokens to and from `metadata` in order to get the tests passing. These upcoming tickets will change this, although I'm happy to remove this process from this PR if preferred.

sgolemon-corp (changes requested):
Similar comments as the other two, but looking at the massive similarity and the dependence on the edge type, I think it might make sense to combine these three into a single PR and make the whole thing a bit more manageable in the process.

joshbsiegel:
Closed due to duplicated work. Combined changes in #899.

PR #898, opened 2024-10-21
MONGOCRYPT-508 add edge parsing to FLE2IndexedRangeEncryptedValueV2

joshbsiegel:
#  Summary
This PR modifies `mc_FLE2IndexedRangeEncryptedValueV2_parse` to parse edges into an array in `FLE2IndexedRangeEncryptedValueV2`.

# Background
See [SPM-3042](https://jira.mongodb.org/browse/SPM-3042). This capability is necessary to eliminate duplicated code for FLE2 support. It seems as if the suggestions mentioned in the body of the ticket are more geared towards V1 of the FLE2 implementation, so they are mostly ignored. 

# Testing
Two new fields are added to the JSON file used to run testing: `edge_count` and `edges` (array).  Similarly, `_mc_fle2_iev_v2_test` is modified to add these new fields and `_mc_fle2_iev_v2_test_parse` is updated to parse them accordingly.

To assert the values of `edge_count` and `edges`, getters are added to `mc-fle2-payload-iev-v2.c`.

> [!NOTE]
> This PR is branched off of https://github.com/mongodb/libmongocrypt/pull/896. The plan is to rebase once the other PR is merged.


src/mc-fle2-payload-iev-v2.c
```diff
@@ -429,9 +483,54 @@ bool mc_FLE2IndexedRangeEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2_t
     const uint64_t SEV_len = SEV_and_edges_len - edges_len;
     CHECK_AND_RETURN(mc_reader_read_buffer(&reader, &iev->ServerEncryptedValue, SEV_len, status));
 
-    // Ignore Metadata block.
-    BSON_ASSERT(mc_reader_get_remaining_length(&reader) == edges_len);
+    iev->metadata = (_mongocrypt_buffer_t *)malloc(iev->edge_count * sizeof(_mongocrypt_buffer_t));
```
joshbsiegel, line 486: I wasn't sure whether to allocate memory like this or to initialize the array to its max size.
``` C
_mongocrypt_buffer_t edges[255]; // max uint8_t
```

src/mc-fle2-payload-iev-private-v2.h
```diff
@@ -73,6 +73,13 @@ bool mc_FLE2IndexedEncryptedValueV2_add_K_Key(_mongocrypt_crypto_t *crypto,
 const _mongocrypt_buffer_t *mc_FLE2IndexedEncryptedValueV2_get_ClientValue(const mc_FLE2IndexedEncryptedValueV2_t *iev,
                                                                            mongocrypt_status_t *status);
 
+uint8_t mc_FLE2IndexedEncryptedValueV2_get_edge_count(const mc_FLE2IndexedEncryptedValueV2_t *iev,
+                                                      mongocrypt_status_t *status);
+
+const _mongocrypt_buffer_t *mc_FLE2IndexedEncryptedValueV2_get_edge(const mc_FLE2IndexedEncryptedValueV2_t *iev,
```
sgolemon-corp, line 79: See my [comment](https://github.com/mongodb/libmongocrypt/pull/899#discussion_r1815322727) on #899.  And yeah, looks like that type is reusable between these different structs, so probably its own header.

src/mc-fle2-payload-iev-v2.c
```diff
@@ -322,9 +325,60 @@ void mc_FLE2IndexedEncryptedValueV2_destroy(mc_FLE2IndexedEncryptedValueV2_t *ie
     _mongocrypt_buffer_cleanup(&iev->DecryptedServerEncryptedValue);
     _mongocrypt_buffer_cleanup(&iev->ServerEncryptedValue);
     _mongocrypt_buffer_cleanup(&iev->S_KeyId);
+    for (int i = 0; i < iev->edge_count; i++) {
+        _mongocrypt_buffer_cleanup(&iev->metadata[i]);
+    }
+
+    // Metadata array is dynamically allocated
+    free(iev->metadata);
```
sgolemon-corp, line 333:
```suggestion
    bson_free(iev->metadata);
```

src/mc-fle2-payload-iev-v2.c
sgolemon-corp: Initializing to a large, but still arbitrary size is definitely not the right move.  I think having a pointer with its own array allocation is fine here. There is a neat trick you can do with C structs where an unsized array dangles off the end, but it's not one that'll work well here since we don't know the size till parsing starts and we don't want the caller to have to think about allocation sizing.

TL;DR - I'm okay with this plan,

But... per my comment on #899, you should be using `bson_malloc0` here.

sgolemon-corp (changes requested):

joshbsiegel:
Closed due to duplicated work. Combined changes in #899.

PR #899, opened 2024-10-21
MONGOCRYPT-510 add FLE2IndexedEncryptedValueV2 serializer

joshbsiegel:
# Summary
This PR implements `mc_FLE2IndexedEncryptedValueV2_serialize` and adds necessary functionality to support edges in libmongocrypt.

# Background
See [SPM-3042](https://jira.mongodb.org/browse/SPM-3042). This capability is necessary to eliminate duplicated code for FLE2 support. 

# `mc_FLE2TagAndEncryptedMetadataBlock_t`
In order to properly serialize an IEV, it's necessary to include the metadata in the serialization. Previously, libmongocrypt ignored all metadata (where the edges are stored for range IEVs) when parsing IEVs as they didn't have a use elsewhere in the codebase, only in SERVER. Now that FLE2 support is being consolidated to this repo, metadata must be accounted for.

A new data type and appropriate files are added to support metadata blocks. Simple functionality is implemented to `init`, `cleanup`, `serialize`, and `parse` this struct. A test is added to check roundtrip functionality using these functions. 

# v2 Payload Changes
Being that the main difference between equality and range IEVs is the metadata format, `mc_FLE2IndexedEncryptedValueV2_parse` was consolidated to account for both types instead of being a wrapper around separate parsing functions. This is done by setting `edge_count` of equality IEVs to 1. Functionally, an equality IEV is just a range IEV with only 1 "edge" (metadata block).

`mc_FLE2IndexedEncryptedValueV2_serialize` is implemented by taking `mc_FLE2IndexedEncryptedValueV2_parse` and performing the operations in reverse. Note while `edge_count` is set for equality IEVs, it is not serialized as this field is not technically present for this subtype.

Three helper functions were added to support testing and future SERVER needs: `_get_edge_count`, `get_edge` (only for range IEVs), and `get_metadata` (only for equality IEVs).

To test these changes, the parsed input is checked for correct edge count (range IEVs only) and metadata content. This requires modifications to the JSON files to add metadata blocks as a field. Along with that, the parsed input is immediately reserialized to ensure that the serialized output is the same as the unparsed input (asserting safe roundtrips).

# What's New
- Metadata/edge helper functions
- `mc_FLE2IndexedEncryptedValueV2_parse` consolidation
- `mc_FLE2IndexedEncryptedValueV2_serialize`
- Accounting for metadata in `mc-fle2-payload-iev-v2`
- `mc_FLE2TagAndEncryptedMetadataBlock_t` functionality
- Appropriate testing for new features


src/mc-fle2-payload-iev-private-v2.h
```diff
@@ -73,6 +73,13 @@ bool mc_FLE2IndexedEncryptedValueV2_add_K_Key(_mongocrypt_crypto_t *crypto,
 const _mongocrypt_buffer_t *mc_FLE2IndexedEncryptedValueV2_get_ClientValue(const mc_FLE2IndexedEncryptedValueV2_t *iev,
                                                                            mongocrypt_status_t *status);
 
+uint8_t mc_FLE2IndexedEncryptedValueV2_get_edge_count(const mc_FLE2IndexedEncryptedValueV2_t *iev,
+                                                      mongocrypt_status_t *status);
+
+const _mongocrypt_buffer_t *mc_FLE2IndexedEncryptedValueV2_get_edge(const mc_FLE2IndexedEncryptedValueV2_t *iev,
```
sgolemon-corp, line 79: Ideally, I'd like the parser to pull out the edge components individually into a sub-struct, so something like:

```c
typedef struct _mc_FLE2EncryptedValueV2_Edge_t {
    _mongocrypt_buffer_t encryptedCount;
    _mongocrypt_buffer_t tag;
    _mongocrypt_buffer_t encryptedZeros;
} mc_FLE2EncryptedValueV2_Edge_t;
```

These are each 32 bytes (of the 96 byte metadatalen total).

This should also return by out-param rather than return value to maintain consistency with other mongocrypt APIs, e.g.

```c
bool mc_FLE2IndexedEncryptionValueV2_get_edge(const mc_FLE2IndexedEncryptedValueV2_t *iev,
            mc_FLE2EncryptedValueV2_Edge_t* out,
            const uint8_t edge_index,
            mongocrypt_status_t *status);
```

The struct might need to be in its own header with `_init` and `_cleanup` methods matching the style of other types.

src/mc-fle2-payload-iev-v2.c
```diff
@@ -322,9 +325,60 @@ void mc_FLE2IndexedEncryptedValueV2_destroy(mc_FLE2IndexedEncryptedValueV2_t *ie
     _mongocrypt_buffer_cleanup(&iev->DecryptedServerEncryptedValue);
     _mongocrypt_buffer_cleanup(&iev->ServerEncryptedValue);
     _mongocrypt_buffer_cleanup(&iev->S_KeyId);
+    for (int i = 0; i < iev->edge_count; i++) {
+        _mongocrypt_buffer_cleanup(&iev->metadata[i]);
+    }
+
+    // Metadata array is dynamically allocated
+    free(iev->metadata);
```
sgolemon-corp, line 333:
```suggestion
    bson_free(iev->metadata);
```

src/mc-fle2-payload-iev-v2.c
```diff
@@ -429,9 +483,55 @@ bool mc_FLE2IndexedRangeEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2_t
     const uint64_t SEV_len = SEV_and_edges_len - edges_len;
     CHECK_AND_RETURN(mc_reader_read_buffer(&reader, &iev->ServerEncryptedValue, SEV_len, status));
 
-    // Ignore Metadata block.
-    BSON_ASSERT(mc_reader_get_remaining_length(&reader) == edges_len);
+    iev->metadata = (_mongocrypt_buffer_t *)malloc(iev->edge_count * sizeof(_mongocrypt_buffer_t));
```
sgolemon-corp, line 486:
```suggestion
    iev->metadata = (_mongocrypt_buffer_t *)bson_malloc0(iev->edge_count * sizeof(_mongocrypt_buffer_t));
```

`bson_` because it keeps it in the correct allocator buckets.
`0` so that it zeroes out the memory it's allocated (similar to `calloc`)

sgolemon-corp (changes requested):

src/mc-fle2-tag-and-encrypted-metadata-block-private.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "mc-reader-private.h"
+#include "mc-writer-private.h"
+#include "mongocrypt-private.h"
+
+typedef struct {
+    _mongocrypt_buffer_t encryptedCount;
+    _mongocrypt_buffer_t tag;
+    _mongocrypt_buffer_t encryptedZeros;
+} mc_FLE2TagAndEncryptedMetadataBlock_t;
```
joshbsiegel, line 28: I named the struct `mc_FLE2TagAndEncryptedMetadataBlock_t` to be consistent with comments I saw in https://github.com/mongodb/libmongocrypt/blob/4cad289ab696339680c90fbab996a456624f517a/src/mc-fle2-payload-iev-private-v2.h#L86.

sgolemon-corp: Good call. I like.

src/mc-fle2-tag-and-encrypted-metadata-block-private.h
```diff
@@ -0,0 +1,44 @@
...
+
+#define kFieldLen 32U
+
+void mc_FLE2TagAndEncryptedMetadataBlock_init(mc_FLE2TagAndEncryptedMetadataBlock_t *metadata);
+
+void mc_FLE2TagAndEncryptedMetadataBlock_cleanup(mc_FLE2TagAndEncryptedMetadataBlock_t *metadata);
+
+bool mc_FLE2TagAndEncryptedMetadataBlock_parse(mc_FLE2TagAndEncryptedMetadataBlock_t *metadata,
+                                               mc_reader_t *reader,
```
sgolemon-corp, line 37: My inclination is, even though it's less efficient, the payload parse should read into a buffer and pass a const buffer into this parse.  The extra copy is small (96) bytes compared to the overall crypto op and it avoids spilling a random reader out.

Same on the serialize end...

@kevinAlbs wdyt?

src/mc-fle2-payload-iev-v2.c
```diff
@@ -322,9 +325,93 @@ void mc_FLE2IndexedEncryptedValueV2_destroy(mc_FLE2IndexedEncryptedValueV2_t *ie
     _mongocrypt_buffer_cleanup(&iev->DecryptedServerEncryptedValue);
     _mongocrypt_buffer_cleanup(&iev->ServerEncryptedValue);
     _mongocrypt_buffer_cleanup(&iev->S_KeyId);
+
+    if (iev->type != kTypeRange) {
+        for (int i = 0; i < iev->edge_count; i++) {
+            mc_FLE2TagAndEncryptedMetadataBlock_cleanup(&iev->metadata[i]);
+        }
+    } else {
+        // kTypeEquality only has one metadata block
```
sgolemon-corp, line 334: `assert(iev->edge_count == 1);`

src/mc-fle2-payload-iev-v2.c
```diff
@@ -322,9 +325,93 @@ void mc_FLE2IndexedEncryptedValueV2_destroy(mc_FLE2IndexedEncryptedValueV2_t *ie
...
+    if (iev->type == kTypeInit) {
+        CLIENT_ERR("mc_FLE2IndexedEncryptedValueV2_get_edge_count "
+                   "must be called after "
+                   "mc_FLE2IndexedEncryptedValueV2_parse");
+        return 0;
+    }
+
+    if (iev->type != kTypeRange) {
+        CLIENT_ERR("mc_FLE2IndexedEncryptedValueV2_get_edge_count must be called with type range");
+        return 0;
```
sgolemon-corp, line 357: Equality has precisely one edge, so what if we simply set `iev->edge_count = 1;` in the parse method for the equality version and then let these methods treat range and equality the same?

src/mc-fle2-payload-iev-private-v2.h
```diff
@@ -122,12 +133,24 @@ bool mc_FLE2IndexedEqualityEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2
...
+ * Serializes an mc_FLE2IndexedEncryptedValueV2_t into a buffer.
+ *
+ * The serialized output follows the same layout as the input `buf` to
+ * mc_FLE2IndexedRangeEncryptedValueV2_parse, allowing for round-trip
+ * conversions between the serialized and parsed forms.
+ *
+ * Retuns an error if the input structure is not valid, or if the buffer
+ * provided is insufficient to hold the serialized data.
+ */
+bool mc_FLE2IndexedRangeEncryptedValueV2_serialize(const mc_FLE2IndexedEncryptedValueV2_t *iev,
```
sgolemon-corp, line 152: Let's add `mc_FLE2IndexedEqualityEncrytedValueV2_serialize()` while we're here.

test/test-mc-fle2-payload-iev-v2.c
```diff
@@ -83,6 +90,37 @@ static bool _mc_fle2_iev_v2_test_parse(_mc_fle2_iev_v2_test *test, bson_iter_t *
                 TEST_ERROR("Unknown type '%s'", value);
             }
             hasType = true;
+        } else if (!strcmp(field, "edge_count")) {
```
sgolemon-corp, line 93: This can probably be implicit since we know each edge is precisely 96 bytes.

So:
```
assert((tmp_buf.len % 96) == 0);
test->edge_count = tmp_buf.len / 96;
```

in the "edges" processing down below.

sgolemon-corp (changes requested):
This looks fantastic overall. Just a few small comments and a question for Kevin.

src/mc-fle2-payload-iev-v2.c
joshbsiegel: Addressed with consolidation of functions.

src/mc-fle2-payload-iev-private-v2.h
joshbsiegel: Addressed with the consolidation of functions.

Does this address [MONGOCRYPT-608](https://jira.mongodb.org/browse/MONGOCRYPT-608)? I'm not certain because the `_write` function in v1 takes each individual `mc_FLE2IndexedEncryptedValue_t` field and serializes it to a buffer (vs. the v2 `_serialize` function which does the same but from a `mc_FLE2IndexedEncryptedValueV2_t` struct passed through to the function).

Functionally, these two methods do the same thing but the `_serialize` function requires the user to create a `mc_FLE2IndexedEncryptedValueV2_t` object before calling the function. I'm unsure on what will be easiest from the `SERVER` perspective and will defer to you.

Honestly, the `_write` function implementation might be as simple as creating a `mc_FLE2IndexedEncryptedValueV2_t` object with the parameters and then calling `_serialize` (?).

src/mc-fle2-tag-and-encrypted-metadata-block-private.h
joshbsiegel: I went ahead and pushed these changes but put them in a single [commit](https://github.com/mongodb/libmongocrypt/pull/899/commits/b0e0fb2d11dea522bf43540bf6448c767b29b606). If we decide to not go ahead with this then it will be an easy revert.

src/mc-fle2-payload-iev-v2.c
```diff
@@ -491,36 +404,49 @@ bool mc_FLE2IndexedRangeEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2_t
...
                    iev->fle_blob_subtype);
         return false;
     }
 
+    if (iev->fle_blob_subtype == MC_SUBTYPE_FLE2IndexedEqualityEncryptedValueV2) {
+        iev->type = kTypeEquality;
+    } else {
+        iev->type = kTypeRange;
+    }
+
```
sgolemon-corp, line 420: I feel like we can combine the type setting the error checking to make something a little more readable, e.g.:
```suggestion
    if (iev->fle_blob_subtype == MC_SUBTYPE_FLE2IndexedEqualityEncryptedValueV2) {
        iev->type = kTypeEquality;
    } else if (iev->fle_blob_subtype == MC_SUBTYPE_FLE2IndexedRangeEncryptedValueV2) {
        iev->type = kTypeRange;
    } else {
        CLIENT_ERR("mc_FLE2IndexedEncryptedValueV2_parse expected "
                   "fle_blob_subtype MC_SUBTYPE_FLE2Indexed(Equality|Range)EncryptedValueV2 got: %" PRIu8,
                   iev->fle_blob_subtype);
        return false;    
    }

```

sgolemon-corp (approved):

src/mc-fle2-payload-iev-private-v2.h
sgolemon-corp: I thought about it overnight and a good checksum on these tickets might be to bring latest libmongocrypt into your local SERVER checkout (along with this change) and see if that enables implementing SERVER-95887 and SERVER-95889.

src/mc-fle2-tag-and-encrypted-metadata-block-private.h
kevinAlbs: No strong opinion. I expect `mc_FLE2TagAndEncryptedMetadataBlock_parse` may only be useful for the server. If a buffer seems preferable, no objection from me.

src/mc-fle2-payload-iev-private-v2.h
```diff
@@ -122,12 +78,81 @@ bool mc_FLE2IndexedEqualityEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2
...
+                                          mongocrypt_status_t *status);
+
+/*
+ * Serializes an mc_FLE2IndexedEncryptedValueV2_t into a buffer.
+ *
+ * The serialized output follows the same layout as the input `buf` to
+ * mc_FLE2IndexedEncryptedValueV2_parse, allowing for round-trip
+ * conversions between the serialized and parsed forms.
+ *
+ * Retuns an error if the input structure is not valid, or if the buffer
```
kevinAlbs, line 114:
```suggestion
 * Returns an error if the input structure is not valid, or if the buffer
```

src/mc-fle2-payload-iev-v2.c
```diff
@@ -412,26 +422,92 @@ bool mc_FLE2IndexedRangeEncryptedValueV2_parse(mc_FLE2IndexedEncryptedValueV2_t
...
     CHECK_AND_RETURN(mc_reader_read_buffer(&reader, &iev->ServerEncryptedValue, SEV_len, status));
 
-    // Ignore Metadata block.
-    BSON_ASSERT(mc_reader_get_remaining_length(&reader) == edges_len);
+    iev->metadata = (mc_FLE2TagAndEncryptedMetadataBlock_t *)bson_malloc0(
+        iev->edge_count * sizeof(mc_FLE2TagAndEncryptedMetadataBlock_t));
+    if (!iev->metadata) {
+        CLIENT_ERR("Failed to allocate memory for metadata array");
+        return false;
+    }
```
kevinAlbs, line 454:
```suggestion
```

Remove unnecessary check. `bson_malloc0` aborts on [failure to allocate](https://mongoc.org/libbson/1.28.1/bson_malloc0.html).

src/mc-fle2-payload-iev-v2.c
```diff
@@ -322,70 +300,99 @@ void mc_FLE2IndexedEncryptedValueV2_destroy(mc_FLE2IndexedEncryptedValueV2_t *ie
...
-    mc_reader_init_from_buffer(&reader, buf, __FUNCTION__);
+    return iev->edge_count;
+}
 
-    CHECK_AND_RETURN(mc_reader_read_u8(&reader, &iev->fle_blob_subtype, status));
+bool mc_FLE2IndexedEncryptedValueV2_get_edge(const mc_FLE2IndexedEncryptedValueV2_t *iev,
+                                             mc_FLE2TagAndEncryptedMetadataBlock_t *out,
+                                             const uint8_t edge_index,
+                                             mongocrypt_status_t *status) {
+    assert(iev);
```
kevinAlbs, line 337:
```suggestion
    BSON_ASSERT_PARAM (iev);
```

Suggest using `BSON_ASSERT_PARAM` instead of `assert`. `BSON_ASSERT_PARAM` produces a descriptive message on error:

```
The parameter: iev, in function mc_FLE2IndexedEncryptedValueV2_get_edge_count, cannot be NULL
```

Suggestion applies to other instances of `assert`.

src/mc-fle2-tag-and-encrypted-metadata-block.c
```diff
@@ -0,0 +1,79 @@
...
+#include "mc-writer-private.h"
+#include "mongocrypt-private.h"
+
+#define CHECK_AND_RETURN(x)                                                                                            \
+    if (!(x)) {                                                                                                        \
+        return false;                                                                                                  \
+    }
+
+void mc_FLE2TagAndEncryptedMetadataBlock_init(mc_FLE2TagAndEncryptedMetadataBlock_t *metadata) {
+    memset(metadata, 0, sizeof(mc_FLE2TagAndEncryptedMetadataBlock_t));
```
kevinAlbs, line 28:
```suggestion
    BSON_ASSERT_PARAM(metadata);
    memset(metadata, 0, sizeof(mc_FLE2TagAndEncryptedMetadataBlock_t));
```

src/mc-fle2-tag-and-encrypted-metadata-block.c
```diff
@@ -0,0 +1,79 @@
...
+
+    _mongocrypt_buffer_cleanup(&metadata->encryptedCount);
+    _mongocrypt_buffer_cleanup(&metadata->tag);
+    _mongocrypt_buffer_cleanup(&metadata->encryptedZeros);
+}
+
+bool mc_FLE2TagAndEncryptedMetadataBlock_parse(mc_FLE2TagAndEncryptedMetadataBlock_t *metadata,
+                                               const _mongocrypt_buffer_t *buf,
+                                               mongocrypt_status_t *status) {
+    BSON_ASSERT_PARAM(buf);
```
kevinAlbs, line 42:
```suggestion
    BSON_ASSERT_PARAM(metadata);
    BSON_ASSERT_PARAM(buf);
```

kevinAlbs (approved):
LGTM with minor feedback.

PR #900, opened 2024-10-24
MONGOCRYPT-730 Add option to configure DEK cache lifetime

adriandole:


test/test-mongocrypt.h
```diff
@@ -239,4 +241,7 @@ void _test_ctx_wrap_and_feed_key(mongocrypt_ctx_t *ctx,
                                  _mongocrypt_buffer_t *key,
                                  mongocrypt_status_t *status);
 
+
```
kevinAlbs, line 244:
```suggestion
```

To fix format. To run format locally, I use:
```
./etc/format-all.sh
```

And modify the `run_python` function to add my path to `python`.

PR #901, opened 2024-10-30
MONGOCRYPT-725 fix UBSan alignment errors

kevinAlbs:
# Summary

- Fix alignment errors reported by UBSan.
- Add a UBSan task to Evergreen.

# Details

This PR is hopefully intended to address the reported errors in MONGOCRYPT-725 by addressing reported UBSan errors like the following:
```
cmake-build/_deps/embedded_mcd-src/src/libbson/src/bson/bson.c:1778:10: runtime error: member access within misaligned address 0x000131f052b0 for type 'bson_impl_inline_t', which requires 128 byte alignment
0x000131f052b0: note: pointer points here
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^
```

The UBSan errors appear due to dynamic allocation of structs containing libbson types defined with a default 128-byte alignment: `bson_t` and `bson_iter_t`. This only appeared to impact `mongocrypt_ctx_t` and `_mongocrypt_key_doc_t`. Utilities added in https://github.com/mongodb/mongo-c-driver/pull/1072 are used to do aligned allocation.

As an added fix, the allocation of `mongocrypt_ctx_t` is updated to account for all possible context implementations.

The new UBSan task was run without the fix to ensure it fails: [failing patch](https://spruce.mongodb.com/version/672255573428fc0007163971). With the fix, the task succeeds: [passing patch](https://spruce.mongodb.com/version/672258362cd0820007527938).

PR #903, opened 2024-11-01
Address memory leaks in test suite

eramongodb:
Observed when running the test suite locally while reviewing https://github.com/mongodb/libmongocrypt/pull/901. It is unclear why these are not being caught by the sanitizer tasks on Evergreen.


kevinAlbs (approved):
Thank you for the fix.

> It is unclear why these are not being caught by the sanitizer tasks on Evergreen.

I suspect it is due to a [misplaced comment](https://github.com/mongodb/libmongocrypt/blob/fb3fc2d508d104c5e90a2dd68271537563524de5/.evergreen/config.yml#L436) 🤦. The YAML is folded into a single-line (with `>-`). Re-enabling ASan resulted in non-obvious errors to me. Filed MONGOCRYPT-731 to track investigating.

PR #910, opened 2024-11-15
SERVER-95889 Make FLE2IndexedEqualityEncryptedValueV2 wrap mc_FLE2IndexedEncryptedValueV2_t

marksg07:
Main repo PR: https://github.com/10gen/mongo/pull/29265

- Moves structs needed by fle_crypto to headers
- Adds validation functions for mc_FLE2IndexedEncryptedValueV2_t and mc_FLE2TagAndEncryptedMetadataBlock_t which are needed by fle_crypto

### DIFF
--- a/src/mc-fle2-payload-iev-private-v2.h
+++ b/src/mc-fle2-payload-iev-private-v2.h
@@ -81,9 +81,9 @@
  */
 
 typedef enum {
-    kTypeInit,
-    kTypeEquality,
-    kTypeRange,
+    kTypeInitV2,
+    kTypeEqualityV2,
+    kTypeRangeV2,
 } _mc_fle2_iev_v2_type;
 
 typedef struct _mc_FLE2IndexedEncryptedValueV2_t {
--- a/src/mc-fle2-payload-iev-private-v2.h
+++ b/src/mc-fle2-payload-iev-private-v2.h
@@ -80,11 +80,43 @@
  *
  */
 
-typedef struct _mc_FLE2IndexedEncryptedValueV2_t mc_FLE2IndexedEncryptedValueV2_t;
+typedef enum {
+    kTypeInit,
+    kTypeEquality,
+    kTypeRange,
+} _mc_fle2_iev_v2_type;
 
-mc_FLE2IndexedEncryptedValueV2_t *mc_FLE2IndexedEncryptedValueV2_new(void);
-bson_type_t mc_FLE2IndexedEncryptedValueV2_get_bson_value_type(const mc_FLE2IndexedEncryptedValueV2_t *iev,
-                                                               mongocrypt_status_t *status);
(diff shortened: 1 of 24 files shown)


src/mc-fle2-payload-iev-private-v2.h
```diff
@@ -80,7 +80,39 @@
  *
  */
 
-typedef struct _mc_FLE2IndexedEncryptedValueV2_t mc_FLE2IndexedEncryptedValueV2_t;
+typedef enum {
+    kTypeInitV2,
```
sgolemon-corp, line 84: I think you're going to find that even these names are too generic given how C ODRs its enums into the global symbol space.  It wasn't a problem when these were scoped to a single TU, but now that they're in a header, we should give them more distinctness.

e.g.
```c
typedef enum {
  kFLE2IEVTypeInitV2,
  kFLE2IEVTypeEqualityV2,
  kFLE2IEVTypeRangeV2,
};
```

src/mongocrypt-util.c
```diff
@@ -124,6 +124,34 @@ const char *mc_bson_type_to_string(bson_type_t t) {
     }
 }
 
+bool mc_is_valid_bson_type(int bson_type) {
```
sgolemon-corp, line 127: This function seems.... sus.

You seem to just be checking if the value is in the explicitly list of named numeric values.

Thing is that's not very meaningful in terms of if a type is "valid" or not.

`MINKEY`-`MAXKEY` are all "valid" as BSON types, for example.  Those two constants define the endpoints of the "User Defined BSONElements" range.

On the other side, EOD and UNDEFINED don't make sense as valid encodable types. (Though "Undefined" is defined, it's just no longer a valid type, so that one gets.... complicated).

TL;DR - The only truly "invalid" type is EOD.

Unless I'm misunderstanding the purpose of this method, I'd probably just reduce it to:
```c
return bson_type != BSON_TYPE_EOD;
```

sgolemon-corp (changes requested):

src/mongocrypt-util.c
joshbsiegel: I went ahead and modified this in order to get a test passing, feel free to change it back if necessary.

marksg07: Huh, I didn't realize that `[MAXKEY, MINKEY]` are valid, I was wondering what those meant. Thanks for the clarification.

src/mc-fle2-payload-iev-private-v2.h
marksg07: Changed

kevinAlbs (approved):
LGTM with minor comments.

sgolemon-corp (approved):

markbenvenuto (approved):

PR #912, opened 2024-11-18
DEVPROD-6063: update to macos 11

drichmdb:
This commit bumps our builds from MacOS 10.14 to MacOS 11. 10.14 has been
unsupported for several years. MacOS 11 is also EOL, but it seems to pass
all of our tests and isn't quite as old as 10.14. I tried to bump all the
way to MacOS 14, but ran into some kind of linking error. I think we can
return to that upgrade at a different time.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1017,8 +1017,8 @@ buildvariants:
     distros:
     - ubuntu2004-small
 - name: macos_x86_64
-  display_name: "macOS (x86_64) 10.14"
-  run_on: macos-1014
+  display_name: "macOS (x86_64) 11"
+  run_on: macos-1100
   tasks:
   - build-and-test-and-upload
   - build-and-test-shared-bson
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -1017,8 +1017,8 @@ buildvariants:
     distros:
     - ubuntu2004-small
 - name: macos_x86_64
-  display_name: "macOS (x86_64) 10.14"
-  run_on: macos-1014
+  display_name: "macOS (x86_64) 14"
+  run_on: macos-14
```
kevinAlbs, line 1021:
```suggestion
  display_name: "macOS (x86_64) 11"
  run_on: macos-11
```

kevinAlbs (commented):
Thank you for the PR. Filed MONGOCRYPT-748 to track changes in libmongocrypt.

.evergreen/config.yml
drichmdb: Whoops, forgot to add my local changes after patch testing. Fixed now!

kevinAlbs (approved):

PR #915, opened 2024-11-18
MONGOCRYPT-744 modify _mongocrypt_hmac_sha_256 to accept crypto as nullptr

joshbsiegel:
# Summary
This PR allows `_mongocrypt_hmac_sha_256` to accept `NULL` for the `crypto` parameter.

# Background
Currently, the comment above `_mongocrypt_hmac_sha_256` describes that the function will call the native implementation if there is no hook set on the `crypto` parameter.

In cases where the program desires the native implementation to be called, it is helpful to simply allow `NULL` to be passed through instead of creating a dummy crypto object.

### DIFF
--- a/src/mongocrypt-crypto.c
+++ b/src/mongocrypt-crypto.c
@@ -299,7 +299,6 @@ bool _mongocrypt_hmac_sha_256(_mongocrypt_crypto_t *crypto,
                               const _mongocrypt_buffer_t *in,
                               _mongocrypt_buffer_t *out,
                               mongocrypt_status_t *status) {
-    BSON_ASSERT_PARAM(crypto);
     BSON_ASSERT_PARAM(key);
     BSON_ASSERT_PARAM(in);
     BSON_ASSERT_PARAM(out);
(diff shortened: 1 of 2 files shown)


src/mongocrypt-crypto.c
```diff
@@ -299,7 +299,6 @@ bool _mongocrypt_hmac_sha_256(_mongocrypt_crypto_t *crypto,
                               const _mongocrypt_buffer_t *in,
                               _mongocrypt_buffer_t *out,
                               mongocrypt_status_t *status) {
-    BSON_ASSERT_PARAM(crypto);
```
addaleax, line 302: - When would calling `_mongocrypt_hmac_sha_256` with `crypto == nullptr` not be a bug? I imagine there's some use case for this in the server code that's the reason for this ticket, but in libmongocrypt itself, I think that would always be a bug.
- Why doesn't this apply to the other functions in this file that use the hook if provided and fall back to the native implementation otherwise?

joshbsiegel: There's a rather simple workaround that we can do in the server code for this specific case. I think it would be better to do that than to change the behaviors/expectations of multiple functions in `libmongocrypt`. Thanks for bringing up your concerns.

PR #918, opened 2024-11-22
Fix format script

kevinAlbs:
# Summary
- Update formatting script for Python 3.12
- Add CONTRIBUTING.md
- Relocate design rationale from README.md to integrating.md

# Details
The formatting script is updated to address an observed error on Python 3.12:

```
$ bash etc/format-all.sh
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pkg_resources'
```

The error comes from this check:
```
run_python -c 'import pkg_resources; pkg_resources.require("pipx>=0.17.0,<2.0")'
```

`pkg_resources` is removed in Python 3.12. I expect that check is not strictly needed. This check has been removed.

A `MONGOCRYPT_PYTHON` environment variable is added to specify an override python version. This is documented in a new CONTRIBUTING.md document.

### DIFF
--- a/.evergreen/init.sh
+++ b/.evergreen/init.sh
@@ -222,6 +222,7 @@ run_ctest() {
 
 run_python() {
     pys=(
+        "${MONGOCRYPT_PYTHON:-}"
         py
         python3.14
         python3.13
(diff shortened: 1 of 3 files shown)


vector-of-bool (approved):
LGTM

PR #923, opened 2024-12-12
MONGOCRYPT-760 Rename `set_error` to `kms_set_error`

MonkeybreadSoftware:
set_error is an awful generic name for a global function.

Sadly this easily conflicts with other libraries if you statically link this into an application.

please have us rename set_error to kms_set_error to avoid link errors with libraries/applications that also have set_error as function.


kevinAlbs (approved):
Thank you for the improvement. LGTM

PR #930, opened 2024-12-31
PYTHON-5017 Use a separate PyPI publish step

blink1073:


.github/workflows/release-python.yml
```diff
@@ -68,6 +68,22 @@ jobs:
       ref: ${{ needs.pre-publish.outputs.version }}
 
   publish:
```
ShaneHarvey, line 70: Do we need to add?:
```
needs: [build-dist, ...
```

ShaneHarvey (changes requested):

.github/workflows/release-python.yml
blink1073: Fixed

ShaneHarvey (approved):

blink1073:
I'll make an rc release to to verify things are working and cut the final on Thursday.

PR #938, opened 2025-01-09
mount proc and sys in Debian package build chroot

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/67803945ed23370007a05deb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


kevinAlbs (approved):
Thank you for the fix!

PR #939, opened 2025-01-10
MONGOCRYPT-756 Convert FLE2EncryptionPlaceholder to FLE2InsertUpdatePayloadV2 for text search indexed fields

erwee:


erwee:
tests to follow..

src/mc-fle2-insert-update-payload-v2.c
```diff
@@ -348,8 +401,117 @@ bool mc_FLE2InsertUpdatePayloadV2_serializeForRange(const mc_FLE2InsertUpdatePay
...
+                                                                                                                       \
+        for (size_t i = 0; i < array->len; i++) {                                                                      \
+            mc_Text##Type##TokenSet_t ts = _mc_array_index(array, mc_Text##Type##TokenSet_t, i);                       \
+            bson_uint32_to_string(index, &index_string, storage, sizeof(storage));                                     \
+                                                                                                                       \
+            if (!_fle2_serialize_Text##Type##TokenSet(&arr_bson, index_string, &ts)) {                                 \
+                return false;                                                                                          \
+            }                                                                                                          \
+            if (index == UINT32_MAX) {                                                                                 \
```
marksg07, line 455: Can we just iterate up to `BSON_MIN(array->len, UINT32_MAX)` instead of doing this check every iteration?

src/mongocrypt-marking.c
```diff
@@ -271,6 +271,118 @@ DERIVE_TOKEN_IMPL(ESC)
...
 #undef DERIVE_TOKEN_IMPL
 
+/**
+ * Calculates:
+ * E?CToken = HMAC(collectionLevel1Token, n)
+ * E?CText<T>Token = HMAC(E?CToken, t)
+ * E?CText<T>DerivedFromDataTokenAndContentionFactorToken = HMAC(HMAC(E?CText<T>Token, v) cf)
+ *
+ * E?C = EDC|ESC|ECC
+ * n = 1 for EDC, 2 for ESC, 3 for ECC
```
marksg07, line 281: Remove reference to ECC.

src/mongocrypt-marking.c
```diff
@@ -271,6 +271,118 @@ DERIVE_TOKEN_IMPL(ESC)
...
+ * <T> = Exact|Substring|Suffix|Prefix
+ * t = 1 for Exact, 2 for Substring, 3 for Suffix, 4 for Prefix
+ * cf = contentionFactor
+ *
+ * E?CText<T>DerivedFromDataTokenAndContentionFactorToken is saved to out.
+ * Note that {out} is initialized even on failure.
```
marksg07, line 287: Why don't we cleanup on failure?

src/mongocrypt-marking.c
```diff
@@ -271,6 +271,118 @@ DERIVE_TOKEN_IMPL(ESC)
...
+                                                                                    textToken,                         \
+                                                                                    value,                             \
+                                                                                    (uint64_t)contentionFactor,        \
+                                                                                    status);                           \
+        mc_##Name##Text##Type##Token_destroy(textToken);                                                               \
+        if (!fromDataAndContentionFactor) {                                                                            \
+            return false;                                                                                              \
+        }                                                                                                              \
+        BSON_ASSERT(contentionFactor >= 0);                                                                            \
```
marksg07, line 322: Let's move this to the start.

test/test-mc-fle2-encryption-placeholder.c
```diff
@@ -0,0 +1,448 @@
...
+        "t" : {"$numberInt" : "1"},
+        "a" : {"$numberInt" : "4"},
+        "ki" : {"$binary" : {"base64" : "EjRWeBI0mHYSNBI0VniQEg==", "subType" : "04"}},
+        "ku" : {"$binary" : {"base64" : "q83vqxI0mHYSNBI0VniQEg==", "subType" : "04"}},
+        "v" : % s,
+        "cm" : {"$numberLong" : "7"}
+    });
+    bson_t *const as_bson = TMP_BSON(template, spec_json_in);
+
+    mongocrypt_status_t *inner_status = mongocrypt_status_new();
```
marksg07, line 334: Is there a reason we have an inner status here, rather than using the passed status for everything? If we get a non-OK status, we fail anyways, so it shouldn't matter, right?

marksg07 (changes requested):

src/mc-fle2-encryption-placeholder-private.h
```diff
@@ -119,6 +119,72 @@ bool mc_FLE2RangeInsertSpec_parse(mc_FLE2RangeInsertSpec_t *out,
                                   bool use_range_v2,
                                   mongocrypt_status_t *status);
 
+/* mc_FLE2SubstringInsertSpec_t holds the parameters used to encode for substring search. */
+typedef struct {
```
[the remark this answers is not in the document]
erwee: yours can go in first, since it's almost approved for merge :)

src/mc-fle2-encryption-placeholder-private.h
```diff
@@ -119,6 +119,72 @@ bool mc_FLE2RangeInsertSpec_parse(mc_FLE2RangeInsertSpec_t *out,
...
+    // lower bound code point length of valid prefix queries
+    int32_t lb;
+} mc_FLE2PrefixInsertSpec_t;
+
+/** mc_FLE2TextSearchInsertSpec_t represents the text search insert specification that is
+ * encoded inside of a FLE2EncryptionPlaceholder. See
+ * https://github.com/mongodb/mongo/blob/master/src/mongo/crypto/fle_field_schema.idl
+ * for the representation in the MongoDB server. */
+typedef struct {
+    // v is the value to encrypt.
```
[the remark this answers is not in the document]
erwee: It's simpler to keep it a `bson_iter_t` of type string, because when we encrypt it, we encrypt the bson value (which includes the 4 byte length + null terminator), not the raw string.

test/test-mc-fle2-encryption-placeholder.c
erwee: yep. removed

src/mongocrypt-marking.c
erwee: done

src/mongocrypt-marking.c
erwee: "initialized" here just means we memset the buffer struct, not allocate memory, so there's nothing to clean up on failure.

src/mongocrypt-marking.c
erwee: done

src/mc-fle2-insert-update-payload-v2.c
erwee: done

src/mc-fle2-insert-update-payload-v2.c
```diff
@@ -21,11 +21,63 @@
...
+
+DEF_TEXT_SEARCH_TOKEN_SET_INIT_CLEANUP(Exact)
+DEF_TEXT_SEARCH_TOKEN_SET_INIT_CLEANUP(Substring)
+DEF_TEXT_SEARCH_TOKEN_SET_INIT_CLEANUP(Suffix)
+DEF_TEXT_SEARCH_TOKEN_SET_INIT_CLEANUP(Prefix)
+#undef DEF_TEXT_SEARCH_TOKEN_SET_INIT_CLEANUP
+
+void mc_TextSearchTokenSets_init(mc_TextSearchTokenSets_t *tsts) {
+    BSON_ASSERT_PARAM(tsts);
+    memset(tsts, 0, sizeof(mc_TextSearchTokenSets_t));
```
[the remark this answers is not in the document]
erwee: done

src/mc-fle2-encryption-placeholder-private.h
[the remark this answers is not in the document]
marksg07: Wait, when you say "we encrypt the bson element" - do you mean that the string passed into StrEncode is not the raw string, but the BSON element including the length + null terminator?

src/mc-fle2-insert-update-payload-v2.c
```diff
@@ -444,18 +444,22 @@ SERIALIZE_TEXT_TOKEN_SET_FOR_TYPE_IMPL(Prefix)
...
             if (!_fle2_serialize_Text##Type##TokenSet(&arr_bson, index_string, &ts)) {                                 \
                 return false;                                                                                          \
             }                                                                                                          \
-            if (index == UINT32_MAX) {                                                                                 \
-                break;                                                                                                 \
-            }                                                                                                          \
             index++;                                                                                                   \
+        }                                                                                                              \
+        if (index == UINT32_MAX && array->len > UINT32_MAX) {                                                          \
```
marksg07, line 457: Not sure if I understand -- Why are we appending one extra element in this case?

src/mc-fle2-insert-update-payload-v2.c
```diff
@@ -444,18 +444,22 @@ SERIALIZE_TEXT_TOKEN_SET_FOR_TYPE_IMPL(Prefix)
         const char *index_string = NULL;                                                                               \
         char storage[16];                                                                                              \
         uint32_t index = 0;                                                                                            \
-                                                                                                                       \
-        for (size_t i = 0; i < array->len; i++) {                                                                      \
-            mc_Text##Type##TokenSet_t ts = _mc_array_index(array, mc_Text##Type##TokenSet_t, i);                       \
+        uint32_t limit = (array->len > UINT32_MAX) ? UINT32_MAX : (uint32_t)array->len;                                \
+        while (index < limit) {                                                                                        \
```
marksg07, line 448: Nit: 

```suggestion
        for (; index < limit; index++) {                                                                                        \
        
```
And remove the increment at the end of the loop.

src/mongocrypt-marking.c
marksg07: Makes sense, thanks.

src/mc-fle2-encryption-placeholder-private.h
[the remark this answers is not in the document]
erwee: No, I mean when we do the AES-CBC encrypt & derive the ESC/EDC/Server value-derived tokens, we pass in the (length + string + nullbyte). For StrEncode, we pass in just the string.

src/mc-fle2-insert-update-payload-v2.c
erwee: done

src/mc-fle2-insert-update-payload-v2.c
erwee: bc the loop above it is not inclusive of index at UINT32_MAX if the array length exceeds that.  highly unlikely, but this is the same logic for range edge tokens in `mc_FLE2InsertUpdatePayloadV2_serializeForRange()`.

kevinAlbs (commented):
LGTM with addition of public docs to note this is experimental.

kevinAlbs (approved):

src/mc-fle2-encryption-placeholder-private.h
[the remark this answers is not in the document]
marksg07: Wait, if this is true, wouldn't it throw off the CBC length calculation in `StrEncode`? I'm calculating the CBC length based on the base string, but if we have 5 extra bytes of info, the cutoff for rounding up will be different than it is otherwise, no?

src/mc-fle2-insert-update-payload-v2.c
marksg07: My question is - if we are ignoring all the elements beyond index `UINT32_MAX`, why not also ignore the element at `UINT32_MAX`? Why do we have an artificial limit of `UINT32_MAX + 1` elements?

src/mc-fle2-encryption-placeholder-private.h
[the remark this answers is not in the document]
erwee: No, because StrEncode and the CBC length calculation operates on the code point length of the raw string. The extra 5 bytes of info does not factor in the StrEncode calculation, but only during token derivation.

marksg07 (commented):
Couple of follow-up questions.

src/mc-fle2-insert-update-payload-v2.c
erwee: I agree its arbitrary and we don't need the extra code. Removed.

marksg07 (approved):

PR #941, opened 2025-01-25
MONGOCRYPT-759 Implement CFold

marksg07:
The new `unicode/` implements folding as implemented in `mongo::unicode::String::caseFoldAndStripDiacritics` in the server. The two `_map.c` files are generated by `gen_[diacritic|casefold]_map.py` in the server, with modifications so that they work in C. Note that these maps are not the same as the ones on the server: Those use unicode 8.0, while we use unicode 13.0.0 (for the simple reason that this is the latest unicode version supported by the version of python we use for the server).

We now do random unicode string generation for unit testing, rather than the static strings we had before. My hope is that this will be able to test a wider variety of cases than I myself can think of.


src/unicode/fold.c
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        if (!append_utf8_codepoint(codepoint, &output_it, status)) {
+            free(*out_str);
+            return false;
+        }
+    }
+
+    *out_len = (size_t)(output_it - *out_str);
+    *out_str = realloc(*out_str, *out_len);
```
erwee, line 133: I personally wouldn't bother with the `realloc` here just to shrink it. The folded string is not gonna be around long enough to make it worth the realloc cost.

Need to null terminate output string:
```suggestion
    *output_it = '\0';
    *out_len = (size_t)(output_it - *out_str);
```

test/test-unicode-fold.c
```diff
@@ -0,0 +1,92 @@
...
+    TEST_UNICODE_FOLD_ALL_CASES("СКОЛЬКО ТЕБЕ ЛЕТ?", "сколько тебе лет?", "СКОЛЬКО ТЕБЕ ЛЕТ?", "сколько тебе лет?");
+    TEST_UNICODE_FOLD_ALL_CASES("Πόσο χρονών είσαι?", "πόσο χρονών είσαι?", "Ποσο χρονων εισαι?", "ποσο χρονων εισαι?");
+    // Tests of decomposed unicode
+    const char nfd1[] = {'C', 'a', 'f', 'e', 0xcc, 0x81, 0};
+    const char nfd1_lower[] = {'c', 'a', 'f', 'e', 0xcc, 0x81, 0};
+    TEST_UNICODE_FOLD_ALL_CASES(nfd1, nfd1_lower, "Cafe", "cafe");
+    const char nfd2[] = {'C', 'a', 'f', 'E', 0xcc, 0x81, 0};
+    const char nfd2_lower[] = {'c', 'a', 'f', 'e', 0xcc, 0x81, 0};
+    TEST_UNICODE_FOLD_ALL_CASES(nfd2, nfd2_lower, "CafE", "cafe");
+
```
erwee, line 86: Add:
```suggestion
    TEST_UNICODE_FOLD("fo\0bar", 6, "fo\0bar", 6, kUnicodeFoldToLower | kUnicodeFoldRemoveDiacritics);

```

test/test-unicode-fold.c
```diff
@@ -0,0 +1,92 @@
...
+                    (int)input_len,                                                                                    \
+                    input,                                                                                             \
+                    (int)expected_len,                                                                                 \
+                    expected,                                                                                          \
+                    (int)_len,                                                                                         \
+                    _buf);                                                                                             \
+        ASSERT_CMPSIZE_T(_len, ==, expected_len);                                                                      \
+        ASSERT_CMPBYTES((uint8_t *)_buf, _len, (uint8_t *)expected, expected_len);                                     \
+        free(_buf);                                                                                                    \
```
erwee, line 36:
```suggestion
        bson_free(_buf);                                                                                                    \
```

test/test-unicode-fold.c
```diff
@@ -0,0 +1,92 @@
...
+        // Diacritics
+        if (ch >= 'A' && ch <= 'Z') {
+            buf2[0] = ch + 0x20;
+            TEST_UNICODE_FOLD_ALL_CASES(buf1, buf2, buf1, buf2);
+        } else if (ch == '^' || ch == '`') {
+            TEST_UNICODE_FOLD_ALL_CASES(buf1, buf1, "", "");
+        }
+    }
+    free(buf1);
+    free(buf2);
```
erwee, line 70:
```suggestion
    bson_free(buf1);
    bson_free(buf2);
```

src/unicode/fold.c
```diff
@@ -0,0 +1,135 @@
...
+    BSON_ASSERT_PARAM(str);
+    BSON_ASSERT_PARAM(out_str);
+    BSON_ASSERT_PARAM(out_len);
+    BSON_ASSERT_PARAM(status);
+
+    if (!(options & (kUnicodeFoldRemoveDiacritics | kUnicodeFoldToLower))) {
+        CLIENT_ERR("unicode_fold: Either case or diacritic folding must be enabled");
+        return false;
+    }
+    *out_str = bson_malloc(len);
```
kevinAlbs, line 75: Following [comment](https://github.com/10gen/mongo/blob/f35c66a4417ab546443730bba3549cf53a314ba1/src/mongo/db/fts/unicode/string.cpp#L171-L174), allocate space to account for cases where folding increases UTF-8 encoded length:

```suggestion
    // Allocate space for possible growth. Folding characters may result in longer UTF-8 sequences.
    // 2x is an upper bound. With current fold maps, the largest growth is a 2-byte sequence mapping to a 3-byte sequence.
    *out_str = bson_malloc(2 * len);
```

The true maximum growth appears to be 3/2: https://gist.github.com/kevinAlbs/bd4d916f62fa3a49f07bc1c6a679b67a

Add a test to confirm:

```c
TEST_UNICODE_FOLD("\xe2\xb1\xa6", 4, "\xc8\xbe", 3, kUnicodeFoldToLower);
```

erwee (changes requested):

src/unicode/fold.c
marksg07: Thanks so much for this! To give some background -- I originally constructed the two folding maps such that any cases where byte length would increase were excluded, so I changed this so that we only allocate the exact size of the buffer. But later, I realized that the algorithm in the paper as it was written can actually handle the byte-length-increasing case as long as the codepoint length doesn't increase, so I reconstructed the maps without fixing this. Reading your comment, I realized I actually may have to change this back, since in the new iteration of the paper I believe increasing byte length is also bad. Confirming with the crypto team on this.

marksg07: Changed and added a test.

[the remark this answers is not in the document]
marksg07: Now null-terminating.

test/test-unicode-fold.c
marksg07: Added

test/test-unicode-fold.c
```diff
@@ -0,0 +1,93 @@
...
+    char *buf2 = bson_malloc0(2);
+    for (unsigned char ch = 0; ch <= 0x7f; ch++) {
+        buf1[0] = ch;
+        // Diacritics
+        if (ch >= 'A' && ch <= 'Z') {
+            buf2[0] = ch + 0x20;
+            TEST_UNICODE_FOLD_ALL_CASES(buf1, buf2, buf1, buf2);
+        } else if (ch == '^' || ch == '`') {
+            TEST_UNICODE_FOLD_ALL_CASES(buf1, buf1, "", "");
+        }
```
erwee, line 68: Add an `else` block to test all other single-byte cases. We just need to make sure they don't expand to >2 bytes, since that's assumed by the memory allocation in `unicode_fold`.

erwee (changes requested):
one more nit, and lgtm!

erwee (approved):

test/test-mc-text-search-str-encode.c
```diff
@@ -293,246 +321,774 @@ static void test_nofold_substring_case_multiple_mlen(_mongocrypt_tester_t *teste
...
+}
+
+const uint32_t UNFOLDED_CASES[] = {0, 1, 3, 16};
+// Predefined lengths to test a variety of cases
+const uint32_t SHORT_LEN = 9;
+const uint32_t MEDIUM_LEN = 16;
+const uint32_t LONG_LEN = 27;
+
+static void _test_text_search_str_encode_suffix_prefix(_mongocrypt_tester_t *tester) {
+    srand((unsigned int)time(0));
```
kevinAlbs, line 512:
```suggestion
    unsigned int seed = (unsigned int)time(0);
    TEST_PRINTF("Testing with seed: %u", seed);
    srand(seed);
```

Suggest printing seed. That may make it easer to reproduce a test failure if the log output cannot render all UTF-8 characters in a failed test case.

test/test-mc-text-search-str-encode.c
```diff
@@ -293,246 +321,774 @@ static void test_nofold_substring_case_multiple_mlen(_mongocrypt_tester_t *teste
...
+                                             long_unfolded_codepoint_len,
+                                             casef,
+                                             diacf,
+                                             foldable_codepoints);
 }
 
-static void _test_text_search_str_encode_substring_utf8(_mongocrypt_tester_t *tester) {
-    test_text_search_str_encode_substring(tester, short_unicode_string, medium_unicode_string, long_unicode_string);
+static void _test_text_search_str_encode_substring(_mongocrypt_tester_t *tester) {
+    srand((unsigned int)time(0));
```
kevinAlbs, line 1051:
```suggestion
    unsigned int seed = (unsigned int)time(0);
    TEST_PRINTF("Testing with seed: %u", seed);
    srand(seed);
```

PR #945, opened 2025-01-27
MONGOCRYPT-766 Fix compilation in C23

jeroen:
```
kms/kms_request.c:792:14: warning: initialization of pointer of type
'char *' to null from a constant boolean expression
[-Wbool-conversion]
      return false;
             ^~~~~
kms/kms_request.c:860:14: warning: initialization of pointer of type
'char *' to null from a constant boolean expression
[-Wbool-conversion]
      return false;
             ^~~~~
kms/kms_request.c:864:14: warning: initialization of pointer of type
'char *' to null from a constant boolean expression
[-Wbool-conversion]
      return false;
             ^~~~~
```

### DIFF
--- a/kms-message/src/kms_request.c
+++ b/kms-message/src/kms_request.c
@@ -789,7 +789,7 @@ kms_request_get_signed (kms_request_t *request)
    }
 
    if (!check_and_prohibit_kmip (request)) {
-      return false;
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you! LGTM

PR #946, opened 2025-01-29
MONGOCRYPT-762 Generate text search token sets from StrEncode output

erwee:


src/mongocrypt-marking.c
```diff
@@ -1164,72 +1162,130 @@ static bool _fle2_generate_TextSearchTokenSets(_mongocrypt_key_broker_t *kb,
...
-                                                  common.collectionsLevel1Token,
-                                                  common.serverTokenDerivationLevel1Token,
-                                                  status)) {
-            mc_TextSubstringTokenSet_cleanup(&substrSet);
-            goto fail;
+        mc_substring_set_iter_t set_itr;
+        mc_substring_set_iter_init(&set_itr, encodeSets->substring_set);
+
+        while (mc_substring_set_iter_next(&set_itr, &substring, &bytelen, &appendCount)) {
+            for (; appendCount > 0; appendCount--) {
```
marksg07, line 1195: Let's move everything from 1196-1214 outside the loop, and then append tset multiple times. I don't think it's a problem that the copies all point at the same underlying buffer, since from this point the token sets should be read-only.

src/mongocrypt-marking.c
```diff
@@ -1164,72 +1162,130 @@ static bool _fle2_generate_TextSearchTokenSets(_mongocrypt_key_broker_t *kb,
...
+            }
         }
-        _mc_array_append_val(&tsts->prefixArray, prefixSet);
     }
     payload->textSearchTokenSets.set = true;
     res = true;
 fail:
     _FLE2EncryptedPayloadCommon_cleanup(&common);
-    _mongocrypt_buffer_cleanup(&foldedValue);
+    mc_str_encode_sets_destroy(encodeSets);
```
marksg07, line 1288: Don't we have to also clean up tsts, since it's holding a bunch of initialized buffers?

test/test-mongocrypt-marking.c
```diff
@@ -1344,21 +1345,105 @@ static void validate_text_search_ciphertext(_mongocrypt_tester_t *tester,
         size_t tscount = 0;
         ASSERT(bson_iter_init_find(&b_iter, &b_bson, "s"));
         tscount = validate_text_search_token_set_array_common(&b_iter, crypt);
-        ASSERT((tscount > 0) == !!(flags & TEXTSEARCH_SPEC_HAS_SUBSTRING));
+        ASSERT(expected_tag_counts.substrings == tscount);
```
marksg07, line 1348: nit - 
```suggestion
        ASSERT_CMPSIZE_t(expected_tag_counts.substrings, ==, tscount);
```

for all of these numerical comparison assertions, so that we get more debugging info on failure.

marksg07 (changes requested):

src/mongocrypt-buffer.c
```diff
@@ -345,6 +349,25 @@ void _mongocrypt_buffer_from_iter(_mongocrypt_buffer_t *plaintext, bson_iter_t *
     bson_destroy(&wrapper);
 }
 
+bool append_iter(bson_t *bson, const void *iter, int len) {
```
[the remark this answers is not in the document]
erwee: done

src/mongocrypt-marking.c
erwee: Done, but the array append just does a memcpy of the `mc_Text<?>TokenSet_t`, so we have to make a deep copy of it and the mongocrypt_buffers within it otherwise we end up with multiple owning mongocrypt_buffers pointing to the same heap address.

src/mongocrypt-marking.c
erwee: they get cleaned up when the caller of this calls `mc_FLE2InsertUpdatePayloadV2_cleanup`.

src/mongocrypt-marking.c
```diff
@@ -1192,29 +1192,36 @@ static bool _fle2_generate_TextSearchTokenSets(_mongocrypt_key_broker_t *kb,
...
+                                                        status)) {
                 _mongocrypt_buffer_cleanup(&asBsonValue);
+                mc_TextSubstringTokenSet_cleanup(&tset);
+                goto fail;
+            }
+            _mongocrypt_buffer_cleanup(&asBsonValue);
+
+            for (; appendCount > 1; appendCount--) {
+                mc_TextSubstringTokenSet_t tset_copy;
+                mc_TextSubstringTokenSet_init(&tset_copy);
```
marksg07, line 1220: We can skip `init` here, since `copy` sets all of our members.

marksg07 (changes requested):

test/test-mongocrypt-marking.c
```diff
@@ -1443,6 +1535,8 @@ static void test_mc_marking_to_ciphertext_fle2_text_search(_mongocrypt_tester_t
         _mongocrypt_ciphertext_t ciphertext;
         _mongocrypt_ciphertext_init(&ciphertext);
         mongocrypt_t *crypt = _mongocrypt_tester_mongocrypt(TESTER_MONGOCRYPT_DEFAULT);
+        text_search_expected_token_counts counts = {0};
+        counts.prefixes = calculate_expected_nfix_tag_count(1, 100, 10);
```
kevinAlbs, line 1539: Though `v` is an empty string, is beta 1 to account for the `0xFF` added later?

Consider adding 1 to other `calculate_expected_nfix_tag_count` to account for the `0xFF` (though I expect is not strictly needed due to CBC padding).

src/mongocrypt-marking.c
```diff
@@ -1164,72 +1162,147 @@ static bool _fle2_generate_TextSearchTokenSets(_mongocrypt_key_broker_t *kb,
...
+                                                          (const char *)encodeSets->exact.data,
+                                                          (int)encodeSets->exact.len);
+        if (!_fle2_generate_TextExactTokenSet(kb,
+                                              &tsts->exact,
+                                              &asBsonValue,
+                                              contentionFactor,
+                                              common.collectionsLevel1Token,
+                                              common.serverTokenDerivationLevel1Token,
+                                              status)) {
+            goto fail;
```
kevinAlbs, line 1180:
```suggestion
            _mongocrypt_buffer_cleanup(&asBsonValue);
            goto fail;
```

kevinAlbs (approved):
LGTM with minor comments.

test/test-mongocrypt-marking.c
marksg07: We explicitly define beta as being 1 for empty strings, as having beta = 0 would mean we generate no tokens in StrEncode, which would leak that the string is the empty string.

src/mongocrypt-marking.c
```diff
@@ -1217,16 +1216,15 @@ static bool _fle2_generate_TextSearchTokenSets(_mongocrypt_key_broker_t *kb,
 
             for (; appendCount > 1; appendCount--) {
                 mc_TextSubstringTokenSet_t tset_copy;
```
marksg07, line 1218: nit - we only need to do the shallow copy once. all the references are the same in memory.

marksg07 (changes requested):
one more small optimization.

src/mongocrypt-marking.c
erwee: good catch! thanks!

src/mongocrypt-marking.c
erwee: done

marksg07 (approved):
LGTM!

PR #948, opened 2025-02-03
MONGOCRYPT-771 Remove check for `$jsonSchema` siblings

kevinAlbs:
# Summary

Do not error if remote `$jsonSchema` has sibling fields.

# Background & Motivation

When processing a response to `listCollections`, libmongocrypt only checks the response for a top-level `$jsonSchema` within `validator`.

libmongocrypt errors if the `$jsonSchema` has sibling fields and query analysis (mongocryptd/crypt_shared) indicates the JSON Schema has encrypted fields:

```python
db.command({
    "create": "coll",
    "validator": {
        "$jsonSchema": csfle_schema, # Defines 'secret' as encrypted.
        "secret": "foo" # Sibling validator contradicts the JSON schema.
    }
})

encrypted_db["coll"].find_one({}) # Error: schema requires encryption, but collection JSON schema validator has siblings
```

The [spec says this of the motivation](https://github.com/mongodb/specifications/blob/d9b434db87abec375df46864d4fedb2297bc80ec/source/client-side-encryption/client-side-encryption.md#why-limit-to-one-top-level-jsonschema):

> If we allow siblings, we can run into cases where the user specifies a top-level `$and`/`$or` or any arbitrary match-expression that could have nested `$jsonSchema`'s.

However, this check seems incomplete. A validator can have a single top-level JSON schema keyword, and reference encrypted fields:

```python
db.command({
    "create": "coll",
    "validator": {
        "$or": [{"$jsonSchema": csfle_schema}, {"foo": "bar"}]
    }
})

encrypted_db["coll"].insert_one({"secret": "blah", "foo": "bar"}) # libmongocrypt ignores $jsonSchema.
self.assertEqual (db["coll"].find_one()["secret"], "blah") # Unencrypted!
```

And this check can error in valid cases:
```python
db.command({
    "create": "coll",
    "validator": {
        "$jsonSchema": csfle_schema, # Defines 'secret' as encrypted.
        "unrelated": "foo" # Sibling validator does not contradict the JSON schema.
    }
})

with self.assertRaises(EncryptionError) as ctx:
    encrypted_db["coll"].find_one({})
self.assertIn("schema requires encryption, but collection JSON schema validator has siblings", str(ctx.exception))
```

I do not expect this check is useful. If a `$jsonSchema` has sibling fields that contradict the `$jsonSchema`, I expect that is a misconfiguration that would be discovered soon after attempting to insert data. If the siblings do not contradict the JSON Schema, then there is no reason to error.

Related to: https://github.com/mongodb/libmongocrypt/pull/947, query analysis 8.1 supports a new field `csfleEncryptionSchemas` to accept multiple JSON Schemas. The replied `schemaRequiresEncryption` only indicates if __any__ of the sent schemas require encryption. Multiple schemas further complicates this check (I.e. if libmongocrypt gets `"schemaRequiresEncryption": true` it does not know which of the JSON Schemas required encryption).

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -426,8 +426,7 @@ static bool _set_schema_from_collinfo(mongocrypt_ctx_t *ctx, bson_t *collinfo) {
                     return _mongocrypt_ctx_fail_w_msg(ctx, "malformed $jsonSchema");
                 }
                 found_jsonschema = true;
-            } else {
-                ectx->collinfo_has_siblings = true;
+                break;
             }
         }
     }
@@ -800,15 +799,6 @@ static bool _mongo_feed_markings(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *in)
                             "local schema used but does not have encryption specifiers");
         }
         return true;
-    } else {
-        /* if the schema requires encryption, but has sibling validators, error.
-         */
-        if (ectx->collinfo_has_siblings) {
-            return _mongocrypt_ctx_fail_w_msg(ctx,
-                                              "schema requires encryption, "
-                                              "but collection JSON schema "
-                                              "validator has siblings");
-        }
     }
 
     if (bson_iter_init_find(&iter, &as_bson, "hasEncryptedPlaceholders") && !bson_iter_as_bool(&iter)) {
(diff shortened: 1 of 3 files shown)


rcsanchez97 (approved):
LGTM

PR #954, opened 2025-02-14
MONGOCRYPT-723 support `$lookup`

kevinAlbs:
# Summary

Support encrypting "aggregate" commands with `$lookup`

Tested with libmongocrypt patch: https://spruce.mongodb.com/version/67af62d2eff4dc0007490c8f
Tested with C driver patch pointing to this branch: https://spruce.mongodb.com/version/67af5960bad7a30007a37b90

# Background & Motivation

## Parsing

Enough of the aggregate pipeline is parsed to find occurrences of `$lookup`. `$lookup` can be nested within `$lookup`, `$facet`, and `$unionWith`.

mongocryptd/crypt_shared return [an error for `$unionWith` and `$facet`](https://github.com/kevinAlbs/iue-bootstrap/blob/219aa33638498ff835b67514ef25b9c43f3bd9ee/tools/generate-markings/tests/lookup-csfle-unionWith.yml#L4-L5): `Aggregation stage $unionWith is not allowed or supported with automatic encryption.`. Regardless, this PR parses within `$unionWith` and `$facet` for completeness.

Parsing in libmongocrypt may require future changes. If a future server version extends the aggregate pipeline, libmongocrypt may not find all `$lookup` stages. However, if libmongocrypt does not pass a schema for a needed collection, mongocryptd/crypt_shared errors (e.g. `Missing encryption schema for namespace: db.c2`) rather than assumes the collection has no schema. This was decided acceptable in [Technical Design: Support $lookup in CSFLE and QE](https://docs.google.com/document/d/1HQKM6s8naGFcArVpBgZSnKCWANJ7qhKm_QFM_iFY1TI/edit?tab=t.0#heading=h.6568vliqavvt):

## mongocryptd/crypt_shared version check

A version check is added to ensure mongocryptd/crypt_shared supports multiple schemas. This is intended to improve the error that would otherwise be returned from a too-old mongocryptd/crypt_shared. Testing mongocryptd 8.0 gets the following errors:
- `jsonSchema or encryptionInformation is required` when passing multiple CSFLE schemas.
- `Exactly one namespace is supported with encryptionInformation` when passing multiple QE schemas.

Instead, libmongocrypt returns an error like the following:
- `Encrypting 'aggregate' requires multiple schemas. Detected mongocryptd with wire version 17, but need 26. Upgrade mongocryptd to 8.1 or newer.` when using mongocryptd
- `Encrypting 'aggregate' requires multiple schemas. Detected crypt_shared with version 8.0.4, but need 8.1. Upgrade crypt_shared to 8.1 or newer.` when using crypt_shared

## Opt-in to support multi-collection commands

The state `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` requests the driver send `listCollections` to check for server-side schemas. Before this PR, this state only expected at most one result from `listCollections`. Quoting [integrating.md](https://github.com/mongodb/libmongocrypt/blob/c3dff011c9c0dc326c31414471a4b613e0d91365/integrating.md#state-mongocrypt_ctx_need_mongo_collinfo):

> Return the first result (if any) with `mongocrypt_ctx_mongo_feed`

To support `$lookup`, drivers are now expected to return all results from `listCollections`.

libmongocrypt cannot distinguish between "did not pass all results" (following old protocol) and "server did not have results". libmongocrypt applies empty schemas to collections that have no known schema. If a driver upgrades libmongocrypt and does not implement the new protocol, an empty schema might be applied to a collection that really has a server-side schema.

To ensure drivers upgrading libmongocrypt follow the new protocol, drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to enable support for multi-collection commands. Without opting-in, libmongocrypt returns an error if a command requires multiple collections. This is to avoid the risk that a driver upgrades libmongocrypt without implementing the new protocol.

## Increasing `TEST_DATA_COUNT`

Increasing `TEST_DATA_COUNT` was done to support the new tests. Increasing too high resulted in a segfault on Windows (too large stack frame?). Filed MONGOCRYPT-775 to investigate separately (not strictly needed for this PR).

### DIFF
--- a/integrating.md
+++ b/integrating.md
@@ -144,7 +144,9 @@ All contexts.
 > [!IMPORTANT]
 > **Multi-collection commands**: prior to 1.13.0, drivers were expected to pass _at most one result_ from `listCollections`. In 1.13.0, drivers are expected to pass _all results_ from `listCollections` to support multi-collection commands (e.g. aggregate with `$lookup`).
 >
-> Drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and enable support for multi-collection commands.
+> Drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and opt-in to support for multi-collection commands.
+>
+> The opt-in is avoids the following example scenario: Driver upgrades to 1.13.0 and does not update the behavior. A multi-collection command requests schemas for both `db.c1` and `db.c2`. Driver only passes the result for `db.c1` (even though `db.c2` has a result). libmongocrypt mistakenly assumes `db.c2` has no schema.
 
 **libmongocrypt needs**...
 
--- a/integrating.md
+++ b/integrating.md
@@ -140,6 +140,14 @@ All contexts.
 
 #### State: `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` ####
 
+##### Multi-collection commands
+Prior to 1.13.0, drivers were expected to return at most one result from `listCollections`. In 1.13.0, drivers are
+expected to return all results to support multi-collection commands (e.g. aggregate with `$lookup`).
+
+libmongocrypt cannot distinguish between "did not pass all results" and "server did not have results". Drivers must call
+`mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and enable support for
+multi-collection commands.
+
 **libmongocrypt needs**...
 
 A result from a listCollections cursor.
@@ -148,7 +156,7 @@ A result from a listCollections cursor.
 
 1.  Run listCollections on the encrypted MongoClient with the filter
     provided by `mongocrypt_ctx_mongo_op`
-2.  Return the first result (if any) with `mongocrypt_ctx_mongo_feed` or proceed to the next step if nothing was returned.
+2.  Return all results (if any) with calls to `mongocrypt_ctx_mongo_feed` or proceed to the next step if nothing was returned.
 3.  Call `mongocrypt_ctx_mongo_done`
 
 **Applies to...**
@@ -156,6 +164,8 @@ A result from a listCollections cursor.
 auto encrypt
 
 #### State: `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO_WITH_DB` ####
+
+See [note](#multi-collection-commands) about multi-collection commands.
 
 **libmongocrypt needs**...
 
@@ -165,7 +175,7 @@ Results from a listCollections cursor from a specified database.
 
 1.  Run listCollections on the encrypted MongoClient with the filter
     provided by `mongocrypt_ctx_mongo_op` on the database provided by `mongocrypt_ctx_mongo_db`.
-2.  Return the first result (if any) with `mongocrypt_ctx_mongo_feed` or proceed to the next step if nothing was returned.
+2.  Return all results (if any) with calls to `mongocrypt_ctx_mongo_feed` or proceed to the next step if nothing was returned.
 3.  Call `mongocrypt_ctx_mongo_done`
 
 **Applies to...**
--- a/integrating.md
+++ b/integrating.md
@@ -143,9 +143,11 @@ All contexts.
 > [!IMPORTANT]
 > <a name="multi-collection-commands"></a> **Multi-collection commands**: prior to 1.13.0, drivers were expected to pass _at most one result_ from `listCollections`. In 1.13.0, drivers are expected to pass _all results_ from `listCollections` to support multi-collection commands (e.g. aggregate with `$lookup`).
 >
-> Drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and opt-in to support for multi-collection commands.
->
-> The opt-in is avoids the following example scenario: Driver upgrades to 1.13.0 and does not update the behavior. A multi-collection command requests schemas for both `db.c1` and `db.c2`. Driver only passes the result for `db.c1` (even though `db.c2` has a result). libmongocrypt mistakenly assumes `db.c2` has no schema.
+> Drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and opt-in to support for multi-collection commands. This opt-in is to prevent the following bug scenario:
+> > A driver upgrades to 1.13.0, but does not update prior behavior which passes at most one result of a multi-collection command.
+> > A multi-collection command requests schemas for both `db.c1` and `db.c2`.
+> > The driver only passes the result for `db.c1` even though `db.c2` also has a result.
+> > Therefore, libmongocrypt incorrectly believes `db.c2` has no schema.
 
 **libmongocrypt needs**...
 
--- a/integrating.md
+++ b/integrating.md
@@ -140,9 +140,8 @@ All contexts.
 
 #### State: `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` ####
 
-<a name="multi-collection-commands"></a>
 > [!IMPORTANT]
-> **Multi-collection commands**: prior to 1.13.0, drivers were expected to pass _at most one result_ from `listCollections`. In 1.13.0, drivers are expected to pass _all results_ from `listCollections` to support multi-collection commands (e.g. aggregate with `$lookup`).
+> <a name="multi-collection-commands"></a> **Multi-collection commands**: prior to 1.13.0, drivers were expected to pass _at most one result_ from `listCollections`. In 1.13.0, drivers are expected to pass _all results_ from `listCollections` to support multi-collection commands (e.g. aggregate with `$lookup`).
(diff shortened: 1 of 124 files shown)


src/mc-schema-broker.c
```diff
@@ -93,6 +93,11 @@ bool mc_schema_broker_request(mc_schema_broker_t *sb, const char *db, const char
     return true;
 }
 
+bool mc_schema_broker_has_multiple_requests(const mc_schema_broker_t *sb) {
+    BSON_ASSERT_PARAM(sb);
+    return sb->ll_len > 1;
```
eramongodb, line 98: Suggest adding assertions of `mc_schema_broker_has_multiple_requests` in test cases for `mc_schema_broker_request` which ensure adding a request for the same collection (duplicates) preserves the "unique list of collections" invariant which this function requires for correctness.

test/test-mongocrypt-csfle-lib.c
```diff
@@ -170,6 +170,26 @@ static void _test_override_error_includes_reason(_mongocrypt_tester_t *tester) {
...
+
+#define CRYPT_SHARED_8_1 0x0008000100000000ull
+    mongocrypt_t *crypt = _mongocrypt_tester_mongocrypt(TESTER_MONGOCRYPT_WITH_CRYPT_SHARED_LIB);
+    uint64_t version = crypt->csfle.get_version();
+    mongocrypt_ctx_t *ctx = mongocrypt_ctx_new(crypt);
+    mongocrypt_binary_t *cmd = TEST_FILE("./test/data/lookup/csfle/01-cmd.json");
+    if (version >= CRYPT_SHARED_8_1) {
+        ASSERT_OK(mongocrypt_ctx_encrypt_init(ctx, "db", -1, cmd), ctx);
+    } else {
+        ASSERT_FAILS(mongocrypt_ctx_encrypt_init(ctx, "db", -1, cmd), ctx, "Upgrade crypt_shared");
```
eramongodb, line 187: Confirming: this is just a convenient top-level test suite requirement. Runtime crypt_shared compatibility requirements are checked and reported via error handling in `mongocrypt_ctx_encrypt_ismaster_done`. Is this correct?

test/test-mongocrypt-ctx-encrypt.c
```diff
@@ -4856,6 +4855,944 @@ static void _test_fle2_collinfo_with_bad_str_encode_version(_mongocrypt_tester_t
...
+        mongocrypt_ctx_destroy(ctx);
+        mongocrypt_destroy(crypt);
+    }
+#undef TF
+
+    // Test $lookup errors if multiple-collection support is not opted-in.
+#define TF(suffix) TEST_FILE("./test/data/lookup/csfle/" suffix)
+    {
+        mongocrypt_t *crypt = _mongocrypt_tester_mongocrypt(TESTER_MONGOCRYPT_DEFAULT);
+        crypt->multiple_collinfo_enabled = false;
```
eramongodb, line 4890:
```suggestion
```
This line does not appear to be necessary.

src/mongocrypt-ctx-encrypt.c
```diff
@@ -2367,20 +2540,52 @@ static bool mongocrypt_ctx_encrypt_ismaster_done(mongocrypt_ctx_t *ctx) {
...
+                return true;
+            }
+        }
+
+        if (mc_schema_broker_has_multiple_requests(ectx->sb)) {
+            // Ensure mongocryptd supports multiple schemas.
+            if (ectx->ismaster.maxwireversion < WIRE_VERSION_SERVER_8_1) {
+                mongocrypt_status_t *status = ctx->status;
+                CLIENT_ERR("Encrypting '%s' requires multiple schemas. Detected mongocryptd with wire version %" PRId32
+                           ", but need %" PRId32 ". Upgrade mongocryptd to 8.1 or newer.",
```
eramongodb, line 2565: Confirming, does the wire version number for mongocryptd always match the wire version of the MongoDB server? If not, should this error message refer to the server instead of mongocryptd?

src/mongocrypt-ctx-encrypt.c
```diff
@@ -2341,11 +2513,8 @@ bool mongocrypt_ctx_encrypt_init(mongocrypt_ctx_t *ctx, const char *db, int32_t
         bson_free(cmd_val);
     }
 
-    /* The "create" and "createIndexes" command require sending an isMaster
-     * request to mongocryptd. */
+    // Check if an isMaster request to mongocryptd is needed check for feature support:
```
eramongodb, line 2516:
```suggestion
    // Check if an isMaster request to mongocryptd is needed to detect feature support:
```

src/mongocrypt-ctx-encrypt.c
```diff
@@ -2220,9 +2220,165 @@ static bool needs_ismaster_check(mongocrypt_ctx_t *ctx) {
...
+    if (!bson_iter_init_find(&iter, &cmd_bson, "pipeline")) {
+        // Command may be malformed. Let server error.
+        return true;
+    }
+
+    if (!find_collections_in_pipeline(sb, iter, db, mstrv_lit("aggregate.pipeline"), status)) {
+        return false;
+    }
+
+    return true;
```
eramongodb, line 2381:
```suggestion
    return find_collections_in_pipeline(sb, iter, db, mstrv_lit("aggregate.pipeline"), status);
```
Minor simplification.

eramongodb (changes requested):

integrating.md
```diff
@@ -140,6 +140,14 @@ All contexts.
 
 #### State: `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO` ####
 
+<a name="multi-collection-commands"></a>
+> [!IMPORTANT]
+> **Multi-collection commands**: prior to 1.13.0, drivers were expected to pass _at most one result_ from `listCollections`. In 1.13.0, drivers are expected to pass _all results_ from `listCollections` to support multi-collection commands (e.g. aggregate with `$lookup`).
```
eramongodb, line 145:
```suggestion
> [!IMPORTANT]
> <a name="multi-collection-commands"></a> **Multi-collection commands**: prior to 1.13.0, drivers were expected to pass _at most one result_ from `listCollections`. In 1.13.0, drivers are expected to pass _all results_ from `listCollections` to support multi-collection commands (e.g. aggregate with `$lookup`).
```
Looks like the current anchor placement is being treated as its own paragraph. Suggest moving it into the admonition block to fix jump navigation.

integrating.md
```diff
@@ -140,6 +140,14 @@ All contexts.
...
+>
+> Drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and opt-in to support for multi-collection commands.
+>
+> The opt-in is avoids the following example scenario: Driver upgrades to 1.13.0 and does not update the behavior. A multi-collection command requests schemas for both `db.c1` and `db.c2`. Driver only passes the result for `db.c1` (even though `db.c2` has a result). libmongocrypt mistakenly assumes `db.c2` has no schema.
```
eramongodb, line 149:
```suggestion
> Drivers must call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented and opt-in to support for multi-collection commands. This opt-in is to prevent the following bug scenario:
> > A driver upgrades to 1.13.0, but does not update prior behavior which passes at most one result of a multi-collection command.
> > A multi-collection command requests schemas for both `db.c1` and `db.c2`.
> > The driver only passes the result for `db.c1` even though `db.c2` also has a result.
> > Therefore, libmongocrypt incorrectly believes `db.c2` has no schema.
```
Minor formatting suggestions.

test/test-mc-schema-broker.c
```diff
@@ -40,7 +40,9 @@ static void test_mc_schema_broker_request(_mongocrypt_tester_t *tester) {
         mongocrypt_status_t *status = mongocrypt_status_new();
         mc_schema_broker_t *sb = mc_schema_broker_new();
         ASSERT_OK_STATUS(mc_schema_broker_request(sb, "db", "coll1", status), status);
+        ASSERT(!mc_schema_broker_has_multiple_requests(sb));
```
eramongodb, line 43: Suggest also asserting the `len == 0` case prior to the first call to `mc_schema_broker_request` both here (0 -> 1 -> 2) and below (0 -> 1 -> 1 (duplicate)).

eramongodb (approved):

test/test-mongocrypt-ctx-encrypt.c
kevinAlbs: Oop. I missed this comment on first pass. This line is needed since `_mongocrypt_tester_mongocrypt` enables the opt-in with a call to `mongocrypt_setopt_enable_multiple_collinfo`. This test expects an error due to the opt-in being disabled.

eramongodb: It turns out my suggestion was due to a local test execution misconfiguration. Disregard, but also thank you for the explanation.

PR #926, opened 2024-12-19
MONGOCRYPT-731 re-enable ASan tasks

kevinAlbs:
# Summary
- Fix Evergreen config to apply `-fsanitize=address`.
- Suppress `dlerror` leak. Fix test leaks.
- Restore default of `ENABLE_EXTRA_ALIGNMENT=OFF` to imported libbson.

Verified by this [patch build](https://spruce.mongodb.com/version/676571030624120007db3430/).

As a drive-by fix, the unused `build-and-test-asan-s390x` is removed.

# Background

https://github.com/mongodb/libmongocrypt/pull/622 accidentally disabled ASan due to a bash comment (`# Add ...`) appearing in a folded (`>-`) YAML block:

```yaml
compile_env: >-
    ${compile_env|}
    LIBMONGOCRYPT_EXTRA_CFLAGS="-fsanitize=address -pthread"
    # Add detect_odr_violation=0 to ASAN_OPTIONS to ignore odr-violation in IntelDFP symbol: __dpml_bid_globals_table
    ASAN_OPTIONS="detect_leaks=1 detect_odr_violation=0"
```

This PR moves the comment outside of the block to fix:
```yaml
# Add detect_odr_violation=0 to ASAN_OPTIONS to ignore odr-violation in IntelDFP symbol: __dpml_bid_globals_table
compile_env: >-
    ${compile_env|}
    LIBMONGOCRYPT_EXTRA_CFLAGS="-fsanitize=address -pthread"
    ASAN_OPTIONS="detect_leaks=1 detect_odr_violation=0"
```

Re-enabling ASan revealed an [Indirect leak](https://spruce.mongodb.com/task/libmongocrypt_ubuntu2204_arm64_build_and_test_asan_patch_66f3c6a4771786ba221a58cc638ad1837027477b_67226c1571b6b30007305a63_24_10_30_17_25_42/logs?execution=0) reported in `dlerror` only on `Ubuntu 22.04 arm64`. The leak is now suppressed. I expect it is not an issue in libmongocrypt.

[...]

PR #971, opened 2025-02-28
MONGOCRYPT-784 avoid passing over-aligned bson_iter_t by value in parameters

eramongodb:
Resolves MONGOCRYPT-784.

Followup to https://github.com/mongodb/libmongocrypt/pull/954, which introduced a by-value `bson_iter_t` parameter for [find_collections_in_pipeline](https://github.com/mongodb/libmongocrypt/blob/33fdf65cce5a0c0cdd293c64ed40e4a8205c3ce0/src/mongocrypt-ctx-encrypt.c#L2242) which triggers the following MSVC warning when `ENABLE_EXTRA_ALIGNMENT` is enabled for the bson library:

```
error C2719: 'pipeline_iter': formal parameter with requested alignment of 128 won't be aligned
```

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -2246,7 +2246,6 @@ static bool find_collections_in_pipeline(mc_schema_broker_t *sb,
     BSON_ASSERT_PARAM(sb);
     BSON_ASSERT_PARAM(pipeline_iter_ptr);
     BSON_ASSERT_PARAM(db);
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Nice catch, and thank you for the fix!

As of https://github.com/mongodb/libmongocrypt/pull/926, enabling extra alignment is not the default for the bundled libbson. But I expect this could be triggered if using a system install of an over-aligned libbson. I request backporting the merged commit to r1.13 for the next patch release.

PR #975, opened 2025-02-28
MONGOCRYPT-783 address miscellaneous warnings

eramongodb:
Related to MONGOCRYPT-783.

Addresses miscellaneous warnings which do not have any particular common theme. These are all Clang warnings:

- `-Wold-style-cast`: 213 unique, 231 total.
- `-Wmissing-prototypes`: 116 unique, 155 total.
- `-Wstrict-prototypes`: 24 unique, 618 total.
- `-Wundef`: 20 unique, 321 total.
- `-Wmissing-variable-declarations`: 20 unique, 32 total.
- `-Wassign-enum`: 10 unique, 10 total.
- `-Wunused-macros`: 7 unique, 12 total.
- `-Wunreachable-code-break`: 3 unique, 5 total.
- `-Wunreachable-code-return`: 1 unique, 4 total.
- `-Wvla`: 1 unique, 1 total.

---

`-Wold-style-cast` warnings are generated when compiling `.cpp` components for C++ compatibility and testing purposes. Since the cast <-> compound literal equivalence is deliberate, these warnings are simply ignored.

---

Both `-Wmissing-prototype` and `-Wmissing-variable-declarations` help inform when a given function or variable may be declared `static` or if the header providing the declaration has not been included as it should be. All such symbols (which may be declared `static`) and components (missing an include directive) have been updated accordingly. Functions and variables which are deliberately not forward-declared in a header are denoted by a preceeding declaration labeled with `// -Wmissing-prototypes: <reason>`.

---

`-Wundef` warnings are addressed for external macros (i.e. `_WIN32`) by using `#ifdef` and `defined`, or by guarding their use with a definition check. For our own macros (e.g. `MONGOCRYPT_HAVE_DECIMAL128_SUPPORT`), the suggestion in https://github.com/mongodb/mongo-cxx-driver/pull/1333#discussion_r1954786256 is applied to catch cases where the required header providing said macro definition may not be included as it should. This revealed a missing include directive in `mc-fle2-payload-iev-v2.c` which impacts the behavior of `is_fle2_range_indexed_supported_type()`.


kms-message/test/test_kmip_reader_writer.c
```diff
@@ -6,7 +6,7 @@
  * You may obtain a copy of the License at
  *
  *   http://www.apache.org/licenses/LICENSE-2.0
- *
+
```
vector-of-bool, line 9: lost star

```suggestion
 *
```

vector-of-bool (approved):
LGTM

PR #976, opened 2025-03-02
MONGOCRYPT-786 Fix incompatible type error

songkomkrit:
Building libmongocrypt on Linux from the current master commit 8270531 was unsuccessful. The following error was produced.

````{verbatim, lang = "markdown"}
[ 22%] Building C object CMakeFiles/mongocrypt.dir/src/mongocrypt-ctx-encrypt.c.o
In file included from libmongocrypt/src/mongocrypt-opts-private.h:27,
                 from libmongocrypt/src/mongocrypt-key-private.h:22,
                 from libmongocrypt/src/mongocrypt-cache-key-private.h:22,
                 from libmongocrypt/src/mongocrypt-key-broker-private.h:24,
                 from libmongocrypt/src/mongocrypt-ctx-private.h:26,
                 from libmongocrypt/src/mongocrypt-ctx-encrypt.c:22:
libmongocrypt/src/mongocrypt-ctx-encrypt.c: In function ‘_mongo_op_collinfo’:
libmongocrypt/src/mongocrypt-ctx-encrypt.c:43:74: error: incompatible type for argument 1 of ‘tmp_json’
   43 | RACEF(&ectx->parent.crypt->log, "constructed: %s\n", tmp_json(filter));
      |                                                               ^~~~~~
      |                                                               |
      |                                                               bson_t {aka struct _bson_t}
libmongocrypt/src/mongocrypt-log-private.h:44:91: note: in definition of macro ‘CRYPT_TRACEF’
   44 | CRYPT_LOG_LEVEL_TRACE, "(%s:%d) " fmt, BSON_FUNC, __LINE__, __VA_ARGS__)
      |                                                             ^~~~~~~~~~~
In file included from libmongocrypt/src/mongocrypt-ctx-private.h:28:
libmongocrypt/src/mongocrypt-private.h:57:36: note: expected ‘const bson_t *’ {aka ‘const struct _bson_t *’} but argument is of type ‘bson_t’ {aka ‘struct _bson_t’}
   57 | const char *tmp_json(const bson_t *bson);
      |                      ~~~~~~~~~~~~~~^~~~
make[2]: *** [CMakeFiles/mongocrypt.dir/build.make:639: CMakeFiles/mongocrypt.dir/src/mongocrypt-ctx-encrypt.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1296: CMakeFiles/mongocrypt.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
````

This problem can be fixed by a minor change in this pull request. The caller ``tmp_json`` in ``src/mongocrypt-ctx-encrypt.c`` must take the address ``&filter``, not the object ``filter`` because a pointer argument is passed to the function pointer ``tmp_json`` defined in ``src/mongocrypt-private.h``. Now it is successful to build libmongocrypt.

### DIFF
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -40,7 +40,6 @@ static bool _mongo_op_collinfo(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out) 
         _mongocrypt_ctx_fail(ctx);
         return false;
     }
-    CRYPT_TRACEF(&ectx->parent.crypt->log, "constructed: %s\n", tmp_json(&filter));
     _mongocrypt_buffer_steal_from_bson(&ectx->list_collections_filter, &filter);
     out->data = ectx->list_collections_filter.data;
     out->len = ectx->list_collections_filter.len;
--- a/src/mongocrypt-ctx-encrypt.c
+++ b/src/mongocrypt-ctx-encrypt.c
@@ -40,7 +40,7 @@ static bool _mongo_op_collinfo(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out) 
         _mongocrypt_ctx_fail(ctx);
         return false;
     }
-    CRYPT_TRACEF(&ectx->parent.crypt->log, "constructed: %s\n", tmp_json(filter));
+    CRYPT_TRACEF(&ectx->parent.crypt->log, "constructed: %s\n", tmp_json(&filter));
     _mongocrypt_buffer_steal_from_bson(&ectx->list_collections_filter, &filter);
     out->data = ectx->list_collections_filter.data;
     out->len = ectx->list_collections_filter.len;

### CODE REVIEW

File: src/mongocrypt-ctx-encrypt.c [hunk 1]
```diff
@@ -40,7 +40,7 @@ static bool _mongo_op_collinfo(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *out)
         _mongocrypt_ctx_fail(ctx);
         return false;
     }
-    CRYPT_TRACEF(&ectx->parent.crypt->log, "constructed: %s\n", tmp_json(filter));
+    CRYPT_TRACEF(&ectx->parent.crypt->log, "constructed: %s\n", tmp_json(&filter));
```
kevinAlbs, line 43:
```suggestion
```

Suggest removing this trace line entirely. `tmp_json` uses [static data](https://github.com/mongodb/libmongocrypt/blob/69d9984c76a41ca32d650d44317ee57ab99707df/src/mongocrypt.c#L71) and is unsafe to call from multiple threads. I do not think this trace is well motivated (callers can check the produced filter).

kevinAlbs (commented):
Thank you for the fix @songkomkrit.

This only appears if libmongocrypt is configured with trace logging. Trace logging is enabled by configuring with the CMake option `ENABLE_TRACE=ON`.

MONGOCRYPT-791 was relatedly filed to track testing trace logging in CI or removing entirely. Is trace logging needed for your use-case? That may help inform a decision in MONGOCRYPT-791.

src/mongocrypt-ctx-encrypt.c
songkomkrit: Thank you for taking care of this issue @kevinAlbs.

I forgot to tell you that I built libmongocrypt with ``ENABLE_TRACE=ON`` and that is why it was unsuccessful. I rebuild it again today by disabling trace logging through the option ``ENABLE_TRACE=OFF`` and it works. The final decision is up to you. The post above is simply based on my observation.

kevinAlbs: Thank you. Applying suggested change to remove log line.

PR #981, opened 2025-03-05
MONGOCRYPT-738 Improve error on failure to load crypt_shared

kevinAlbs:
# Summary

- Improve error message when loading incorrect library for crypt_shared.
- Change "CSFLE" to "crypt_shared" in log and error messages to match the library rename in MONGOCRYPT-434.

Verified with this [Evergreen patch](https://spruce.mongodb.com/version/67c85a9a61fd92000765f040).

# Background & Motivation

https://jira.mongodb.org/browse/MONGOCRYPT-738 references several reports of users accidentally passing the path to libmongocrypt instead of the crypt_shared library. This produces an error message like:

```
A CSFLE override path was specified [./cmake-build/libmongocrypt.dylib], but we failed to open a dynamic library at that location. Load error: [(null)]
```

With this PR, accidentally passing a path to libmongocrypt results in an error like:
```
Tried to load crypt_shared dynamic library at path [./cmake-build/libmongocrypt.dylib] but detected libmongocrypt
```

Attempting to load a different library (not libmongocrypt or crypt_shared) results in an error like:
```
A crypt_shared override path was specified [./cmake-build/libmongocrypt.dylib], but we failed to open a dynamic library at that location. Load error: [One or more required symbols are missing from crypt_shared dynamic library [/Users/kevin.albertson/code/tasks/libmongocrypt-M738/./cmake-build/libmongocrypt.dylib], so this dynamic library will not be used.]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -533,6 +533,8 @@ target_compile_definitions (test-mongocrypt PRIVATE
    "TEST_MONGOCRYPT_OUTPUT_PATH=\"$<TARGET_FILE:test-mongocrypt>\""
    # Tell test-mongocrypt whether we have a real csfle library for testing
    TEST_MONGOCRYPT_HAVE_REAL_CRYPT_SHARED_LIB=$<BOOL:${MONGOCRYPT_TESTING_CRYPT_SHARED_FILE}>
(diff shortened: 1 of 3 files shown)


vector-of-bool (approved):
LGTM

PR #985, opened 2025-03-07
MONGOCRYPT-788 Define FLE2FindTextPayload type

erwee:


test/test-mc-fle2-find-text-payload.c
```diff
@@ -0,0 +1,320 @@
...
+}
+
+static const char *k_edcToken = "280edab4190763d1f4183d383f8830773859c3a74a13161094e6fd44e7390fed";
+static const char *k_escToken = "c3c8980ed11e63ec199104bc9a0889322c9eb80d00eee0b5148dc83f7e78adb7";
+static const char *k_svrToken = "8773322a2b9e6c08886db6bc65b46ffdd64651e8a49400a9e55ff5bc550d45bf";
+
+// Tests mc_FLE2FindTextPayload_serialize() works correctly, and non-error cases
+// for mc_FLE2FindTextPayload_parse().
+static void _test_FLE2FindTextPayload_roundtrip(_mongocrypt_tester_t *tester) {
+    // Test exact token set + no specs present
```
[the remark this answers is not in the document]
erwee: Comment revised.

kevinAlbs (approved):

src/mc-fle2-find-text-payload-private.h
```diff
@@ -0,0 +1,122 @@
...
+ *   ts: {
+ *      e: optional<TextExactFindTokenSet>      // required tokens if doing full string match
+ *      s: optional<TextSubstringFindTokenSet>  // required tokens if doing substring match
+ *      u: optional<TextSuffixFindTokenSet>     // required tokens if doing suffix match
+ *      p: optional<TextPrefixFindTokenSet>     // required tokens if doing prefix match
+ *   }
+ *   cm: <int64>                                // Queryable Encryption max contentionFactor
+ *   cf: <bool>                                 // case folding parameter
+ *   df: <bool>                                 // diacritic folding parameter
+ *   ss: optional<FLE2SubstringInsertSpec>      // substring-indexing parameters (if applicable)
```
marksg07, line 81: Should we rename FLE2<Substring|Suffix|Prefix>InsertSpec to FLE2<Substring|Suffix|Prefix>Spec?

src/mc-fle2-find-text-payload.c
```diff
@@ -0,0 +1,477 @@
...
+    }                                                                                                                  \
+    END_IF_FIELD
+
+typedef struct {
+    _mongocrypt_buffer_t *edcDerivedToken;
+    _mongocrypt_buffer_t *escDerivedToken;
+    _mongocrypt_buffer_t *serverDerivedToken;
+} mc_TextFindTokenSetIndirection_t;
+
+typedef struct {
```
marksg07, line 58: Why do we need this? Can't we just use `const mc_TextFindTokenSetIndirection_t`?

src/mc-fle2-find-text-payload-private.h
```diff
@@ -0,0 +1,122 @@
...
+
+#ifndef MC_FLE2_FIND_TEXT_PAYLOAD_PRIVATE_H
+#define MC_FLE2_FIND_TEXT_PAYLOAD_PRIVATE_H
+
+#include "mc-fle2-encryption-placeholder-private.h"
+#include "mongocrypt-buffer-private.h"
+#include "mongocrypt-private.h"
+#include "mongocrypt.h"
+
+#define DEF_TEXT_SEARCH_FIND_TOKEN_SET(Type)                                                                           \
```
marksg07, line 25: Since these are all the same, can't we just have one `mc_TextSearchFindTokenSet_t` type? If we /really/ want the split, we could just `typedef mc_TextSearchFindTokenSet_t mc_TextExactFindTokenSet_t`, etc. But that seems unnecessary.

marksg07 (changes requested):
Sorry for the delay on the review! I think we can coalesce `mc_Text*FindTokenSet_t` into a single type to simplify the code.

src/mc-fle2-find-text-payload-private.h
erwee: I want these to be distinct types, so that you won't be able to mindlessly assign one type to the other without the compiler yell at you.

src/mc-fle2-find-text-payload-private.h
marksg07: Ok, as long as that's logged somewhere.

src/mc-fle2-find-text-payload.c
marksg07: I see, thanks for the explanation.

src/mc-fle2-find-text-payload-private.h
marksg07: Sure... Even in that case, I think we should simplify some logic in the .c file.

src/mc-fle2-find-text-payload.c
```diff
@@ -0,0 +1,477 @@
...
+
+#define INDIRECT(type, ts)                                                                                             \
+    (type) {                                                                                                           \
+        .edcDerivedToken = &(ts).edcDerivedToken, .escDerivedToken = &(ts).escDerivedToken,                            \
+        .serverDerivedToken = &(ts).serverDerivedFromDataToken                                                         \
+    }
+#define INDIRECT_TOKENSET(ts) INDIRECT(mc_TextFindTokenSetIndirection_t, ts)
+#define INDIRECT_TOKENSET_CONST(ts) INDIRECT(mc_TextFindTokenSetIndirectionConst_t, ts)
+
+#define DEF_TEXT_SEARCH_FIND_TOKEN_SET_CLEANUP(Type)                                                                   \
```
marksg07, line 151: I don't like defining all these macros just to make a bunch of functions that are scoped to this .c file. I also don't like that we are copying the pointers into a new Indirection_t struct to get around casting -- we are relying on the C compiler to optimize this away, but I'm not convinced that it always will. My suggestion:

```
// Replace Indirection_t and IndirectionConst_t with
typedef struct {
    _mongocrypt_buffer_t edcDerivedToken;                                                                          
    _mongocrypt_buffer_t escDerivedToken;                                                                          
    _mongocrypt_buffer_t serverDerivedFromDataToken;
} mc_FindTokenSet_t;

// Redefine Indirection_cleanup, _serialize, _parse using pointers/const pointers to this struct
// ...

#define CLEANUP_FIND_TOKEN_SET(fts) mc_FindTokenSet_cleanup((mc_FindTokenSet_t *)fts)
#define SERIALIZE_FIND_TOKEN_SET(fts, parent, field_name) mc_FindTokenSet_serialize((const mc_FindTokenSet_t *)fts, parent, field_name)
#define PARSE_FIND_TOKEN_SET(fts, type, in, status) mc_FindTokenSet_parse((mc_FindTokenSet_t *)out, "Text" type "FindTokenSet", in, status)

// Use these macros directly in following code.
```

This way, we eliminate the Indirection/IndirectionConst split, save the copy, reduce the macro usage, and end up with much less code.

erwee: I tried doing it that way at first, but decided against it because of the unsafe cast and lack of compile-time checks. While we can guarantee safe cast using static asserts, I'd rather just copy pointers.

marksg07: I mean, sure it's an unsafe cast that lacks compile-time checks, but that's kinda just C for you. Since we are only using these macros within the scope of this file, we can ensure that we are using it correctly in the few places we do use it. I think it's worth it to save a bunch of LOC and confusing macros in order to preserve readability.

erwee: > I mean, sure it's an unsafe cast that lacks compile-time checks, but that's kinda just C for you.

Try telling that to the trail of bits folks : ).

I'm still not convinced we should trade type safety for saving several lines of macros. Like, it's very easy to shoot yourself in the foot if, for instance, you pass the pointer to an ExactFindTokenSet to `SERIALIZE_FIND_TOKEN_SET` when it should've been a `SubstringFindTokenSet`.

marksg07: Ok, I think if you’re satisfied with this approach, let’s leave it as is and move on. No point of thinking any more about this one.

marksg07 (approved):

PR #911, opened 2024-11-18
Use Artifactory for image registry

kevinAlbs:
This PR is intended to address rate-limiting errors pulling from dockerhub causing task failures on Evergreen. [Example](https://spruce.mongodb.com/task/libmongocrypt_alpine_amd64_earthly_build_with_earthly_b150e745c573793e7d27568f4be2d6f83d879f82_24_11_15_14_38_07/logs?execution=0):

> docker: Error response from daemon: toomanyrequests: Too Many Requests (HAP429).

[DEVPROD-12779](https://jira.mongodb.org/browse/DEVPROD-12779) [notes](https://jira.mongodb.org/browse/DEVPROD-12779?focusedCommentId=6856395&focusedId=6856395&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6856395):

> In the meantime there is an artifactory mirror for dockerhub that can be used for pull containers.

Running `docker pull artifactory.corp.mongodb.com/dockerhub/<DOCKER_IMAGE>:<DOCKER_TAG>` appears to pull and add the image to the artifactory registry. DEVPROD-12779 [also notes](https://jira.mongodb.org/browse/DEVPROD-12779?focusedCommentId=6856869&focusedId=6856869&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6856869):

> The proper long term solution seems to be to use authentication for pull requests as suggested by the Docker support.

Getting credentials is tracked in DRIVERS-3047.

As an added fix, the `check-format` target is updated to install `make` to address an observed error:

```
+check-format *failed* | Some possibly relevant errors from pip install:
+check-format *failed* |     error: subprocess-exited-with-error
+check-format *failed* |     CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
+check-format *failed* |     ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (clang-format)
```

Verified with [this patch build](https://spruce.mongodb.com/version/673b690551cfa80007577948/)

PR #987, opened 2025-03-11
MONGOCRYPT-747 Use less Artifactory for Earthly images

kevinAlbs:
Follow-up to https://github.com/mongodb/libmongocrypt/pull/911

Use of Artifactory is likely the cause of frequently encountered HTTP errors ([example](https://spruce.mongodb.com/task/mongo_c_driver_alpine3.17_clang_check:sasl=Cyrus%C2%A0%E2%80%A2%C2%A0tls=OpenSSL%C2%A0%E2%80%A2%C2%A0test_mongocxx_ref=r3.9.0_patch_c186295845d80a6047fd7ba6b7fde56567080ab9_67d03ad1274f720007fd9433_25_03_11_13_29_55/logs?execution=0)). Quoting [slack](https://mongodb.slack.com/archives/C0V7VEU15/p1741630372464459):

> Our Artifactory cluster is just three hosts, and they run really hot.

Using Artifactory is not needed. Quoting [DEVPROD-12970](https://jira.mongodb.org/browse/DEVPROD-12970):
> We're unblocked for now after Docker agreed to add our IPv6 block to their allowlist

Unqualified images names (e.g. `ubuntu:22.04`) are used rather than a qualified image name (e.g. `docker.io/library/ubuntu:22.10`). This may help enable using a different registry for CI and local development. [DEVPROD-15649](https://jira.mongodb.org/browse/DEVPROD-15649) proposes AWS ECR.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -380,13 +380,7 @@ functions:
       params:
         shell: bash
         working_dir: ${working_dir|libmongocrypt}
-        script: |
-          # Authenticate to artifactory.
-          echo "${artifactory_password}" | docker login --password-stdin --username "${artifactory_username}" artifactory.corp.mongodb.com
-          # TODO(MONGOCRYPT-747): remove `--persist-build=false`. 
-          # Pass `--persist-build=false` to avoid using Docker Hub.
-          # Earthly hardcodes use of docker/dockerfile-copy:v0.1.9 for the CACHE command.
-          bash ${workdir}/libmongocrypt/.evergreen/earthly.sh ${args} --persist-build=false
+        script: bash ${workdir}/libmongocrypt/.evergreen/earthly.sh ${args}
 
   sbom:
     - command: ec2.assume_role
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -380,13 +380,7 @@ functions:
       params:
         shell: bash
         working_dir: ${working_dir|libmongocrypt}
-        script: |
-          # Authenticate to artifactory.
-          echo "${artifactory_password}" | docker login --password-stdin --username "${artifactory_username}" artifactory.corp.mongodb.com
```
vector-of-bool, line 385: The login may still be required for Garasign and Silkbomb, although I don't recall if we have CI tasks that run those.

vector-of-bool (approved):
LGTM with one minor question.

.evergreen/config.yml
kevinAlbs: Good point. There is a call of "earthly" Evergreen function that using the Garasign image. The caller authenticates.
Though I expect authenticating may be an easy step to miss if future Evergreen tasks use "earthly" requiring Artifactory. Updated to authenticate as part of the "earthly" Evergreen function, rather than in the caller.

PR #996, opened 2025-04-08
MONGOCRYPT-797 bump libbson to 1.30.3

kevinAlbs:
# Summary

Update downloaded libbson version from 1.28.1 to 1.30.3.

Verified with this patch build: https://spruce.mongodb.com/version/67f51fcc5e1706000725b5ab

# Background & Motivation

See MONGOCRYPT-797. Intended to include CDRIVER-5967 fix Homebrew install of libmongocrypt.

libbson is updated to 1.30 instead of 2.0 to avoid disrupting package releases. The Fedora package [builds with a system install of libbson](https://src.fedoraproject.org/rpms/libmongocrypt/blob/rawhide/f/libmongocrypt.spec#_75) and does not-yet appear to have a libbson package for 2.0 to depend on.

### DIFF
--- a/.evergreen/linker-tests.sh
+++ b/.evergreen/linker-tests.sh
@@ -59,8 +59,9 @@ if is_true USE_NINJA; then
 fi
 
 run_chdir "$MONGOC_DIR" git apply --ignore-whitespace "$linker_tests_deps_root/bson_patches/libbson1.patch"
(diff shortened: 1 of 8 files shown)


eramongodb (approved):
Should this version bump be documented in `CHANGELOG.md`?

kevinAlbs:
> downloaded libbson version from 1.28.1 to 1.30.3.

Good idea. Done.

PR #999, opened 2025-04-16
PYTHON-5188 Make version setting a part of the release process

blink1073:


sleepyStick:
As a fun aside, my cat is sitting on my lap and happened to stretch out his arm over the 'approve' button on my screen right as I was about to approve it. So you also have my cat's LGTM HAHA

blink1073:
> So you also have my cat's LGTM HAHA

Yay!  😹

PR #1016, opened 2025-05-23
MONGOCRYPT-813 publish packages for Ubuntu 24.04

kevinAlbs:
Tested with this patch build: https://spruce.mongodb.com/version/682f7f70b42d5300076c8de4


rcsanchez97 (approved):
LGTM.

PR #1024, opened 2025-06-09
MONGOCRYPT-810 Extend _fle2_append_compactionTokens to append tokens …

marksg07:
…for text search

With updates to compaction for text search, we now need the EncryptionInformation and the anchorPaddingTokenRoot not only for range, but also for all text search types.


erwee (approved):
should a unit test be added/updated for this?

kevinAlbs (approved):

marksg07:
@erwee Good point, updated the compact/cleanup unit tests.

erwee (changes requested):

erwee (approved):

PR #1037, opened 2025-07-17
MONGOCRYPT-821 text indexes for explicit encryption

mdb-ad:
Adds explicit encryption support for textPreview indexes.

- New algorithm type: `mongocrypt_setopt_algorithm(ctx, "textPreview", -1)`
- New query types: `mongocrypt_setopt_query_type(ctx, "{prefix,suffix,substring}Preview", -1)`
- Pass options for text encryption: `mongocrypt_setopt_algorithm_text`


src/mongocrypt.h
```diff
@@ -712,6 +712,9 @@ bool mongocrypt_ctx_setopt_algorithm(mongocrypt_ctx_t *ctx, const char *algorith
 // DEPRECATED: support "RangePreview" has been removed in favor of "range".
 #define MONGOCRYPT_ALGORITHM_RANGEPREVIEW_DEPRECATED_STR "RangePreview"
 #define MONGOCRYPT_ALGORITHM_RANGE_STR "Range"
+#define MONGOCRYPT_ALGORITHM_SUFFIXPREVIEW_STR "suffixPreview"
```
kevinAlbs, line 715: We discussed offline, but I had another realization.

Text fields support multiple query types (currently appears to only be  `suffixPreview` + `prefixPreview` from [tests](https://github.com/10gen/mongo/blob/8413adfa0b2f5b63497201b673c7f4c5ce8a81ef/src/mongo/db/modules/enterprise/jstests/fle2/basic_create_collection_text.js#L227)). Creating an insert payload needs to support specifying all supported query types. And each of the `suffixPreview` and `prefixPreview` could have different sub-options (e.g. different `strMaxQueryLength`).

Proposal:
- Use a generic "text" value for `mongocrypt_setopt_algorithm`.
- Support specifying all prefix/suffix/substring options in `mongocrypt_ctx_setopt_algorithm_text`.

Example:
```c
// Set text specific options:
ASSERT_OK(mongocrypt_ctx_setopt_algorithm(ctx, "text", -1), ctx);
ASSERT_OK(mongocrypt_ctx_setopt_algorithm_text(
                ctx,
                TEST_BSON(
                    BSON_STR({
                        "substring" : {"strMinQueryLength" : 5, "strMaxQueryLength" : 10},
                        "prefix" : {"strMinQueryLength" : 10, "strMaxQueryLength" : 20},
                        "case_folding" : true,
                        "diacritic_folding" : false
                    }))),
            ctx);

// To create a find payload, also set the query type:
ASSERT_OK(mongocrypt_ctx_setopt_query_type(ctx, "prefixPreview", -1), ctx);
```

src/mc-textopts.c
```diff
@@ -0,0 +1,305 @@
...
+            bson_iter_t subdoc;
+            if (!BSON_ITER_HOLDS_DOCUMENT(&iter) || !bson_iter_recurse(&iter, &subdoc)) {
+                CLIENT_ERR(ERROR_PREFIX "Expected document for substring, got: %s",
+                           mc_bson_type_to_string(bson_iter_type(&iter)));
+                return false;
+            }
+
+            if (!mc_TextOptsPerIndex_parse(&txo->substring, &subdoc, status)) {
+                return false;
+            }
```
kevinAlbs, line 151: Suggest erroring here if `strMaxLength` is not set for substring.

```c
if (!txo->substring.strMaxLength.set) {
    CLIENT_ERR(ERROR_PREFIX "'strMaxLength' must be set for substring");
    return false;
}
```

Without this check, omitting `strMaxLength` produces an error message that is likely less useful to users (referencing payloads, rather than options):
```
Error parsing FLE2SubstringInsertSpec: Missing field 'mlen' in placeholder
```

src/mc-textopts.c
```diff
@@ -0,0 +1,305 @@
...
+    }
+    return true;
+}
+
+bool mc_TextOpts_parse(mc_TextOpts_t *txo, const bson_t *in, mongocrypt_status_t *status) {
+    bson_iter_t iter = {0};
+    BSON_ASSERT_PARAM(txo);
+    BSON_ASSERT_PARAM(in);
+    BSON_ASSERT(status || true);
+    bool has_caseSensitive = false, has_diacriticSensitive = false, has_substring = false, has_prefix = false,
```
kevinAlbs, line 106: Suggest erroring if `caseSensitive` and `diacriticSensitive` options are not passed. The server [defaults to `true`](https://github.com/10gen/mongo/blob/5a2f1b9e213f4c7c1dde28faabffb486830f811d/src/mongo/crypto/encryption_fields.idl#L135-L144). I expect if the user's value does not match the server, that could lead to incorrect results rather than an error (similar to SERVER-91887).

src/mc-textopts.c
```diff
@@ -0,0 +1,305 @@
...
+         has_suffix = false;
+
+    *txo = (mc_TextOpts_t){0};
+    txo->bson = bson_copy(in);
```
kevinAlbs, line 110: I expect this copied `bson` is leaked. Since it appears not-needed, suggest removing the `bson` field.

src/mc-textopts.c
```diff
@@ -0,0 +1,305 @@
...
+        }
+        END_IF_FIELD;
+
+        CLIENT_ERR(ERROR_PREFIX "Unrecognized field: '%s'", field);
+        return false;
+    }
+    return true;
+}
+
+bool mc_TextOpts_parse(mc_TextOpts_t *txo, const bson_t *in, mongocrypt_status_t *status) {
```
kevinAlbs, line 101: Suggest adding tests similar to `test-mc-rangeopts.c` to exercise different parsing inputs and error scenarios: https://github.com/mongodb/libmongocrypt/commit/368036a2e0a0742ef7c7aaf1b8a8688aa93a0837 (can be copied from [this branch](https://github.com/mongodb/libmongocrypt/compare/master...kevinAlbs:libmongocrypt:text-search?expand=1))

src/mongocrypt-ctx-encrypt.c
```diff
@@ -1852,6 +1872,12 @@ static bool explicit_encrypt_init(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *ms
         }
     }
 
+    if (ctx->opts.textopts.set && ctx->opts.index_type.set) {
```
kevinAlbs, line 1875: Recommend adding test cases in `_test_setopt_for_explicit_encrypt` to test combinations of text opts with other options. Should libmongocrypt error if text opts are set with range opts?

src/mongocrypt.h
```diff
@@ -712,6 +712,7 @@ bool mongocrypt_ctx_setopt_algorithm(mongocrypt_ctx_t *ctx, const char *algorith
 // DEPRECATED: support "RangePreview" has been removed in favor of "range".
 #define MONGOCRYPT_ALGORITHM_RANGEPREVIEW_DEPRECATED_STR "RangePreview"
 #define MONGOCRYPT_ALGORITHM_RANGE_STR "Range"
+#define MONGOCRYPT_ALGORITHM_TEXTPREVIEW_STR "textPreview"
```
kevinAlbs, line 715:
```suggestion
/// NOTE: "textPreview" is experimental only and may be removed in a future non-major release.
#define MONGOCRYPT_ALGORITHM_TEXTPREVIEW_STR "textPreview"
```

Suggested warning follows the warning above the `MONGOCRYPT_QUERY_TYPE_*` constants.

src/mongocrypt.h
```diff
@@ -1530,6 +1531,33 @@ bool mongocrypt_ctx_setopt_query_type(mongocrypt_ctx_t *ctx, const char *query_t
 MONGOCRYPT_EXPORT
 bool mongocrypt_ctx_setopt_algorithm_range(mongocrypt_ctx_t *ctx, mongocrypt_binary_t *opts);
 
+/**
+ * Set options for explicit encryption with the "textPreview" algorithm.
```
kevinAlbs, line 1535:
```suggestion
 * Set options for explicit encryption with the "textPreview" algorithm.
 * 
 * NOTE: "textPreview" is experimental only and may be removed in a future non-major release.
```

Suggested warning follows the warning above the `MONGOCRYPT_QUERY_TYPE_*` constants.

kevinAlbs (commented):
Test additions are much appreciated. This looks close to merge.

I suggest running `git merge master`. That might address the `project fails validation` error preventing the Evergreen check from running.

kevinAlbs (approved):

PR #1054, opened 2025-08-13
Fix mingw-w64 GCC toolchain compatibility issues

eramongodb:
This PR fixes and improves libmongocrypt compatibility with the mingw-w64 GCC toolchain (referred to as just "mingw" going forward for brevity) to unblock upcoming changes to downstream mingw task coverage improvements in the C and C++ drivers.

---

The first issue is with libbson compatibility:

```
 error: #error "__USE_MINGW_ANSI_STDIO > 0 is required for correct PRI* macros"
   27 | #error "__USE_MINGW_ANSI_STDIO > 0 is required for correct PRI* macros"
      |  ^~~~~
```

This is an old mingw compatibility problem that is well-described by [this comment](/fmtlib/fmt/issues/2952#issuecomment-1208246629) concerning format string tests:

> The test failure appears if you compile for the target platform MINGW64 or MINGW32 which uses the old Microsoft C standard library called "msvcrt" which is C89 and also uses some MinGW functions that implement some C99 functionality. `strftime` is part of msvcrt which means it conforms only to C89 and it does not have the new conversion specifiers like `%G` or `%g`. The fix is to upgrade to MinGW+UCRT and just use the new C standard library which properly conforms to new standards.

Upgrading to a UCRT-based mingw-w64 toolchain is beyond the scope of this PR (although we should investigate doing so properly; filed [DEVPROD-20813](https://jira.mongodb.org/browse/DEVPROD-20813) in advance to prepare EVG Windows distros for such a transition). For now, given this mingw workaround applies to the entire codebase (not just libbson), a call to `add_compile_definitions()` is added to the root CMakeLists.txt that defines `__USE_MINGW_ANSI_STDIO=1` for the entire project when CMake detects a `MINGW` compiler is being used.

---

The second issue is with calls to `BCrypt*` API:

```
error: passing argument 9 of 'BCryptEncrypt' from incompatible pointer type [-Werror=incompatible-pointer-types]
      |                               args.bytes_written,
      |                               ~~~~^~~~~~~~~~~~~~
      |                                   |
      |                                   uint32_t * {aka unsigned int *}
note: expected 'ULONG *' {aka 'long unsigned int *'} but argument is of type 'uint32_t *' {aka 'unsigned int *'}
```

The platform-independent struct [`aes_256_args_t`](https://github.com/mongodb/libmongocrypt/blob/6b5d38241b31978efa00b0d21b629c2751530098/src/mongocrypt-crypto-private.h#L178) contains a `uint32_t *bytes_written` pointer, where `uint32_t` is an `unsigned int`. This pointer type is incompatible with mingw where `ULONG` is defined as `unsigned long`. Therefore, a `ULONG bytes_written` local variable is used to obtain a valid pointer to `ULONG` (guarded by `args.bytes_written != nullptr`).

---

The third issue is related to the `__USE_MINGW_ANSI_STDIO` problem above:

```
warning: unknown conversion type character 'z' in format [-Wformat=]
      |       kms_request_str_appendf (v, "%zu", request->payload->len);
      |                                     ^
```

It seems `__attribute__((format(__printf__, ...)))` is not aware of `__USE_MINGW_ANSI_STDIO=1` and uses the MSVCRT-compatible (non-C99-conforming) `printf` mingw definition rather than a C99-conforming definition. These warnings are addressed by using `gnu_printf` rather than `__printf__` (or `printf`) when building with GCC, as currently done [in mongo-c-driver](https://github.com/mongodb/mongo-c-driver/blob/694fbf3da960478f4ec5589df625549a72856777/src/libbson/src/bson/macros.h#L306-L312) (the GCC version check is omitted, as I do not believe we are, or want to, continue supporting such old GCC versions).


kevinAlbs (approved):
LGTM! Thank you for investigating and the fix.

PR #1066, opened 2025-09-09
MONGOCRYPT-824 Allow setting deterministic contention factor during FLE2 field encryption

Julia-Garland:
### Summary

Currently, libmongocrypt chooses a random value if a FLE2 field has a contention factor max greater than zero. This PR introduces a callback `contention_factor_fn` in `crypt->opts` which can be set by the internal `_mongocrypt_opts_set_contention_factor_fn`.

An additional test in `test-monongocrypt-ctx-encrypt` shows how this functionality can be used to deterministically select a contention factor.

### DIFF
--- a/src/mongocrypt-marking.c
+++ b/src/mongocrypt-marking.c
@@ -729,15 +729,17 @@ fail:
 }
 
 static bool _fle2_choose_contention_factor(mongocrypt_t *crypt,
-                                           int64_t exclusive_upper_bound,
+                                           int64_t maxContentionFactor,
                                            int64_t *out,
                                            mongocrypt_status_t *status) {
+    BSON_ASSERT_PARAM(crypt);
+    BSON_ASSERT_PARAM(out);
     if (crypt->opts.contention_factor_fn) {
-        if (!crypt->opts.contention_factor_fn(exclusive_upper_bound, out)) {
+        if (!crypt->opts.contention_factor_fn(maxContentionFactor + 1, out)) {
             CLIENT_ERR("contention_factor_fn failed");
             return false;
         }
-        if (*out < 0 || *out > exclusive_upper_bound) {
+        if (*out < 0 || *out > maxContentionFactor) {
             CLIENT_ERR("chosen contentionFactor out of range");
             return false;
         }
@@ -771,7 +773,7 @@ static bool _mongocrypt_fle2_placeholder_to_insert_update_common(_mongocrypt_key
         /* Choose a contentionFactor in the inclusive range [0,
          * placeholder->maxContentionFactor] */
         if (!_fle2_choose_contention_factor(kb->crypt,
-                                            placeholder->maxContentionFactor + 1,
+                                            placeholder->maxContentionFactor,
                                             &out->contentionFactor,
                                             status)) {
             goto fail;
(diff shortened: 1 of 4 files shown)


addaleax (commented):
I see @kevinAlbs mentioned on the ticket that this is something that probably doesn't need to be exposed to users.

I'm wondering if the same users who have the requirement of being able to specify key material manually via `mongocrypt_ctx_setopt_key_material()` would also want this ability exposed to them?

src/mongocrypt-opts-private.h
```diff
@@ -138,6 +143,10 @@ bool _mongocrypt_opts_kms_providers_validate(_mongocrypt_opts_t *opts,
                                              _mongocrypt_opts_kms_providers_t *kms_providers,
                                              mongocrypt_status_t *status) MONGOCRYPT_WARN_UNUSED_RESULT;
 
+// For testing only: register a custom contention factor function with crypt.
```
kevinAlbs, line 146: FYI @erwee the API for setting a contention factor is added `mongocrypt-opts-private.h`. The header is not installed, but I expect this API can be used by the server (similar to e.g. [mc_FLE2IndexedEncryptedValueV2_parse](https://github.com/10gen/mongo/blob/d433b42be0531d12228e5705ffe6a4483b14c979/src/mongo/crypto/fle_crypto.cpp#L2970))

erwee (changes requested):

eramongodb (approved):
Minor suggestion remaining; otherwise, LGTM.

src/mongocrypt-marking.c
```diff
@@ -728,13 +728,33 @@ static bool _mongocrypt_fle2_placeholder_common(_mongocrypt_key_broker_t *kb,
     return false;
 }
 
+static bool _fle2_choose_contention_factor(mongocrypt_t *crypt,
+                                           int64_t exclusive_upper_bound,
+                                           int64_t *out,
+                                           mongocrypt_status_t *status) {
+    if (crypt->opts.contention_factor_fn) {
```
eramongodb, line 735:
```suggestion
                                           mongocrypt_status_t *status) {
    BSON_ASSERT_PARAM(crypt);
    BSON_ASSERT_PARAM(out);
    if (crypt->opts.contention_factor_fn) {
```
Assert preconditions.

src/mongocrypt-marking.c
```diff
@@ -747,25 +767,12 @@ static bool _mongocrypt_fle2_placeholder_to_insert_update_common(_mongocrypt_key
...
-            }
-        }
-
-        else if (!_mongocrypt_random_int64(crypto,
-                                           placeholder->maxContentionFactor + 1,
-                                           &out->contentionFactor,
-                                           status)) {
+        /* Choose a contentionFactor in the inclusive range [0,
+         * placeholder->maxContentionFactor] */
+        if (!_fle2_choose_contention_factor(kb->crypt, placeholder->maxContentionFactor + 1, &out->contentionFactor, status)) {
```
eramongodb: Recommend moving `exclusive_upper_bound = maxContentionFactor + 1` into `_fle2_choose_contention_factor`.

erwee (approved):

PR #1146, opened 2026-04-10
MONGOCRYPT-895 Add "prefix" and "suffix" query types as aliases for "prefixPreview" and "suffixPreview"

erwee:


kevinAlbs (approved):
Thank you @erwee, LGTM

mdb-ad (approved):

PR #1148, opened 2026-04-17
Migrate Earthly environments to newer supported distro images

eramongodb:
Motivated by MONGOCRYPT-872, where `uv` requires environments to support Python 3.8 or newer. This revealed various EOL distros in the Earthfile which can/should be removed in favor of newer versions. As a drive-by improvement, the Earthfile version is bumped from 0.6 to 0.8 (requiring `COMMAND` -> `FUNCTION`), removing the need for several feature flags.

- Replaced Alpine 3.18 (EOL: May 2025) with 3.23 (latest).
- Centos 6 and 7 are replaced with CentOS Stream 9 and 10. See [prior discussion](https://github.com/mongodb/mongo-c-driver/pull/2083#discussion_r2294204535).
- RockyLinux 8 is replaced with AlmaLinux 8 for better consistency with the C Driver's own Earthfile. See [prior discussion](https://github.com/mongodb/mongo-c-driver/pull/2083#discussion_r2294204535).
- Ubuntu 14.04 (EOL: 2021) and 16.04 (EOL: 2023) are removed. If we exclude Expanded Security Maintenance (ESM) (which is more consistent with DevProd's support policy), 18.04 and 20.04 can/should be dropped as well. Added Ubuntu 24.04 (latest) coverage in their place.
  - Update: per feedback, also removed Ubuntu 18.04 and 20.04.
- Debian 9 (EOL: 2022) and 10 (EOL: 2024) are removed. Debian 13 (latest) was already added by https://github.com/mongodb/libmongocrypt/commit/5a33cd5fc5f960e1ee603323ef96614f6e8f1329.
- Removed OpenSUSE Leap 15.0. It does not appear to be used or required by anything.

The `debian9-arm64-earthly` and `debian10-arm64-earthly` EVG tasks are removed accordingly. I do not expect the packages published by these tasks are (or should be) still in use. The remaining EVG tasks are also migrated from Ubuntu 20.04 distros to run on Ubuntu 24.04 distros instead.


.evergreen/config.yml
```diff
@@ -2081,34 +2081,6 @@ buildvariants:
...
-  display_name: "Debian 9 arm64 (via Earthly)"
-  expansions:
-    packager_distro: debian92
-    packager_arch: arm64
-    earthly_env: deb9
-  tasks:
-  - name: build-deb-packages-with-earthly
-    # Use an arm64 distro to match the intended target architecture of .deb packages.
-    run_on: ubuntu2204-arm64-small
-  - name: publish-deb-packages-with-earthly
```
kevinAlbs, line 2108: Suggest also updating README.md to remove reference to `stretch`, `buster` in the install steps, and noting in the CHANGELOG that Debian 9 and 10 PPA packages will no longer be updated.

Earthfile
```diff
@@ -33,23 +33,19 @@
...
-    #   • rl8 - RockyLinux 8 - Stand-in for RHEL 8
-    #   • c7 - CentOS 7 - Stand-in for RHEL 7
-    #   • c6 - CentOS 6 - Stand-in for RHEL 6
+    #   • rl8 - AlmaLinux 8 - Stand-in for RHEL 8
+    #   • c10 - CentOS Stream 10 - Stand-in for RHEL 10
+    #   • c9 - CentOS Stream 9 - Stand-in for RHEL 9
     #   • amzn1 - AmazonLinux (2018.03)
     #   • amzn2 - AmazonLinux 2
-    #   • deb9 - Debian 9.2
-    #   • deb10 - Debian 10.0
```
kevinAlbs, line 47: Suggest also removing no-longer-needed [workarounds](https://github.com/eramongodb/libmongocrypt/blob/026cb4b04dd8e1df35ab21129c02bb3b98882588/Earthfile#L169-L180) for Debian 9 and 10.

kevinAlbs (commented):
Changes look good with a tweak to the README and CHANGELOG.

> 18.04 and 20.04 can/should be dropped as well

I support dropping in the Earthfile. I do not expect the Earthly targets for Ubuntu 18.04 and 20.04 impact downstream drivers.

.evergreen/config.yml
eramongodb: Done. Also updated the list of Ubuntu names.

Earthfile
eramongodb: Also removed the now-unused `get-deb-signing-keys` target.

eramongodb (commented):
Ubuntu 18.04 and 20.04 have also been removed.

eramongodb:
Given this PR already removes some Earthly-related EVG tasks due to EOL distros, opted to also migrate the remaining Earthly-related EVG tasks from Ubuntu 20.04 to run on the latest Ubuntu 24.04 distros instead.

kevinAlbs (approved):

Earthfile
```diff
@@ -82,91 +76,65 @@ init:
...
 
-env.c7:
-    # A CentOS 7 environment.
-    FROM +init --base=centos:7
+env.c10:
+    # A CentOS Stream 10 environment.
+    FROM +init --base=quay.io/centos/centos:stream10
     DO +REDHAT_SETUP
 
 env.rl8:
```
vector-of-bool: Should probably rename from `rl8` to reflect the distro change

vector-of-bool (approved):
LGTM with one minor comment

eramongodb:
Overlooked migration of the Amazon Linux environments in Earthly.

- Amazon Linux 2 (amzn2) will be EOL in June 2026 (note: not to be confused with the 20YY versioning scheme; it is [its own release](https://aws.amazon.com/amazon-linux-2/) dated [June 2018](https://endoflife.date/amazon-linux)), but the environment only supports up to Python 3.7 (too old for `uv`).
- Amazon Linux 2018 (amzn1) is EOL since Dec 2023.

Replaced both with Amazon Linux 2023 (latest), whose environment supports Python 3.9+.

rcsanchez97 (approved):

PR #1151, opened 2026-04-23
fix rpm-package-build task

rcsanchez97:
* use Fedora 45 for building RPMs (to get the appropriate C driver version)
* tweak __install script to leverage 'dnf build-dep' command

Evergreen patch build: https://spruce.corp.mongodb.com/version/69ea90094789950007a664a3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/etc/install-package.sh
+++ b/etc/install-package.sh
@@ -21,9 +21,14 @@ if test -f /etc/debian_version ; then
         apt-get -y install -- "$@"
 elif test -f /etc/redhat-release || grep 'ID="amzn"' /etc/os-release >/dev/null 1>&2; then
     if test -f /usr/bin/dnf; then
-        # 'dnf' will "do the right thing"
-        # --allow-erasing: see https://github.com/jeroen/curl/issues/350.
-        dnf install -y --allowerasing -- "$@"
+        if echo "$@" | grep -q '[.]spec'; then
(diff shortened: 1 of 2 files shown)


etc/install-package.sh
```diff
@@ -21,9 +21,14 @@ if test -f /etc/debian_version ; then
         apt-get -y install -- "$@"
 elif test -f /etc/redhat-release || grep 'ID="amzn"' /etc/os-release >/dev/null 1>&2; then
     if test -f /usr/bin/dnf; then
-        # 'dnf' will "do the right thing"
-        # --allow-erasing: see https://github.com/jeroen/curl/issues/350.
-        dnf install -y --allowerasing -- "$@"
+        if echo "$@" | grep -q '[.]spec'; then
+            # We use the build-dep command to let dnf do the heavy lifting from the spec file
+            dnf build-dep -y --allowerasing "$@"
```
eramongodb, line 26: Confirming: does `build-dep` also need `--allowerasing`?

eramongodb (approved):

etc/install-package.sh
rcsanchez97: It doesn't strictly need it, but it doesn't hurt anything. However, it could become necessary in the case that one of the default installed packages becomes incompatible with one of our build dependencies. So, I'm inclined to keep it.

PR #1152, opened 2026-04-24
MONGOCRYPT-834 fuzzing for oss-fuzz

mdb-ad:
# Background

Libmongocrypt will be added to [OSS-Fuzz](https://github.com/google/oss-fuzz) for continuous fuzz testing. On our end we only have to provide the fuzzing entry point: `LLVMFuzzerTestOneInput` which is linked into the fuzzing binary. 

# Design

The fuzzing engine which accepts binary inputs uses the first byte to decide on the target state. It provides some defaults (local encryption keys, etc.) until the state machine is ready to accept input and feeds the fuzzer generated data. A corpus of starting data derived from test inputs improves fuzzing efficiency.

# Coverage

The fuzzing added here covers 14.33% of libmongocrypt's functions.


test/data/fuzz_mongocrypt_corpus/op0_encrypt_aggregate
kevinAlbs, line 1: How were these binary files generated? Consider documenting in a new `doc/fuzzing.md` to describe how we can regenerate these if needed.

test/fuzz_mongocrypt.c
kevinAlbs, line 1: Though it may not be strictly needed for OSS-Fuzz, suggest adding a CMake target to help test locally:

```cmake
option (ENABLE_FUZZING "Build fuzzing targets. Requires a compiler supporting -fsanitize=fuzzer." OFF)
# ...
if (ENABLE_FUZZING)
   include (CheckCCompilerFlag)
   check_c_compiler_flag ("-fsanitize=fuzzer" COMPILER_SUPPORTS_FUZZER)
   if (NOT COMPILER_SUPPORTS_FUZZER)
      message (FATAL_ERROR
         "ENABLE_FUZZING is ON but the compiler does not support -fsanitize=fuzzer. "
      )
   endif ()

   add_executable (fuzz_mongocrypt EXCLUDE_FROM_ALL test/fuzz_mongocrypt.c)
   target_include_directories (fuzz_mongocrypt PRIVATE ./src)
   target_link_libraries (fuzz_mongocrypt PRIVATE mongocrypt_static _mongocrypt::libbson_for_static)
   target_compile_options (fuzz_mongocrypt PRIVATE -fsanitize=fuzzer)
   target_link_options (fuzz_mongocrypt PRIVATE -fsanitize=fuzzer)
endif ()
```

Tip: I can run this locally on macOS by using the compilers included in `brew install llvm`:
```bash
cmake \
  -DCMAKE_C_COMPILER="/opt/homebrew/opt/llvm/bin/clang" \
  -DCMAKE_CXX_COMPILER="/opt/homebrew/opt/llvm/bin/clang++" \
  -DENABLE_FUZZING=ON \
  -Bcmake-build
```

test/fuzz_mongocrypt.c
```diff
@@ -0,0 +1,329 @@
...
+        if (input_bin) {
+            init_ok = mongocrypt_ctx_explicit_decrypt_init(ctx, input_bin);
+        }
+        break;
+
+    case OP_DATAKEY: {
+        /* Set up key encryption key for local provider. */
+        uint8_t kek_bson[] = {
+            /* {"provider": "local"} as BSON */
+            0x1b, 0x00, 0x00, 0x00,                                     /* doc len = 27 */
```
kevinAlbs:
```suggestion
            0x19, 0x00, 0x00, 0x00,                                     /* doc len = 27 */
```

Careful. Length appears incorrect. Contents can be checked with PyMongo using `bson.decode`:
```python
bson.decode(bytearray([
    0x19, 0x00, 0x00, 0x00,
    0x02,
    ord('p'), ord('r'), ord('o'), ord('v'), ord('i'), ord('d'), ord('e'), ord('r'), 0x00,
    0x06, 0x00, 0x00, 0x00,
    ord('l'), ord('o'), ord('c'), ord('a'), ord('l'), 0x00,
    0x00
]))
# {'provider': 'local'}
```

test/fuzz_mongocrypt.c
```diff
@@ -0,0 +1,329 @@
...
+            0x02,                                                        /* type: string */
+            'p', 'r', 'o', 'v', 'i', 'd', 'e', 'r', 0x00,              /* key */
+            0x06, 0x00, 0x00, 0x00,                                      /* string len = 6 */
```
kevinAlbs: Suggest running file through clang-format to align comments.

test/fuzz_mongocrypt.c
```diff
@@ -0,0 +1,329 @@
...
+
+    case OP_EXPLICIT_ENCRYPT_EXPRESSION: {
+        uint8_t key_id[16];
+        memset(key_id, 0x61, sizeof(key_id));
+        mongocrypt_binary_t *kid_bin = mongocrypt_binary_new_from_data(key_id, sizeof(key_id));
+        if (kid_bin) {
+            mongocrypt_ctx_setopt_key_id(ctx, kid_bin);
+            mongocrypt_binary_destroy(kid_bin);
+        }
+        mongocrypt_ctx_setopt_algorithm(ctx, "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic", -1);
```
kevinAlbs: Set algorithm and queryType to "range".

I expect the call to `mongocrypt_ctx_explicit_encrypt_expression_init` always fails. Encrypt expression requires algorithm and queryType both be set to range.

libmongocrypt checks the queryType for encryptExpression is "range" ([1](https://github.com/mdb-ad/libmongocrypt/blob/695fd85c6f178e8c5456ac12f615dc49878baa06/src/mongocrypt-ctx-encrypt.c#L2263-L2267)) and later that queryType matches algorithm ([2](https://github.com/mdb-ad/libmongocrypt/blob/695fd85c6f178e8c5456ac12f615dc49878baa06/src/mongocrypt-ctx-encrypt.c#L2164)).

test/fuzz_mongocrypt.c
mdb-ad: Added libbson to the fuzzer so the manual encoding is no longer needed

test/data/fuzz_mongocrypt_corpus/op0_encrypt_aggregate
mdb-ad: Added a standalone Python script that generates each input

kms-message/src/kms_kmip_response_parser.c
```diff
@@ -105,6 +105,10 @@ kms_kmip_response_parser_feed (kms_kmip_response_parser_t *parser,
       uint32_t temp;
       memcpy (&temp, parser->buf->str + FIRST_LENGTH_OFFSET, sizeof (uint32_t));
       parser->first_len = KMS_UINT32_FROM_BE (temp);
+      if (parser->first_len > (uint32_t) INT32_MAX) {
+         KMS_ERROR (parser, "KMIP response length field too large");
+         return false;
+      }
```
kevinAlbs, line 117: Do `git merge master`. This change should no longer be needed with the fix of 92b51448b2a1808b6d8a66e5bc22637228cc520d

test/fuzz_mongocrypt.c
```diff
@@ -0,0 +1,380 @@
...
+            }
+            if (feed_len > 0) {
+                const uint8_t *feed_data = fuzz_consume(data, remaining, feed_len);
+                if (feed_data) {
+                    mongocrypt_binary_t *bin =
+                        mongocrypt_binary_new_from_data((uint8_t *)feed_data, (uint32_t)feed_len);
+                    if (bin) {
+                        mongocrypt_ctx_mongo_feed(ctx, bin);
+                        mongocrypt_binary_destroy(bin);
+                    }
```
kevinAlbs: Suggest removing unnecessary NULL check. I expect `mongocrypt_binary_new_from_data` cannot return NULL (it uses `bson_malloc0`, which aborts on allocation failure).

Suggestion also applies to other calls of `mongocrypt_binary_new_from_data`.

test/gen_fuzz_corpus.py
```diff
@@ -0,0 +1,200 @@
...
+machine.  For NEED_MONGO_* states it reads a 2-byte LE length then that many
+bytes; for NEED_KMS it reads exactly bytes_needed bytes from the KMS context.
+
+Op selectors (must match the enum in fuzz_mongocrypt.c):
+  0  OP_ENCRYPT
+  1  OP_DECRYPT
+  2  OP_EXPLICIT_ENCRYPT
+  3  OP_EXPLICIT_DECRYPT
+  4  OP_DATAKEY            (input_bin ignored; extra bytes unused unless
+  5  OP_REWRAP_MANY_DATAKEY  state machine needs feeding)
```
kevinAlbs: I expect comment only applies to OP_DATAKEY.
```suggestion
  4  OP_DATAKEY            (input_bin ignored; extra bytes unused unless state machine needs feeding)
  5  OP_REWRAP_MANY_DATAKEY  
```

test/gen_fuzz_corpus.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env python3
+"""
+Generate seed corpus files for the fuzz_mongocrypt fuzzer.
+
+Each corpus file has the format:
+  [1 byte: op selector] [BSON payload for the init call] [optional extra bytes]
+
+The optional extra bytes are consumed by drive_ctx() as it walks the state
+machine.  For NEED_MONGO_* states it reads a 2-byte LE length then that many
+bytes; for NEED_KMS it reads exactly bytes_needed bytes from the KMS context.
```
kevinAlbs: IIUC this is describing how the input bytes are read:

```suggestion
bytes; for NEED_KMS it reads exactly bytes_needed bytes from the input.
```

test/gen_fuzz_corpus.py
```diff
@@ -0,0 +1,200 @@
...
+# Corpus entries
+# ---------------------------------------------------------------------------
+
+def build_entries() -> dict[str, bytes]:
+    entries: dict[str, bytes] = {}
+
+    # -- OP_ENCRYPT (0) ------------------------------------------------------
+    # mongocrypt_ctx_encrypt_init expects a command BSON doc.
+    for cmd, coll in [("find", "test"), ("aggregate", "test"), ("insert", "test")]:
+        name = f"op0_encrypt_{cmd}"
```
kevinAlbs: Minor: remove "test" from list since it does not differ.
```suggestion
    for cmd in ["find", "aggregate", "insert"]:
        coll = "test"
```

test/gen_fuzz_corpus.py
kevinAlbs, line 1: Re-run `test/gen_fuzz_corpus.py`. Running locally generates files not in `data/fuzz_mongocrypt_corpus` (e.g. `crash-kmip-parser-overflow`).

kevinAlbs (approved):

PR #1153, opened 2026-04-27
MONGOCRYPT-872 Use `uv tool` to obtain CMake, Ninja, and Python

eramongodb:
Resolves MONGOCRYPT-872. Successor to https://github.com/mongodb/mongo-cxx-driver/pull/1428 and https://github.com/mongodb/mongo-c-driver/pull/2097.

Introduces the `install-build-tools.sh` script, which installs latest versions of `cmake`, `ninja`, and `python` for subsequent use by build-and-test EVG tasks, reducing dependence on system-provided binaries (numerous inconsistent versions across distros). Because of the age of certain EVG distros still in use by the current configuration, the latest `uv-installer.sh` script (version 0.11.8) is bundled and used as a fallback when a system-provided `uv` binary is not available (see: DEVPROD-13413 and DEVPROD-22169). Note that `uv-installer.sh` finally embeds checksums in the script itself, so the checksum patch script used in prior PRs is no longer required.

The Earthfile is also updated to install and use `uv` whenever possible (for compatibility with other updated build scripts which now require `uv`). When a system package provides `uv` directly, it is used (regardless of its version); otherwise, the newest-available `pip` provided by system packages is used to install `uv` instead (always the latest available version). This is made possible following https://github.com/mongodb/libmongocrypt/pull/1148.

All calls to `run_*` (formerly provided by `init.sh`) are replaced with simple commands instead (available via `PATH` exported by `install-build-tools.sh`). All calls to `python` are invoked through `uv`. Note that `install-build-tools.sh` exports `CMAKE_GENERATOR=Ninja` on all non-Windows environments by default, simplifying other helper scripts. However, a comment is added to [this line](https://github.com/mongodb/libmongocrypt/blob/c0353cde50aa9745fe59c14a07772ab7aa9e3777/.evergreen/config.yml#L119) which is responsible for selecting the Ninja generator _even on Windows_, which as a single-config generator (not-obviously) impacts the layout of build artifacts that are depended-on by downstream test scripts such as [bindings/python/.evergreen/test.sh](https://github.com/mongodb/libmongocrypt/blob/c0353cde50aa9745fe59c14a07772ab7aa9e3777/bindings/python/.evergreen/test.sh#L81).

Due to the age of certain EVG distros (notably RHEL 6 and RHEL 7) which do not provide a sufficiently recent Python version to support `uv`, these distros are special-cased to defer partially or entirely to the mongodbtoolchain binaries present in the those environments. These workarounds should be removed once the EVG configuration is updated to migrate away from these EOL, unsupported distros (outside the scope of this PR).


.evergreen/benchmark-python.sh
```diff
@@ -9,23 +9,27 @@ fi
...
 # Include path to installed libmongocrypt.so
 export LD_LIBRARY_PATH="$MONGOCRYPT_INSTALL_PREFIX/lib64"
 cd bindings/python/
 
-/opt/mongodbtoolchain/v4/bin/python3 -m venv venv
+uv venv venv
 . ./venv/bin/activate
-python -m pip install --prefer-binary -r requirements-test.txt
-python -m pip install -e .
+uv python install
```
kevinAlbs: Remove `uv python install`? I expect `python` is already installed by the `uv venv` command.

.evergreen/install-build-tools.sh
```diff
@@ -0,0 +1,82 @@
...
+    script="$(mktemp)" || return
+    cp -f "${EVG_DIR:?}/uv-installer.sh" "${script:?}" || return
+    chmod +x "${script:?}" || return
+    env \
+        UV_INSTALL_DIR="${uv_install_dir:?}" \
+        UV_UNMANAGED_INSTALL=1 \
+        INSTALLER_PRINT_VERBOSE=1 \
+        "${script:?}" || return
+
+    PATH="${uv_install_dir:?}:$PATH" || return
```
kevinAlbs: Suggest removing `|| return` since I expect the variable assignment cannot fail.

```suggestion
    PATH="${uv_install_dir:?}:$PATH"
```

kevinAlbs (approved):
LGTM with minor comments.

vector-of-bool (approved):
LGTM

PR #1156, opened 2026-05-01
MONGOCRYPT-899 fix `upload-all` copy

kevinAlbs:
The `upload-all` task on the 1.18.0 release [failed](https://spruce.corp.mongodb.com/task/libmongocrypt_release_publish_upload_all_0dfa86a00f207395600be7059afadd73bdfc69b3_26_05_01_14_03_13/logs?execution=1):

```
cp: 'libmongocrypt-all-1.18.0.tar.gz' and 'libmongocrypt-all-1.18.0.tar.gz' are the same file
```

Due to this copy:
```bash
if [ -n "${tag_upload_location}" ]; then
   # ...
   cp -a libmongocrypt-all-${tag_upload_location!|*revision}.tar.gz libmongocrypt-all-${tag_upload_location}.tar.gz
```

`${tag_upload_location!|*revision}` and `${tag_upload_location}` evaluate the same. But since the copy is guarded by a check for a non-empty `${tag_upload_location}`, the copy is unnecessary.

I do not plan to immediately release 1.18.1 with this fix. libmongocrypt-all.tar.gz is planned for removal in MONGOCRYPT-894. No drivers appear to get binaries directly from this tarball. But I erred towards making this small fix in case there is a need for a 1.18.1 release.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -871,7 +871,6 @@ tasks:
           if [ -n "${tag_upload_location}" ]; then
             # the "fetch source" step detected a release tag on HEAD, so we
             # prepare a local file for upload to a location based on the tag
-            cp -a libmongocrypt-all-${tag_upload_location!|*revision}.tar.gz libmongocrypt-all-${tag_upload_location}.tar.gz
 
             if [[ "$tag_upload_location" = *-* ]]; then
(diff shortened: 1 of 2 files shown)


rcsanchez97 (approved):
LGTM

PR #1177, opened 2026-05-21
MONGOCRYPT-917 fix padding calculation for prefix/suffix

kevinAlbs:
# Summary
Fix padding calculation when generating affixes for `FLE2InsertUpdatePayloadV2`.
# Background & Motivation
Intended to ensure the count of prefix or suffix tokens ("b.p" / "b.u") increase along with the length of encrypted value ("v"). See report in MONGOCRYPT-913 for details.
# Testing
Assertions are added to `test_ciphertext_len_steps_fle2_text_search` to count of suffix tokens. This serves as a regression test, and seemed simpler than the test in the report's Appendix B.

To additionally verify changes, the QE Text driver [spec](https://github.com/mongodb/specifications/tree/25bee54ba6db27c11ea53e9608f4bd65e4552ac7/source/client-side-encryption/tests/unified) and [prose](https://github.com/mongodb/specifications/tree/25bee54ba6db27c11ea53e9608f4bd65e4552ac7/source/client-side-encryption/tests#27-text-explicit-encryption) tests were run locally with the C driver.


src/mc-text-search-str-encode.c
```diff
@@ -38,7 +38,7 @@ static mc_affix_set_t *generate_prefix_or_suffix_tree(const mc_utf8_string_with_
     BSON_ASSERT_PARAM(base_str);
     BSON_ASSERT_PARAM(out_msize);
     // We encrypt (unfolded string + 5 bytes of extra BSON info) with a 16-byte block cipher.
-    uint32_t encrypted_len = 16 * (uint32_t)((unfolded_byte_len + OVERHEAD_BYTES + 15) / 16);
+    uint32_t encrypted_len = 16 * (uint32_t)((unfolded_byte_len + OVERHEAD_BYTES + 16) / 16);
```
erwee, line 42: Please add a comment above noting that PKCS7 adds an extra 16-byte padding for 16-byte aligned plaintext lengths.

test/test-mc-text-search-str-encode.c
```diff
@@ -52,7 +52,7 @@ static void test_nofold_suffix_prefix_case(_mongocrypt_tester_t *tester,
     uint32_t byte_len = (uint32_t)strlen(str);
     uint32_t unfolded_codepoint_len = byte_len == 0 ? 1 : get_utf8_codepoint_length(str, byte_len);
     uint32_t folded_codepoint_len = byte_len == 0 ? 0 : unfolded_codepoint_len - foldable_codepoints;
-    uint32_t padded_len = 16 * (uint32_t)((byte_len + 5 + 15) / 16) - 5;
```
erwee, line 55: Would it be possible to just invoke the cbc encryption here and get the actual ciphertext length from it?

test/test-mc-text-search-str-encode.c
erwee, line 53: Since the tests that call this use randomized lengths, I think we need a test that calls this function with a fixed input string that would be 16-byte aligned after the 5-byte overhead is added, and with a `ub` that is larger than the would-be padded-length. (the rest of the parameters can be fixed). That would ensure that the calculated msize would be incorrect and fail the test if our calculation is wrong.

erwee: this line also needs to be updated

erwee (changes requested):

test/test-mc-text-search-str-encode.c
kevinAlbs: Replaced with a call to a new `calculate_padded_length` helper. The helper uses `_mcFLE2v2AEADAlgorithm()->get_ciphertext_len` to calculate the padding matching CBC.

test/test-mc-text-search-str-encode.c
kevinAlbs: Great catch. Replaced with a call to a new `calculate_padded_length` helper. This caused a test failure due to `generate_substring_tree` having the same padding calculation issue.

kevinAlbs: Added a test to call this with a string of 11 characters and an `ub` of 30. Confirmed the new test fails without the fix.

erwee (approved):

rcsanchez97 (approved):

PR #1184, opened 2026-05-28
MONGOCRYPT-870 drop `prefixPreview` and `suffixPreview`

kevinAlbs:
# Summary

This PR applies renames for query types and algorithms related to QE string features.

- Setting the query type `prefixPreview` and `suffixPreview` is now an error.
- `suffix` and `prefix` require use of the `string` algorithm (formerly `textPreview`).

# Details

The `textPreview` algorithm is still used for `substringPreview`. But `textPreview` may be dropped entirely if [WRITING-37758](https://jira.mongodb.org/browse/WRITING-37758) is approved. This is tracked in MONGOCRYPT-931.

Choosing `string` (rather than `text`) was motivated by a request to rename public driver APIs from "text" to "string": DRIVERS-3531

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Added
+- Stable support for `prefix` and `suffix` query types.
+
+### Removed
+- Support for the experimental `prefixPreview` and `suffixPreview` query types.
+
 <!-- TODO: add entries for next release -->
 
 ## 1.18.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,10 @@
 ## Unreleased
 
 ### Added
-- Stable support for `prefix` and `suffix` query types.
+- Stable support for prefix and suffix queries:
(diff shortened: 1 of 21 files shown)


erwee (approved):

blink1073:
I would have expected the python test that uses `suffixPreview` to have failed.  I need to double-check that we're using the as-build version of `libmongocrypt` in our Evergreen tests.

blink1073:
> I would have expected the python test that uses suffixPreview to have failed. I need to double-check that we're using the as-build version of libmongocrypt in our Evergreen tests.

Ah, I missed that you had update the python tests.

blink1073 (approved):
LGTM!

kevinAlbs:
Drafting spec changes motivated an additional change: https://github.com/mongodb/libmongocrypt/pull/1184/changes/f1df06ed5b23e6637221ace16f90fb65fad844dc drops "textPreview" entirely. "substringPreview" still signals the feature is experimental, and avoids continued use of the term "text".

mdb-ad (approved):

blink1073 (approved):

PR #1197, opened 2026-06-29
PYTHON-5896 - [pymongocrypt] Bundle libmongocrypt 1.20.0

NoahStapp:


bindings/python/pymongocrypt/binding.py
```diff
@@ -1522,9 +1528,9 @@ def _parse_version(version):
 
 /// String constants for setopt_query_type
 // DEPRECATED: Support "rangePreview" has been removed in favor of "range".
-/// NOTE: "substringPreview" is experimental and may be removed in a future non-major release.
-/// NOTE: "suffixPreview" is experimental and may be removed in a future non-major release.
-/// NOTE: "prefixPreview" is experimental and may be removed in a future non-major release.
+/// DEPRECATED: Support for "substringPreview" has been removed in favor of "substring"
+/// DEPRECATED: Support for "suffixPreview" has been removed in favor of "suffix"
+/// DEPRECATED: Support for "prefixPreview" has been removed in favor of "suffix"
```
blink1073: This comment is wrong.  We should fix it [here](https://github.com/mongodb/libmongocrypt/blob/906010bdf5a9c7d5fc40743c49b313cdd57effa4/src/mongocrypt.h#L1609) and re-generate.

NoahStapp: Which comment specifically in the context of libmongocrypt 1.20?

blink1073: "/// DEPRECATED: Support for "prefixPreview" has been removed in favor of "suffix"" should be "prefix" at the end.

kevinAlbs (approved):
Oop, thank you for the comment fix. LGTM.

blink1073 (approved):
LGTM!

PR #1200, opened 2026-07-10
MONGOCRYPT-951: Pin cmake to 4.3.4

prestonvasquez:
MONGOCRYPT-951

Pin to avoid issue in parent task while investigating the following error from the cmake 4.4.0 release:

```
CMake Error: The warning category "error -DCMAKE_CXX_LINK_EXECUTABLE='echo not linking now...'" is not known.
CMake Error at cmake-build/_deps/embedded_mcd-src/src/libmongoc/CMakeLists.txt:425 (TRY_COMPILE):
  Failed to configure test project build system.
  cmake-build/_deps/embedded_mcd-src/src/libmongoc/CMakeLists.txt:449 (mongoc_get_accept_args)
```

### DIFF
--- a/.evergreen/install-build-tools.sh
+++ b/.evergreen/install-build-tools.sh
@@ -51,7 +51,7 @@ install_build_tools() {
   fi
 
   if [[ "${rhel7:?}" == "1" ]]; then
-    uv tool install -q cmake || return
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you for the fix.

PR #1201, opened 2026-07-13
Do not invoke `ctest` when `BUILD_TESTING=OFF`

eramongodb:
Resolve the following error when libmongocrypt is built by the C Driver with CMake 4.4.0 and `BUILD_TESTING=OFF`:

```
*********************************
No test configuration file found!
*********************************
Usage
  ctest [options]
```

Per CMake 4.4.0 [release notes](https://cmake.org/cmake/help/latest/release/4.4.html):

> `ctest` now exits with non-zero if invoked with no arguments and no test input file exists.

### DIFF
--- a/.evergreen/build_all.sh
+++ b/.evergreen/build_all.sh
@@ -115,7 +115,7 @@ if [[ "$release_os_arch" == *glibc* ]]; then
     echo "glibc version check passed: $actual_glibc"
 fi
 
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you!

eramongodb:
Subtle bugfix: `is_true` expands the given argument as a variable. Need to pass the name of the boolean variable, not the value.

PR #1208, opened 2026-07-23
DEVPROD-34208: use DevProd ECR for container images instead of Artifactory

fetsko:


tarunsethi:
@kevinAlbs @vector-of-bool Do you mind reviewing this PR when you get a chance? Because of security concerns, artifactory.corp.mongodb.com has been shutdown so any builds using it would fail.

bindings/python/CONTRIBUTING.md
vector-of-bool, line 1: The `bindings/python/` subdirectory is owned by Python devs and may require some input from a Python dev member. @kevinAlbs do you know who best to look at this?

.evergreen/earthly.sh
vector-of-bool, line 1: Not sure this change to the earthly script is necessary, and it assumes Docker when some of us use Podman. These tasks aren't run manually very often and I think we can leave the authentication as a manual step for now.

vector-of-bool (approved):
The changes LGTM other than the comment about the earthly script changes, which I think are mostly unnecessary complication.

Also may want more input from the people that maintain the Python bindings' scripts.

vector-of-bool:
Kevin and I will be absent on Friday, July 31. I don't think this is urgent for us as the tasks in question aren't mandatory until we need to do a release.

Recommend consulting the Python team about the Python bindings' scripts.

.evergreen/earthly.sh
fetsko: Should I revert the updates to this file then?

Also feel free to make updates as needed to this PR if preferred.

kevinAlbs: > change to the earthly script is necessary

I expect no. I also expect users running `earthly.sh` can authenticate before running this script. I reverted changes to this file.

> assumes Docker when some of us use Podman

Updated instructions in `releasing.md` to note either Docker or Podman can be used.

bindings/python/CONTRIBUTING.md
kevinAlbs: I tweaked these instructions. The `update-version.sh` script only appears to be run locally. Similar to earthly.sh, I removed the `login` step from `update-version.sh` and updated these instructions to note authentication may be necessary.

> do you know who best to look at this?

@blink1073 would you mind reviewing the changes to the Python libmongocrypt script to switch from Artifactory to ECR?

kevinAlbs (approved):

blink1073 (approved):
Python changes LGTM

kevinAlbs:
Thank you for the PR @fetsko!

PR #1211, opened 2026-08-03
MONGOCRYPT-955 add AGENTS.md

kevinAlbs:
# Summary

Add an AGENTS.md for libmongocrypt.

As a drive-by: fix formatting steps CONTRIBUTING.md to refer to `uv` (not `pipx`).

# Background & Motivation

Organization of build steps was inspired by https://github.com/mongodb/mongo-cxx-driver/pull/1636.

## Evaluating

I used Claude to draft an AGENTS.md evaluator. It tested responses to read-only prompts before / after adding AGENTS.md. The draft is in [this branch](https://github.com/kevinAlbs/libmongocrypt/compare/M955...kevinAlbs:libmongocrypt:with-evals.M955?expand=1). Response correctness did not change much. But token usage reduced:

| model | passed before | passed after | Δ | tokens before | tokens after | Δ |
|---|---|---|---|---|---|---|
| sonnet | 52/60 (87%) | 60/60 (100%) | 🟢 +13% | 1,412,187 | 849,919 | 🟢 -40% |
| opus | 56/60 (93%) | 59/60 (98%) | 🟢 +5% | 1,630,504 | 720,407 | 🟢 -56% |
| haiku | 55/60 (92%) | 51/60 (85%) | 🔴 -7% | 3,010,864 | 690,846 | 🟢 -77% |

The reduced token usage gave me some confidence the AGENTS.md helped lower the need to explore. But I was not confident this evaluator was valuable enough to add to this PR. And I did not want to continue adding scope to this task.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@ This file provides guidance to AI coding agents working with code in this reposi
 
 ## What this is
 
-libmongocrypt is a C library that assists MongoDB drivers in implementing **In-Use Encryption**. The driver-facing behavior it helps support is defined by the Client-Side Encryption specification: <https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md>.
+libmongocrypt is a C library that assists MongoDB Client Libraries in implementing **In-Use Encryption**. The Client Library behavior it helps support is defined by the Client-Side Encryption specification: <https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md>.
 
 ### Naming (from the spec)
 
@@ -123,9 +123,9 @@ Tests are C files, not auto-discovered. Two steps are required:
 
 ### Components
 
-- Driver
+- Client Library (E.g. pymongo, sometimes referred to as a Driver)
     - libmongocrypt acts as a **state machine** and does no I/O.
-    - libmongocrypt provides a C interface to drivers.
+    - libmongocrypt provides a C interface to the Client Library.
     - Most APIs pass and returns BSON, most structs are opaque, and global init is lazy.
     - See `integrating.md` and the public API in `src/mongocrypt.h`.
 - MongoDB server (mongod / mongos)
@@ -136,7 +136,7 @@ Tests are C files, not auto-discovered. Two steps are required:
 - mongocryptd / crypt_shared
     - Required for automatic encryption only (not used by explicit encryption, explicit decryption, or automatic decryption)
     - Interchangeable: both do query analysis for automatic encryption.
-        - mongocryptd is a process the driver talks to.
+        - mongocryptd is a process the Client Library talks to.
         - crypt_shared is a library libmongocrypt loads and calls directly.
     - Only permitted for enterprise or Atlas. This makes automatic encryption an enterprise / Atlas feature.
 
@@ -154,8 +154,8 @@ Uses C99. Contributions shall not use features from newer standards.
 
 ### Core objects & state machine
 
-- **`mongocrypt_t`** — the top-level handle; **thread-safe**, and expected to be owned by a driver's `MongoClient` / `ClientEncryption` object.
-- **`mongocrypt_ctx_t`** — a single operation (encrypt, decrypt, data-key creation, …); **not thread-safe**, and expected to handle one operation. The driver drives it as a state machine, performing the I/O the library requests at each state. See `integrating.md` for the state machine.
+- **`mongocrypt_t`** — the top-level handle; **thread-safe**, and expected to be owned by a Client Library's `MongoClient` / `ClientEncryption` object.
+- **`mongocrypt_ctx_t`** — a single operation (encrypt, decrypt, data-key creation, …); **not thread-safe**, and expected to handle one operation. The Client Library drives it as a state machine, performing the I/O the library requests at each state. See `integrating.md` for the state machine.
 - **`mongocrypt_binary_t`** is a non-owning view; `_destroy` frees the view, not the underlying data.
 
 ## Before Committing
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,171 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents working with code in this repository.
+
+## What this is
+
+libmongocrypt is a C library that assists MongoDB drivers in implementing **In-Use Encryption**. The driver-facing behavior it helps support is defined by the Client-Side Encryption specification: <https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md>
+
+### Naming (from the spec)
+
+The public feature is **In-Use Encryption**, consisting of **Client-Side Field Level Encryption (CSFLE)** and **Queryable Encryption (QE)**. Internally, In-Use Encryption is sometimes called **Field Level Encryption (FLE)**; CSFLE is sometimes called Client Side Encryption (as in the spec title); and CSFLE and QE are sometimes called **FLE1** and **FLE2** respectively. So the `mc-fle2-*` payload files under `src/` implement **Queryable Encryption**, not CSFLE.
+
+Separately (and confusingly), **`csfle`** is sometimes used within libmongocrypt to mean the **crypt_shared** library — an older name for it, unrelated to the CSFLE feature above. For example, `crypt->csfle` is the crypt_shared vtable.
+
+## Build System
+
+This project uses CMake.
+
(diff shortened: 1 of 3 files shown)


AGENTS.md
vector-of-bool, line 1: Can the text in this file be reflowed to a column limit? Not important but aides readability across both very wide and very narrow screens.

AGENTS.md
```diff
@@ -0,0 +1,171 @@
...
+The optional install step:
+
+```bash
+cmake --install cmake-build
+```
+
+> [!IMPORTANT]
+> For multi-configuration generators (e.g. "Visual Studio *", "Ninja Multi-Config", etc.), use `--config <config>` during the build, install, and test steps instead of `CMAKE_BUILD_TYPE=<config>`.
+> The `CMAKE_BUILD_TYPE` option will be ignored by the configuration step.
+> Only use `CMAKE_BUILD_TYPE` with single-configuration generators (e.g. Makefile Generators, Ninja, etc.).
```
vector-of-bool: I believe that models will have information like this baked-in. e.g. I've never needed to guide an LLM on how multi-config generators work, it just knows.

I think the only thing that needs to be specified is the configuration parameters that are specific to this project.

AGENTS.md
```diff
@@ -0,0 +1,171 @@
...
+
+Key CMake configuration options (given `option=(default|alternatives...)`):
+
+- `-G <generator-name>`: specify a build system generator (e.g. `Ninja`).
+- `-D CMAKE_PREFIX_PATH:PATH=<prefix>`: installation prefixes to search with `find_*()` (e.g. an installed libbson with `USE_SHARED_LIBBSON=ON`, or OpenSSL).
+- `-D CMAKE_INSTALL_PREFIX:PATH=<install-prefix>`: install directory used by `install()`. Use `cmake-build/install/` when system modification is undesirable or disallowed by the user.
+- `-D CMAKE_BUILD_TYPE:STRING=<config>`: build type on single-configuration generators.
```
vector-of-bool: Same as comment above

AGENTS.md
```diff
@@ -0,0 +1,171 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents working with code in this repository.
+
+## What this is
+
+libmongocrypt is a C library that assists MongoDB drivers in implementing **In-Use Encryption**. The driver-facing behavior it helps support is defined by the Client-Side Encryption specification: <https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md>
```
connorsmacd:
```suggestion
libmongocrypt is a C library that assists MongoDB drivers in implementing **In-Use Encryption**. The driver-facing behavior it helps support is defined by the Client-Side Encryption specification: <https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md>.
```

Nit: Missing period.

AGENTS.md
```diff
@@ -0,0 +1,171 @@
...
+
+> [!NOTE]
+> `.evergreen/build_all.sh` is the authoritative reference for CI configure-build-install routines. Consult for platform-specific options and flags.
+
+> [!NOTE]
+> For local development and testing, use the Debug configure in the [Running Tests](#running-tests) section instead of the release configure above.
+
+## Dependencies
+
+Dependencies are obtained in three different ways — most are fetched or vendored, so a network connection may be needed at configure time:
```
connorsmacd: > three different ways

It's not immediately clear to me what the three different ways are, so consider briefly elaborating on what this means. You could might even consider just removing the "three" then leaving it as is if enumerating doesn't add much value.

AGENTS.md
```diff
@@ -0,0 +1,171 @@
...
+Tests are C files, not auto-discovered. Two steps are required:
+
+1. Add the file to the `TEST_MONGOCRYPT_SOURCES` list in `CMakeLists.txt`.
+2. In `test/test-mongocrypt.c`, add an `_mongocrypt_tester_install_*(&tester)` call in `main`, and register each test function inside that installer via `_mongocrypt_tester_install(tester, "_test_name", fn, CRYPTO_REQUIRED)` (or `CRYPTO_OPTIONAL`).
+
+## Architecture
+
+### Components
+
+- Driver
```
connorsmacd: I think we already use the term "driver" a bit too liberally in DBX, so consider changing this to something else, e.g., "Library" or "State Machine".

AGENTS.md
```diff
@@ -0,0 +1,171 @@
...
+    - See `integrating.md` and the public API in `src/mongocrypt.h`.
+- MongoDB server (mongod / mongos)
+    - To send the final commands.
+    - To fetch remote schemas.
+- KMS
+    - To encrypt/decrypt Data Encryption Keys (DEK) using a backing Key Encryption Key (KEK).
+- mongocryptd / crypt_shared
+    - Required for automatic encryption only (not used by explicit encryption, explicit decryption, or automatic decryption)
+    - Interchangeable: both do query analysis for automatic encryption.
+        - mongocryptd is a process the driver talks to.
```
connorsmacd: Consider rewording this to "mongocryptd is a daemon that libmongocrypt talks to via ..." (I can't remember if it's ISP, a TCP connection, or something else). This also addresses another potentially confusing use of "driver".

connorsmacd (approved):
LGTM with some minor feedback. Erring on the side of approving this now since I will be away until next week.

AGENTS.md
kevinAlbs: Good point. I removed this sentence since I do not think it adds much to the list of dependencies that follows.

AGENTS.md
kevinAlbs: Good point. I updated references to "Client Library" to match the term used in other [public documentation](https://www.mongodb.com/docs/drivers/). I noted "sometimes referred to as a Driver".

AGENTS.md
kevinAlbs: I like that clarification. I updated to note that the Client Library communicates to mongocryptd using the same wire protocol as the MongoDB server.

AGENTS.md
```diff
@@ -0,0 +1,169 @@
...
+
+```bash
+./cmake-build/test-mongocrypt                              # whole suite
+./cmake-build/test-mongocrypt _test_setopt_kms_providers   # a single test, by function name
+```
+
+> [!IMPORTANT]
+> For multi-configuration generators, the executable appears under a `<config>/` subdirectory (e.g. `cmake-build/Debug/test-mongocrypt`).
+
+A few tests need a real crypt_shared library; download one with the `mongodl.py` script from drivers-evergreen-tools: <https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/mongodl.py>
```
vector-of-bool, line 80: I have a user skill for `mongodl` that may be useful for sharing across projects in the org, but I don't know how to do that.

vector-of-bool (commented):
Sorry, thought I submitted my pending comments a few days ago. A few comments regarding redundant information that the model already knows.

AGENTS.md
kevinAlbs: I originally erred towards consistency with the C and C++ driver AGENTS.md, but I tend to agree. I expect CMake instructions common to any CMake project are unlikely to help an LLM. I removed some CMake notes, and left notes about preferences (e.g. using `cmake-build`) or project specific options.

AGENTS.md
kevinAlbs: I have not tried it, but https://github.com/10gen/agent-skills might be an option for an org-wide skill.

AGENTS.md
kevinAlbs: I am slightly averse to hard-wrapping I think I have a historical peeve of my small text changes in the spec repo creating big diffs after rewrapping. I expect that is less of an issue with modern GitHub highlighting content changes.

I am admittedly biased towards my own workflow with VS Code. VS Code does not appear to have built-in support for hard wrapping. But I can enable soft wrapping in VS Code with user settings:
```json
    "[markdown]": {
        "editor.wordWrap": "wordWrapColumn",
        "editor.wordWrapColumn": 120,
        "editor.rulers": [
            120
        ],
    },
```

Image

With that said, would hard wrapping still seem preferable? I am open to it if it may help with reading (though I might poke you for recommendations on a wrapping tool).

vector-of-bool: Soft wrapping may be a useful compromise.

The wrapping happens automatically for me because I have an extension, "Reflow Markdown", installed that formats markdown files on save. It may not be the optimal formatter though, and it can garble some constructs like admonitions. May be better to adopt a proper tool like mdformat in the future. I'd say leave it for now and we can discuss reformatting of everything altogether at some future date.

vector-of-bool (approved):
LGTM

PR #1213, opened 2026-08-04
check system `uv` for `python` subcommand

kevinAlbs:
This PR checks that the system `uv` supports the `python` sub-command.

Intended to fix tasks on distros with a too-old `uv`. [Example](https://spruce.corp.mongodb.com/task/libmongocrypt_macos_x86_64_build_and_test_and_upload_1cbe6fe06306454cef6a0531c59ccf3f0067800a_26_08_04_18_04_10/logs?execution=0):

```
[2026/08/04 14:08:01.555] error: unrecognized subcommand 'python'
[2026/08/04 14:08:01.555] Usage: uv [OPTIONS] <COMMAND>
```

The `uv python` subcommand was added in `uv` [0.3.0](https://github.com/astral-sh/uv/blob/main/changelogs/0.3.x.md#030).

### DIFF
--- a/.evergreen/install-build-tools.sh
+++ b/.evergreen/install-build-tools.sh
@@ -17,6 +17,15 @@ export_uv_tool_dirs() {
   export PATH UV_TOOL_DIR UV_TOOL_BIN_DIR UV_PYTHON_INSTALL_DIR
 }
 
+# Returns 0 if system uv supports the `uv python` subcommand (added in uv 0.3.0).
+uv_supports_python() {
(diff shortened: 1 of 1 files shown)


eramongodb (approved):
Suggest adding reference to a ticket to remove this workaround once the deprecated `macos-11` distro is dropped from libmongocrypt's EVG config.

kevinAlbs:
> Suggest adding reference to a ticket

Good idea. Filed MONGOCRYPT-961 and noted in comment.

PR #1215, opened 2026-08-14
MONGOCRYPT-834 add fuzz_kms to ENABLE_FUZZING cmake option

mdb-ad:
Add the old fuzzer to the ENABLE_FUZZING target to simplify the oss-fuzz script


test/fuzz_kms.c
kevinAlbs, line 14: Though I expect a fuzzer would not pass an input this large, suggest checking before the cast to ensure no overflow;

```suggestion
int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    if (size > UINT32_MAX) {
        return 0;
    }
```

kevinAlbs (approved):
LGTM with a minor suggestion to check before the cast.

PR #1218, opened 2026-08-20
Run pymongocrypt release task on macOS 14

kevinAlbs:
To address task failures of `release-python-macos-1100` due to a too-old `rustc`. [Example](https://spruce.corp.mongodb.com/task/libmongocrypt_python_release_release_python_macos_1100_edadab0e1bd2693de6116503891c237bccaa1b56_26_08_17_21_01_07/logs?execution=0):

```
Building wheel for cryptography (pyproject.toml) did not run successfully.
...
error: rustc 1.78.0 is not supported by the following packages:
```

MONGOCRYPT-961 plans to (soon) drop support for macOS 11 regardless. But this may be a short-term improvement to fix failing CI tasks.

Disclaimer: I used Claude to identify and propose the fix.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -607,7 +607,7 @@ tasks:
     - func: "test python integ"
       vars: { variant_name: "${build_variant}" }
 
-- name: "release-python-macos-1100"
+- name: "release-python-macos-14-arm64"
   tags: ["release_python_tag"]
   # Run task on macOS 14 since macOS 11 is EOL and fails to run release script.
   # The macOS platform is set by MACOSX_DEPLOYMENT_TARGET/_PYTHON_HOST_PLATFORM in scripts/release.sh, so the produced artifacts do not depend on the host macOS version.
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -609,7 +609,9 @@ tasks:
 
 - name: "release-python-macos-1100"
   tags: ["release_python_tag"]
-  run_on: macos-1100
+  # Run task on macOS 14 since macOS 11 is EOL and fails to run release script.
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Independent corroboration from Evergreen itself: version `6a86f34aaad61f00075b2784` (order 2331, today's Python Daily Build) reports these admin-defined warnings:

> task `release-python-macos-1100` in buildvariant `python-release` references distro `macos-1100` with the following admin-defined warning(s): This distro is now End of Life. Please consider using a supported OS version.

> buildvariant `macos_x86_64` references distro `macos-1100` with the following admin-defined warning: This distro is now End of Life.

So the distro is flagged EOL by Evergreen admins, not just inferred from the `cryptography`/rustc failure.

One scoping note: that second warning is a *separate* exposure this PR deliberately does not touch. The `macos_x86_64` buildvariant (`build-and-test-and-upload`, `build-and-test-shared-bson`, `build-and-test-asan-mac`) also runs on `macos-1100`, but it is currently **green**, and unlike the release task its host choice is not incidental — it is the x86_64 test coverage. Moving it to `macos-14-arm64` would silently drop x86_64 coverage, which is a different decision for the team to make. The release task is safe to move precisely because `scripts/release.sh` pins the wheel platform tag, so the host arch does not affect the artifact.

.evergreen/config.yml
```diff
@@ -609,7 +609,9 @@ tasks:
 
 - name: "release-python-macos-1100"
```
blink1073:
```suggestion
- name: "release-python-macos-14-arm64"
```

rcsanchez97 (approved):
LGTM, with suggested change from @blink1073

blink1073 (approved):
LGTM, thanks!

PR #1219, opened 2026-08-23
MONGOCRYPT-978 Do not build test binaries when BUILD_TESTING is off

tvdw-mongodb:
example-state-machine, example-state-machine-static and kms-message's test_kms_request register themselves with add_test but were gated only on crypto availability. BUILD_TESTING defaults to ON (via include(CTest), added to kms-message for standalone builds), so default builds are unchanged.


kevinAlbs (approved):
Thanks for the PR @tvdw-mongodb. This seems reasonable regardless. But curious: what motivated this change? Did this impact build on another project including libmongocrypt?

tvdw-mongodb:
> Thanks for the PR @tvdw-mongodb. This seems reasonable regardless. But curious: what motivated this change? Did this impact build on another project including libmongocrypt?

We have a statically-linked FIPS build of libmongocrypt that swaps openssl for boringssl, and it has a rather interesting toolchain to get us there. The combination of static linking and the boringssl swap meant it failed to build these two tests, and while that's totally a problem of the way we build things, ensuring that we don't build the tests at all (under BUILD_TESTING=false) was the easier fix 😉

PR #1225, opened 2026-09-04
MONGOCRYPT-985 return NULL instead of false in pointer-returning functions

GromNaN:
Five `return false;` statements sit inside functions that return a pointer. Clang reports each one as `-Wbool-conversion`: "initialization of pointer of type 'X *' to null from a constant boolean expression". This changes them to `return NULL;`.

There is no functional change. In C99 `false` is the macro `0`, which is a null pointer constant, so every one of these already returned exactly `NULL` and no caller could tell the difference. The code is valid only because of that rule; the same lines would be rejected by a C++ compiler, and they warn under C23 where `false` has type `bool`. Writing `NULL` says what is meant and silences the diagnostic.

Sites:

- `src/mc-range-mincover.c`: four returns in the `CHECK_BOUNDS` macro. The macro is expanded only inside `mc_getMincoverInt32`, `mc_getMincoverInt64`, `mc_getMincoverDouble` and `mc_getMincoverDecimal128`, all of which return `mc_mincover_t *`.
- `src/mongocrypt-ctx.c`: the `MONGOCRYPT_CTX_ERROR` case in `mongocrypt_ctx_mongo_db`, which returns `const char *`. Every other branch of that switch already returns `NULL`.

I checked the five other `case MONGOCRYPT_CTX_ERROR: return false;` lines in `mongocrypt-ctx.c`. They are all inside `bool`-returning functions and are left untouched. A sweep of the rest of `src/` found no further instances of this pattern.

Found while building the bundled sources in the MongoDB PHP driver, where the extension compiles them with clang defaults, currently C23.


kevinAlbs (approved):
Thank you @GromNaN!

